PR #8, opened 2020-06-01
Ss testing pipeline

samuelsunvold:
Finished unit tests for users.service and for users.controller. Also setup test.yml so that it creates a Postgres service container


src/users/users.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        expect(usersService.update).toBeCalledTimes(1);
+    });
+
+    // Tests the remove function
+    it("should remove", () => {
+        usersController.remove(consts.ID, consts.DELETE_USER_DTO_TEST_OBJ);
+        expect(usersService.remove).toHaveBeenCalledWith(consts.ID, consts.DELETE_USER_DTO_TEST_OBJ);
+        expect(usersService.remove).toBeCalledTimes(1);
+    });
+});
```
Bialogs, line 63: Please add newline at the end of this file.

src/users/users.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import { UsersController } from './users.controller';
+import { UsersService } from './users.service';
+import * as consts from '../test.constants';
+import { CreateUserDto } from './dto/create-user.dto';
+import { UserDto } from './dto/user.dto';
+import { UpdateUserDto } from './dto/update-user.dto';
+import { DeleteUserDto } from './dto/delete-user.dto';
+
+// Test suite for the UsersController
+describe("UsersController Unit Tests", () => {
```
Bialogs, line 11: Throughout these controller tests it would be good to see some negative testing in addition to the positive tests that you've started on. For example - test what happens when an ID is provided that is not found. Test what happens when missing required parameters during create, etc.

src/users/users.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    // Tests the findById function
+    it("should findById", () => {
+        usersController.findById(consts.ID);
+        expect(usersService.findById).toHaveBeenCalledWith(consts.ID);
+        expect(usersService.findById).toBeCalledTimes(1);
+    });
+
+    // Tests the create function
+    it("should create", async () => {
```
Bialogs, line 44: Attempt to create with all combinations of required/not required attributes.

src/users/users.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        usersController.create(consts.CREATE_USER_DTO_TEST_OBJ);
+        expect(usersService.create).toHaveBeenCalledWith(consts.CREATE_USER_DTO_TEST_OBJ);
+        expect(usersService.create).toBeCalledTimes(1);
+    });
+
+    // Tests the update function
+    it("should update", () => {
```
Bialogs, line 51: Attempt to update with all combinations of required/not required attributes.

src/users/users.controller.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+                        update: jest.fn((number, UpdateUserDto) => UserDto),
+                        remove: jest.fn((number, DeleteUserDto) => UserDto)
+                    })
+                }
+            ],
+        }).compile();
+
+        usersService = module.get<UsersService>(UsersService);
+        usersController = module.get<UsersController>(UsersController);
+    });
```
Bialogs, line 34: What do you think about setting up the controller tests in [this](https://github.com/nestjs/nest/blob/master/sample/01-cats-app/src/cats/cats.controller.spec.ts) way? Using `spyOn` and calling `mockImplementation`. This would allow the assertion of return value types and properties which seems important to me.

src/users/users.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+describe("UsersService Unit Tests", () => {
+    let usersService: UsersService;
+    let module: TestingModule;
+    // Used for the remove() function
+    let userID;
+
+    beforeAll(async () => {
+        module = await Test.createTestingModule({
+            imports: [
+                SequelizeModule.forRoot({
```
Bialogs, line 17: Need to use the database configuration that is already present in the application.

src/users/users.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // Checks to make sure the exists function throws exception for non-existant user
+    it("should throw NotFoundException from exists function", () => {
+        expect(() => { usersService.exists(consts.NULL_USER) }).toThrow("User with given id not found");
+    });
+
+    // Tests the create function
+    it("should create", async () => {
+        const user = await usersService.create(consts.CREATE_USER_DTO_TEST_OBJ);
+        expect(user.email).toEqual(consts.USER_ONE_DTO.email);
```
Bialogs, line 42: Validate all object properties

src/users/users.service.spec.ts
```diff
@@ -0,0 +1,91 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { UsersService } from './users.service';
+import { User } from './user.model';
+import { getModelToken, SequelizeModule } from '@nestjs/sequelize';
+import * as consts from '../../src/test.constants';
+import { Repository } from 'sequelize-typescript';
+
+describe("UsersService Unit Tests", () => {
```
Bialogs, line 8: Needs negative testing for find methods similar to the create/remove/update methods.

src/users/users.controller.spec.ts
samuelsunvold: The first way I did it mocked the function and made sure it returned a type of "UserDTO." I updated it so that the mock function returns the test userdto constant I made, and then I check that the function in users.controller returns that same constant. While it's a different way of doing it then the way you linked, it should accomplish the same thing.

samuelsunvold: I updated the way I did it to run like the way in the link you posted. It's still the same way I originally mocked the users service functions, but now it specifically returns one of the test userdto's I created and checks that the user controller functions return the correct value

src/users/users.controller.spec.ts
samuelsunvold: These tests should be e2e tests since they test the pipelines

src/users/users.controller.spec.ts
samuelsunvold: These tests should be e2e tests since they test the pipelines

src/users/users.service.spec.ts
```diff
@@ -36,53 +42,234 @@ describe("UsersService Unit Tests", () => {
...
+            expect(userdtoArray[userdtoArray.length-1].title).toEqual(consts.USER_DTO_ARRAY[0].title);
+            expect(userdtoArray[userdtoArray.length-1].organization).toEqual(consts.USER_DTO_ARRAY[0].organization);
+
+            /* Expect that the dates equals each other for the createdAt value. Can't just see
+                if the createdAt equal eachother due to time created being off by a couple of seconds */
+            expect(userdtoArray[userdtoArray.length-1].createdAt.getDate()).toEqual(consts.USER_DTO_ARRAY[0].createdAt.getDate());
+
+            /* The id of the test constant is undefined, so if 
+                findAll is successful, the id's should not be equal */
+            let idFlag: boolean;
```
Bialogs, line 133: I think that the tests that are modeled after this (checking to see if this flag is false with toBe can be simplified with https://jestjs.io/docs/en/expect#tobefalsy

src/users/users.service.spec.ts
```diff
@@ -36,53 +42,234 @@ describe("UsersService Unit Tests", () => {
...
+        });
 
-    // Tests the findByEmail function
-    it("should findByEmail", async () => {
-        const user = await usersService.findByEmail(consts.TEST_USER.email)
-        expect(user.email).toEqual(consts.USER_ONE_DTO.email);
+        // Tests the create function with dto that has no passwordConfirmation field
+        it("should test create function with missing password confirmation field", async () => {
+            expect.assertions(1);
+            try {
```
Bialogs, line 107: The try catch tests could be simplified by using https://jestjs.io/docs/en/using-matchers#exceptions

PR #10, opened 2020-06-10
Added unit tests for passwords-match.pipe

samuelsunvold:

### DIFF
--- a/src/pipes/passwords-match.pipe.spec.ts
+++ b/src/pipes/passwords-match.pipe.spec.ts
@@ -10,11 +10,14 @@ describe('PasswordsMatchPipe', () => {
     passwordsMatchPipe = new PasswordsMatchPipe();
   });
 
+  it('should make sure that the passwords-match pipe is defined', () => {
+    expect(passwordsMatchPipe).toBeDefined();
+  });
+
   /* Tests that when password and passwordConfirmation match,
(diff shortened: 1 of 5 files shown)


src/pipes/passwords-match.pipe.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  beforeEach(() => {
+    passwordsMatchPipe = new PasswordsMatchPipe();
+  });
+
+  /* Tests that when password and passwordConfirmation match,
+   the same CreateUserDto obj that is passed to the pipeline, is returned */
+  describe('Test Matching Passwords', () => {
+    it('should return the same CreateUserDto', () => {
+      expect(passwordsMatchPipe).toBeDefined();
```
Bialogs, line 17: I don't think there needs to be a check in every test for the pipe to be defined. Maybe just create one test that checks.

src/users/users.service.spec.ts
```diff
@@ -33,6 +33,8 @@ describe('UsersService', () => {
 
     usersService = module.get<UsersService>(UsersService);
     databaseService = module.get<DatabaseService>(DatabaseService);
+
+    jest.setTimeout(10000);
```
Bialogs, line 37: Move this to a global config in package.json

PR #15, opened 2020-06-23
Added role property to CreateUserDto and UpdateUserDto interfaces, up…

samuelsunvold:
…dated test constants, and updated users.service.spec

### DIFF
--- a/src/users/dto/create-user.dto.ts
+++ b/src/users/dto/create-user.dto.ts
@@ -23,6 +23,6 @@ export class CreateUserDto {
   @IsOptional()
   readonly title: string;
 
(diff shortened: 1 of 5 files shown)


src/users/dto/create-user.dto.ts
```diff
@@ -22,4 +22,7 @@ export class CreateUserDto {
 
   @IsOptional()
   readonly title: string;
+
+  @IsOptional()
```
Bialogs, line 26: My only comment is to make role required on creation.

Bialogs:
Can you also include the updates to the create and update method to include role?

PR #16, opened 2020-06-23
E2e tests

samuelsunvold:


test/app.e2e-spec.ts
```diff
@@ -1,20 +1,307 @@
...
+    describe('POST', () => {
+      it('should return 201 status when user is created', async () => {
+        return await request(app.getHttpServer()).post('/users').set('Content-Type', 'application/json').send(CREATE_USER_DTO_TEST_OBJ).expect(HttpStatus.CREATED).then(response => {
+          expect(response.body.createdAt.valueOf()).not.toBe(TEST_USER.createdAt.valueOf());
+          expect(response.body.email).toEqual(TEST_USER.email);
+          expect(response.body.firstName).toEqual(TEST_USER.firstName);
+          expect(response.body.id).toBeDefined();
+          expect(response.body.lastLogin).toEqual(null);
+          expect(response.body.lastName).toEqual(TEST_USER.lastName);
+          expect(response.body.loginCount).toEqual(TEST_USER.loginCount.toString());
```
Bialogs, line 59: I think it would make more sense to expect this to be an explicit `0` because this user was just created.

test/app.e2e-spec.ts
```diff
@@ -1,20 +1,307 @@
...
-    app.close();
+  beforeEach(async () => {
+    await databaseService.cleanAll();
+  });
+
+  describe('/users', () => {
+
+    describe('POST', () => {
+      it('should return 201 status when user is created', async () => {
+        return await request(app.getHttpServer()).post('/users').set('Content-Type', 'application/json').send(CREATE_USER_DTO_TEST_OBJ).expect(HttpStatus.CREATED).then(response => {
```
Bialogs, line 52: Throughout these tests would it make more sense to have the comparisons be against the DTO that you submitted?

test/app.e2e-spec.ts
```diff
@@ -1,20 +1,307 @@
...
+          expect(response.body.createdAt.valueOf()).not.toBe(TEST_USER.createdAt.valueOf());
```
Bialogs, line 53: Maybe assert that this is defined or at a minimum.
Additional would be making sure that the time is within 1 minute or something reasonable.

test/app.e2e-spec.ts
```diff
@@ -1,20 +1,307 @@
...
+          expect(response.body.email).toEqual(TEST_USER.email);
+          expect(response.body.firstName).toEqual(TEST_USER.firstName);
+          expect(response.body.id).toBeDefined();
+          expect(response.body.lastLogin).toEqual(null);
+          expect(response.body.lastName).toEqual(TEST_USER.lastName);
+          expect(response.body.loginCount).toEqual(TEST_USER.loginCount.toString());
+          expect(response.body.organization).toEqual(TEST_USER.organization);
+          expect(response.body.role).toEqual(TEST_USER.role);
+          expect(response.body.title).toEqual(TEST_USER.title);
+          expect(response.body.updatedAt.valueOf()).not.toBe(TEST_USER.updatedAt.valueOf());
```
Bialogs, line 63: Same as the timing comment above

test/app.e2e-spec.ts
```diff
@@ -1,20 +1,307 @@
...
+
+      it('should return 400 status if passwordConfirmation is not provided', async () => {
+        return await request(app.getHttpServer()).post('/users').set('Content-Type', 'application/json').send(CREATE_USER_DTO_TEST_OBJ_WITH_MISSING_PASSWORD_CONFIRMATION_FIELD)
+          .expect(HttpStatus.BAD_REQUEST).then(response => {
+            expect(response.body.message[0]).toEqual('passwordConfirmation should not be empty');
+            expect(response.body.error).toEqual('Bad Request');
+          });
+      });
+
+      it('should return 400 status if invalid email is provided', async () => {
```
Bialogs, line 91: I would like to see another test for missing email as well.

test/app.e2e-spec.ts
```diff
@@ -1,20 +1,307 @@
...
+
+      it('should return 400 status if password is not provided', async () => {
+        return await request(app.getHttpServer()).post('/users').set('Content-Type', 'application/json').send(CREATE_USER_DTO_TEST_OBJ_WITH_MISSING_PASSWORD_FIELD)
+          .expect(HttpStatus.BAD_REQUEST).then(response => {
+            expect(response.body.message[0]).toEqual('password should not be empty');
+            expect(response.body.error).toEqual('Bad Request');
+          });
+      });
+
+      it('should return 400 status if passwordConfirmation is not provided', async () => {
```
Bialogs, line 83: I would like to see a test with a password that does not meet the complexity requirements.

Bialogs:
An additional thing that you could change is create the organizational structure within the `test/` folder by splitting out some of the tests into different files when logical.

For example create some dedicated authentication tests under a `authn.e2e-spec.ts` and move the tests for user stuff to `users.e2e-spec.ts` instead of all in `app`.

Then in `users.e2e-spec.ts` you can create a way to use the `beforeEach` to log in users for update, delete, get tests.

Bialogs:
Going to merge this after the tests pass from merging #17

PR #22, opened 2020-07-01
Config service unit tests

samuelsunvold:


src/config/config.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      new ConfigService();
+      expect(consoleSpy).toHaveBeenCalledWith('Unable to read configuration file `.env`!')
+      expect(consoleSpy).toHaveBeenCalledWith('Does the file exist and is it readable by the current user?')
+      expect(consoleSpy).toHaveBeenCalledWith('Falling back to default or undefined values!')
+    });
+  });
+
+  describe('Tests the get function when .env file does exist', () => {
+    beforeAll(() => {
+      // Restore the fs binding to the real file system
```
Bialogs, line 31: Don't do this. Create the file how you expect it to look in the mock fs and use it for everything

src/config/config.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      mock.restore();
+    });
+
+    it('should return the correct database name', () => {
+      const configService = new ConfigService();
+      expect(configService.get('DATABASE_NAME')).toEqual('heimdallts_jest_testing_service_db');
```
Bialogs, line 37: with regards to the above comment . this will not work on everyone's system unless they have set up their environment file the same as you. Relying on the mock-fs instead ensures that problem doesn't crop up

src/config/config.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      const configService = new ConfigService();
+      expect(configService.get('INVALID_VARIABLE')).toBe(undefined);
+    });
+  });
+
+  describe('Tests for thrown errors', () => {
+    it('should throw an EACCES error', () => {
+      mock({
+        '.env': mock.file({
+          content: 'DATABASE_NAME=heimdallts_jest_testing_service_db',
```
Bialogs, line 50: This is the exact strategy to use in the above tests.

test/constants/users-test.constant.ts
```diff
@@ -9,6 +9,15 @@ export const ID = 7;
 
 export const MINUTE_IN_MILLISECONDS = 60000;
 
+export const ENV_MOCK_FILE = 'HEIMDALL_SERVER_PORT=8000\n' +
```
Bialogs, line 12: Can you create a new constants file named env-test.constant.ts

src/config/config.service.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+          mode: 0o000, // Set file system permissions to none
+        })
+      });
+      expect(() => new ConfigService()).toThrowError();
```
Bialogs, line 69: Can you assert that EACCESS is being thrown specifically so we know that the full context is being passed?

PR #25, opened 2020-07-06
Evaluation controller

samperry7:
Still a lot of work to do inside the controller file, also not sure if something needs to be modified in the service.ts file. Atom was yelling about missing the EvaluationsService that I'm trying to import at the beginning.

### DIFF
--- a/src/evaluations/evaluations.controller.ts
+++ b/src/evaluations/evaluations.controller.ts
@@ -1,0 +1,36 @@
+import { Controller, Get, Post, Put, Delete, Param, Body, UseGuards } from '@nestjs/common';
+import { EvaluationDto } from './dto/evaluation.dto';
+import { CreateEvaluationDto } from './dto/create-evaluation.dto';
+import { UpdateEvaluationDto } from './dto/update-evaluation.dto';
+import { EvaluationsService } from './evaluations.service';
+import { JwtAuthGuard } from '../guards/jwt-auth.guard';
+
+@Controller('evaluations')
+@UseGuards(JwtAuthGuard)
+export class EvaluationsController {
+  constructor(private evaluationsService: EvaluationsService) {}
+  @Get(':id')
+  async findById(@Param('id') id: number): Promise<EvaluationDto> {
+    return this.evaluationsService.findById(id);
+  }
+
+  @Get()
+  async findAll(): Promise<EvaluationDto[]> {
+    return this.evaluationsService.findAll();
+  }
+
+  @Post()
+  async create(@Body() createEvaluationDto: CreateEvaluationDto): Promise<EvaluationDto> {
+    return this.evaluationsService.create(createEvaluationDto);
+  }
+
+  @Put(':id')
+  async update(@Param('id') id: number, @Body() updateEvaluationDto: UpdateEvaluationDto): Promise<UpdateEvaluationDto> {
+    return this.evaluationsService.update(id, updateEvaluationDto);
+  }
+
+  @Delete(':id')
+  async remove(@Param('id') id: number): Promise<EvaluationDto> {
+    return this.evaluationsService.remove(id)
+  }
+}
--- a/src/evaluations/evaluations.module.ts
+++ b/src/evaluations/evaluations.module.ts
@@ -1,13 +1,13 @@
 import {Module} from '@nestjs/common';
 import {SequelizeModule} from '@nestjs/sequelize';
-import {EvaluationsService} from './evaluations.service';
 import {Evaluation} from './evaluation.model';
+import {EvaluationsService} from './evaluations.service';
+import {EvaluationsController} from './evaluations.controller';
 
 @Module({
-  imports: [
-    SequelizeModule.forFeature([Evaluation])
-  ],
+  imports: [SequelizeModule.forFeature([Evaluation])],
   providers: [EvaluationsService],
+  controllers: [EvaluationsController],
   exports: [EvaluationsService]
 })
 export class EvaluationsModule {}
--- a/src/evaluations/evaluations.service.ts
+++ b/src/evaluations/evaluations.service.ts
@@ -1,6 +1,9 @@
-import {Injectable} from '@nestjs/common';
+import {Injectable, NotFoundException} from '@nestjs/common';
 import {InjectModel} from '@nestjs/sequelize';
 import {Evaluation} from './evaluation.model';
+import {EvaluationDto} from './dto/evaluation.dto';
+import {CreateEvaluationDto} from './dto/create-evaluation.dto';
+import {UpdateEvaluationDto} from './dto/update-evaluation.dto';
 
 @Injectable()
 export class EvaluationsService {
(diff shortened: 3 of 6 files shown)


src/evaluations/evaluations.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+export class EvaluationsController {
+  constructor(private evaluationsService: EvaluationsService) {}
+  @UseGuards(JwtAuthGuard)
+  @Get(':id')
+  async findById(@Param('id') id: number): Promise<EvaluationDto> {
+    return this.evaluationsService.findById(id);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Get(':id')
```
Bialogs, line 18: A method to find all should not take a parameter

src/evaluations/evaluations.controller.ts
```diff
@@ -0,0 +1,39 @@
+import { Controller, Get, Post, Put, Delete, Param, Body, UseGuards } from '@nestjs/common';
+import { EvaluationDto } from './dto/evaluation.dto';
+import { CreateEvaluationDto } from './dto/create-evaluation.dto';
+import { UpdateEvaluationDto } from './dto/update-evaluation.dto';
+import { EvaluationsService } from './evaluations.service';
+import { JwtAuthGuard } from '../guards/jwt-auth.guard';
+
+@Controller('evaluations')
+export class EvaluationsController {
```
Bialogs, line 9: Please look up in the nestjs documentation the way to configure guards at the top level of a controller so you do not have to repeat the `@UseGuards` decorator for each func.

src/evaluations/evaluations.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+    return this.evaluationsService.findById(id);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Get(':id')
+  async findAll(@Param('id') id: number): Promise<EvaluationDto> {
+    return this.evaluationsService.findAll(id);
+  }
+
+  @Post()
```
Bialogs, line 23: Post needs a JwtAuth guard but that should be handled by my earlier comment

src/evaluations/evaluations.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+  async create(@Body() createEvaluationDto: CreateEvaluationDto): Promise<EvaluationDto> {
+    return this.evaluationsService.create(createEvaluationDto);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Put(':id')
+  async update(@Param('id') id: number, @Body() updateEvaluationDto: UpdateEvaluationDto) {
```
Bialogs, line 30: Please specify a return type for this function.

src/evaluations/evaluations.controller.ts
```diff
@@ -0,0 +1,39 @@
...
+  constructor(private evaluationsService: EvaluationsService) {}
+  @UseGuards(JwtAuthGuard)
+  @Get(':id')
+  async findById(@Param('id') id: number): Promise<EvaluationDto> {
+    return this.evaluationsService.findById(id);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Get(':id')
+  async findAll(@Param('id') id: number): Promise<EvaluationDto> {
```
Bialogs, line 19: This is the wrong return type for this function

src/evaluations/evaluations.module.ts
```diff
@@ -1,8 +1,9 @@
 import {Module} from '@nestjs/common';
 import {EvaluationsService} from './evaluations.service';
+import {EvaluationsController} from './evaluations.controller';
 
 @Module({
-  providers: [EvaluationsService],
+  providers: [EvaluationsService, EvaluationsController],
```
Bialogs, line 6: This is wrong. Look at how the other modules specify the controller and/or refresh with the nestjs documentation on controllers and modules.

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
+    return new EvaluationDto(evaluationData);
+  }
+
+  async remove(id: number) {
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
+    await evaluation.destroy();
+    return new EvaluationDto(evaluation);
+  }
+
+  async findById(id: number): Promise<EvaluationDto> {
```
Bialogs, line 41: Need to check to make sure that the evaluation exists.

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
+
+  async update(id: number, updateEvaluationDto: UpdateEvaluationDto) {
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
+    evaluation.version = updateEvaluationDto.version || evaluation.version;
+    const evaluationData = await evaluation.save();
+    return new EvaluationDto(evaluationData);
+  }
+
+  async remove(id: number) {
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
```
Bialogs, line 36: Need to check to make sure that the evaluation exists.

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
+
+  async create(createEvaluationDto: CreateEvaluationDto) {
+      const evaluation = new Evaluation();
+      evaluation.version = createEvaluationDto.version;
+      const evaluationData = await evaluation.save();
+      return new EvaluationDto(evaluationData);
+}
+
+  async update(id: number, updateEvaluationDto: UpdateEvaluationDto) {
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
```
Bialogs, line 29: Need to check to make sure that the evaluation exists.

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
+    evaluation.version = updateEvaluationDto.version || evaluation.version;
```
Bialogs, line 30: look into using the sequelize  `.update` function

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
     private evaluationModel: typeof Evaluation
   ) {}
+
+
+async findAll(): Promise<EvaluationDto[]> {
+    const evaluations = await this.evaluationModel.findAll<Evaluation>();
+    return evaluations.map(evaluation => new EvaluationDto(evaluation));
+  }
+
+  async create(createEvaluationDto: CreateEvaluationDto) {
```
Bialogs, line 21: specify a return type

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
+      const evaluation = new Evaluation();
+      evaluation.version = createEvaluationDto.version;
+      const evaluationData = await evaluation.save();
+      return new EvaluationDto(evaluationData);
+}
+
+  async update(id: number, updateEvaluationDto: UpdateEvaluationDto) {
```
Bialogs, line 28: specify a return type

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,47 @@
...
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
+    evaluation.version = updateEvaluationDto.version || evaluation.version;
+    const evaluationData = await evaluation.save();
+    return new EvaluationDto(evaluationData);
+  }
+
+  async remove(id: number) {
```
Bialogs, line 35: specify a return type

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,56 @@
...
+      return ;
+    }
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
+    await evaluation.destroy();
+    return new EvaluationDto(evaluation);
+  }
+
+  async findById(id: number): Promise<EvaluationDto> {
+    if (this.evaluationModel.findByPk<Evaluation>(id)){
+      return ;
```
Bialogs, line 49: If the object exists it will stop execution of the function here, not performing the find. I  think my previous comment about checking to see if the Evaluation exists for this function was incorrect.

src/evaluations/evaluations.service.ts
```diff
@@ -1,11 +1,56 @@
...
+    }
+    const evaluation = await this.evaluationModel.findByPk<Evaluation>(id);
+    evaluation.version = updateEvaluationDto.version || evaluation.version; //look into sequelize.update function
+    const evaluationData = await evaluation.save();
+    return new EvaluationDto(evaluationData);
+  }
+
+  async remove(id: number): Promise<EvaluationDto> {
+    if (this.evaluationModel.findByPk<Evaluation>(id)){
```
Bialogs, line 39: If the evaluation exists, it will halt execution of this function and not delete the eval.

There are examples of an exists function in the users service...

src/evaluations/evaluations.controller.ts
samperry7: Would this just be ": Promise<UpdateEvaluationDto>"

Bialogs: No because you're not passing any type to Promise

PR #28, opened 2020-07-10
Add tags

Bialogs:
Creates tagging data structure and methods in a service. Full implementation depends on Evaluations because I think the workflow should be to update an Evaluation to add a tag to it, not update a tag to add an evaluation to it.

### DIFF
--- a/src/evaluations/evaluations.module.ts
+++ b/src/evaluations/evaluations.module.ts
@@ -1,7 +1,12 @@
 import {Module} from '@nestjs/common';
+import { SequelizeModule } from '@nestjs/sequelize';
 import {EvaluationsService} from './evaluations.service';
+import {Evaluation} from './evaluation.model';
 
 @Module({
+  imports: [
+    SequelizeModule.forFeature([Evaluation])
+  ],
   providers: [EvaluationsService],
   exports: [EvaluationsService]
 })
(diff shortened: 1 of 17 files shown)


src/evaluations/evaluations.module.ts
```diff
@@ -1,7 +1,12 @@
 import {Module} from '@nestjs/common';
+import { SequelizeModule } from '@nestjs/sequelize';
```
rbclark, line 2: Idk if there is a reason for it, but this is the only import where you have spaces in the brackets.

test/constants/evaluation-tags-test.contant.ts
```diff
@@ -0,0 +1,54 @@
+import {CreateEvaluationTagDto} from '../../src/evaluation-tags/dto/create-evaluation-tag.dto';
+import {EvaluationTag} from '../../src/evaluation-tags/evaluation-tag.model';
+import {EvaluationTagDto} from '../../src/evaluation-tags/dto/evaluation-tag.dto';
+import {UpdateEvaluationTagDto} from '../../src/evaluation-tags/dto/update-evaluation-tag.dto';
+
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
+
+// @ts-ignore
```
rbclark, line 8: Why does this need to be @ts-ignored? All I could find online was:

> A // @ts-ignore comment suppresses all errors that originate on the following line. It is recommended practice to have the remainder of the comment following @ts-ignore explain which error is being suppressed.

Bialogs: It's missing properties from a full EvaluationTag object so this appeases the compiler

Issue #69, opened 2020-08-19
Redesign sidebar to include Results + Profiles


rbclark:
Depends on mitre/heimdall2#65

This is how the sidebar is currently displayed when a user loads a Result and Profile:

<img width="375" alt="Screen Shot 2020-08-19 at 7 32 48 PM" src="https://user-images.githubusercontent.com/2308869/90700087-e4ad9c00-e252-11ea-8e09-5a64397d639b.png">

This has 2 problems:
 - It doesn't provide very good distinction between a Profile and a Result
 - It doesn't do a very good job of showing that comparison View doesn't work with Profiles.

Suggested solution:

<img width="370" alt="Screen Shot 2020-08-19 at 7 38 42 PM" src="https://user-images.githubusercontent.com/2308869/90700341-9947bd80-e253-11ea-88ae-118246641cdb.png">

When a Profile is selected, the Results and Comparison views should be deselected, and the checkboxes next to the Results should be greyed out. This should only allow for 1 profile to be selected at a time.

ejaronne:
Suggestion for hamburger menu:

Views
- Results View
- Results Comparison View
- Profile View

Results (collapsible)

Profiles (collapsible)

Tools

Info

When you click on Results View or Results Comparison View, the Results list expands, you choose see the view building to the right/behind. click off of the menu to the right to collapse the side menu. (maintain state for the selected set of results)

When you click on Profile View, the Profiles list expands, you choose see the view building to the right/behind. click off of the menu to the right to collapse the side menu. (maintain state for the selected set of profiles)

much cleaner!

ejaronne:
"evaluations" should be changed to "results"
"reports" should be changed to "result sets"

ejaronne:
Results
   list
   X compare
Profiles
   list

dmokllm:
Additional information to the above layout:

- when no file is loaded of a certain type, Results or Profiles, still show the drop down and any options
- only one drop down can be open at a time
- expanding a drop down will activate that view (e.g., expanding Profiles dropdown will automatically activate that view)
- when loading a file of a certain type, activate the view for that file.  when loading multiple files of multiple types, maintain the current default behavior of expanding the Results view
- in the above diagram, [https://github.com/mitre/heimdall2/issues/69#issuecomment-691233492](url), Results and Profiles are dropdowns and compare is a toggle to switch between the Results and Comparison view only (not between Profiles and Comparison view)

ejaronne:
[...]

Issue #65, opened 2020-08-19
Create standalone Profile view


rbclark:
The information displayed for Results vs Profiles is different enough to justify creating its own standalone view. This view should reuse many of the components that exist in the Results view.

Amndeep7:
Duplicate: https://github.com/mitre/heimdall2/issues/116

Issue #82, opened 2020-08-10
Update Results page to display Profiles


robthew:
The results page is designed for Evaluations, but it needs to display uploaded Profiles as well.

To handle Profiles, a few changes are needed: 
1. The result counts and charts are not needed, and can be replaced with a chart showing the counts of severities.
2. The control datatable does not need to display the 'Test' section,  but can keep the 'Info' and 'Code' sections.
3. If this Profile is saved in the database, a table of Evaluations performed against this Profile can be listed (based on the SHA256 value).

dmokllm:
The following comment needs to be addressed when addressing this issue: `https://github.com/mitre/heimdall2/pull/126/#discussion_r485155360`

PR #126, opened 2020-08-31
Separation of views based on files uploaded

dmokllm:
This addresses issues #65, #69, and #161 by splitting the Results view into two views (Results and Profiles) and removing dead code.

Since the Comparison view operates on the same files as the Results view, these two views are grouped together.


apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -11,47 +11,113 @@
...
+                <v-list-item-content>
+                  <v-list-item-title>
+                    <div v-if="all_toggled_profiles">Deselect all profiles</div>
+                    <div v-else>Select all profiles</div>
+                  </v-list-item-title>
+                </v-list-item-content>
+              </v-list-item>
+            </div>
+          </v-expansion-panel-content>
+        </v-expansion-panel>
```
rbclark, line 114: The whole `v-expansion-panel` here should be its own component, and it should be reused for the Results and Profiles dropdowns.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -11,47 +11,113 @@
...
+              v-if="
+                visible_profile_files != null &&
+                  visible_profile_files.length > 0
+              "
+            >
+              <ProfileItem
+                v-for="(file, i) in visible_profile_files"
+                :key="i"
+                :file="file"
+              />
```
rbclark, line 97: This seems extremely similar to FileItem, can it just be reused?

apps/frontend/src/views/BaseView.vue
```diff
@@ -50,6 +50,6 @@ const BaseProps = Vue.extend({
 })
 export default class Base extends BaseProps {
   /** Models if the drawer is open */
-  drawer: boolean = false;
+  drawer: boolean = true;
```
rbclark, line 53: This changes the default behavior of the app to open the drawer on load. Is this intended?

apps/frontend/src/views/Compare.vue
```diff
@@ -575,6 +575,43 @@ export default class Compare extends Props {
...
+      }
+    }
+    if (this.file_filter.length > 1) {
+      return (
+        'Compare View (' + this.file_filter.length + ' results selected' + ')'
+      );
+    } else {
+      return 'Compare View (No files selected)';
+    }
+  }
```
rbclark, line 598: This may be slightly simpler

```javascript
  get curr_title(): string {
    let returnText = 'Comparison View'
    if (this.file_filter.length == 1) {
      let file = InspecDataModule.allEvaluationFiles.find(
        f => f.unique_id === this.file_filter[0]
      );
      if (file) {
        returnText += ` (${file.filename} selected)`;
      }
    }
    else {
      returnText += ` (${this.file_filter.length} results selected)`;
    }
    return returnText;
  }
```

rbclark (changes requested):
- [x] Clicking on a single `Profile` deselects all `Results` but doesn't actually switch to Profile view.
- [x] The mouseover text for Results view, Results Comparison View and Profile view are all incorrect.
- [ ] Hide the Profiles dropdown and grey out Profile View when no profiles are loaded
- [ ] Hide the Results dropdown and grey out Results View when no results are loaded
- [x] I only put a few of my suggestions in 1 spot however the same piece of code is duplicated multiple times. It should be updated in all places (or moved to a shared location since it is duplicated).

apps/frontend/src/views/BaseView.vue
dmokllm: Yes, this is intentional.  I had a chat with @ejaronne and the behavior should be the user will explicitly close the sidebar regardless of switching views.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -90,36 +157,115 @@ const SidebarProps = Vue.extend({
...
     HelpModal
   }
 })
 export default class Sidebar extends SidebarProps {
-  //selects/deselects all files. Will select all unless all are already selected
-  toggle_all(): void {
-    if (this.all_toggled) {
-      FilteredDataModule.set_toggled_files([]);
+  // used to toggle v-exapansion-panel
+  file_views = 0;
```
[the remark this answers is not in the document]
dmokllm: This should be an integer as it represents the position of a v-expansion-panel, which is zero-based.  Right now, we only have two expansion panels so 0 and 1 are used.  The number will increase as/if we add more panels.

Maybe an idea to help avoid confusion and increase readability is to do an enumeration.  So, instead of, for instance, 
```file_views = 0```
we could have something like
```file_views = RESULTS_PANEL```

Thoughts?

File: apps/frontend/src/components/global/Sidebar.vue [hunk 6]
```diff
@@ -11,47 +11,113 @@
...
+          </v-expansion-panel-content>
+        </v-expansion-panel>
+        <v-expansion-panel>
+          <v-expansion-panel-header title="View controls' results">
+            <v-list-item>Profiles</v-list-item>
+          </v-expansion-panel-header>
+          <v-expansion-panel-content>
+            <div
+              v-if="
+                visible_profile_files != null &&
```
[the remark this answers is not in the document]
dmokllm: True.  Removed unnecessary check.

apps/frontend/src/views/Compare.vue
dmokllm: I'll use this algorithm for the Profiles, Results, and Compare vue files.

ejaronne:
Small typo when no files are selected to view, remove "have": 
"No files are currently enabled for viewing. Open the  sidebar menu, and ensure that the file(s) you wish to view _**have**_ are  checked."

apps/frontend/src/components/global/sidebaritems/SidebarProfile.vue
```diff
@@ -0,0 +1,136 @@
+<template>
+  <v-list-item @click="select_file_exclusive" :title="file.filename">
+    <v-list-item-action @click="select_file">
```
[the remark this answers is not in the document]
dmokllm: Some instances can be removed but not all.

apps/frontend/src/components/global/Sidebar.vue
dmokllm: These four lines are similar, yes.  Also, the code in SidebarFile and SidebarProfile (which is referenced by FileItem and ProfileItem) are currently similar.  This separation will set up addressing issue #82, where the profile view and results view will differ.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -90,36 +157,115 @@ const SidebarProps = Vue.extend({
...
+    // is there a better way to do this?
+    for (let x = 0; x < files.length; x++)
+      for (let y = 0; y < file_ids.length; y++)
+        if (files[x].unique_id === file_ids[y]) {
+          // remove evalation file
+          file_ids.splice(y, 1);
+          y--;
+        }
+    return file_ids;
+  }
```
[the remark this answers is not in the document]
dmokllm:
```ruby
get selected_profiles(): number[] {
    let file_ids = [...this.selected_file_ids];
    let files = InspecDataModule.allEvaluationFiles;

    return file_ids.filter(file_id =>
      files.every(file => file_id != file.unique_id)
    );
  }
```

apps/frontend/src/components/global/Sidebar.vue
dmokllm: Moving this to be addressed when issue #82 is addressed.

[https://github.com/mitre/heimdall2/issues/82#issuecomment-691059364](url)

apps/frontend/src/components/global/Sidebar.vue
rbclark: The code in SidebarFile and SidebarProfile should not really be different even after the Results view is broken out into Results and Profiles, the only difference should be the path (which can just be an if statement). This should not be duplicated into 2 components.

apps/frontend/src/components/global/sidebaritems/SidebarProfile.vue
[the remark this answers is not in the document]
rbclark: They can all be replaced with `@click.stop` though, `@click.stop="select_file` is the Vue helper that handles
```javascript
    evt.stopPropagation();
    evt.preventDefault();
```

apps/frontend/src/components/global/Sidebar.vue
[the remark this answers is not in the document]
rbclark: I'm not sure I understand the reason for the change to `every`, it seems that would always iterate every file every time, as opposed to `includes` which can short-circuit if it finds a match.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -90,36 +203,92 @@ const SidebarProps = Vue.extend({
...
 
-  //checks if all files are selected
-  get all_toggled(): boolean {
+  // check to see if all evalutions are selected
+  get all_toggled_evaluation(): boolean {
     return (
-      InspecDataModule.allFiles.length ==
-      FilteredDataModule.selected_file_ids.length
+      this.selected_evaluations.length ==
+      InspecDataModule.allEvaluationFiles.length
```
rbclark, line 260: If this is replaced with `FilteredDataModule.selected_evaluations` then the `selected_evaluations` computed property can be removed from Sidebar.vue.

apps/frontend/src/components/global/Sidebar.vue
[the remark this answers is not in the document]
rbclark: Nevermind, every in this case is able to short circuit as well, the only potential advantage to the other way is it doesn't require calling unique_id n^2 times, however javascript may gracefully handle that.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -90,36 +157,115 @@ const SidebarProps = Vue.extend({
...
+  toggle_all_profiles(): void {
+    if (this.all_toggled_profiles) {
+      this.set_selected_files(this.selected_evaluations);
     } else {
-      FilteredDataModule.set_toggled_files(
-        InspecDataModule.allFiles.map(v => v.unique_id)
-      );
+      let files = InspecDataModule.allProfileFiles.map(v => v.unique_id);
+      files.push(...this.selected_evaluations);
+      this.set_selected_files(files);
```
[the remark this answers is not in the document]
dmokllm: Making this change will affect code such as select_file in SidebarFile.  Instead of sending a file to be toggled on or off, an array of one element will need to be created and sent.

Instead, I changed
```ruby
this.set_selected_files(files);
```
to
```ruby
FilterdDataModule.set_toggled_files(files);
```
where files is an array

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -185,8 +185,7 @@ import Vue from 'vue';
 import Component from 'vue-class-component';
 import {EvaluationFile, ProfileFile} from '@/store/report_intake';
 import {InspecDataModule} from '@/store/data_store';
-import FileItem from '@/components/global/sidebaritems/SidebarFile.vue';
-import ProfileItem from '@/components/global/sidebaritems/SidebarProfile.vue';
+import FileLister from '@/components/global/sidebaritems/SidebarFileLister.vue';
```
rbclark, line 174: Preferably call this `FileList` and `SidebarFileList` in order to match the naming of the rest of the components.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -11,47 +11,145 @@
...
+                <v-list-item-content>
+                  <v-list-item-title>
+                    <div v-if="all_toggled_profiles">Deselect all profiles</div>
+                    <div v-else>Select all profiles</div>
+                  </v-list-item-title>
+                </v-list-item-content>
+              </v-list-item>
+            </v-expansion-panel-content>
+          </div>
+        </v-expansion-panel>
```
rbclark, line 146: The `v-expansion-panel` here should be a component, it should be possible to leverage the same component for the `Results` and `Profiles` expansion components.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -11,46 +11,110 @@
...
+          <v-list-item-avatar>
+            <v-icon small>mdi-television-guide</v-icon>
+          </v-list-item-avatar>
+          <v-list-item-content>
+            <v-list-item-title>
+              Profiles View (why is link deactivated?)
+            </v-list-item-title>
+          </v-list-item-content>
+        </v-list-item>
+      </div>
```
rbclark, line 86: This all still needs to be removed correct?

rbclark (changes requested):
Looking over this again, trying to keep the `selected_file_ids` as 1 array in the `FilteredDataModule` is just making a total mess, please go ahead and just make that 2 arrays as part of this PR for simplicity.

aaronlippold:
There also seems to be a mrege conflict now

apps/frontend/src/components/global/Sidebar.vue
dmokllm: Yes, the new UI design removes these completely.

apps/frontend/src/components/global/sidebaritems/DropdownContent.vue
```diff
@@ -0,0 +1,134 @@
...
+              <v-list-item-title>
+                <div v-if="all_toggled_evaluations">
+                  Deselect all {{ text }}
+                </div>
+                <div v-else>Select all {{ text }}</div>
+              </v-list-item-title>
+            </v-list-item-content>
+          </v-list-item>
+        </div>
+      </div>
```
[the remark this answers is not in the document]
dmokllm: I'll look into passing a function to component.

apps/frontend/src/components/global/sidebaritems/DropdownContent.vue
```diff
@@ -0,0 +1,134 @@
...
+      <div v-else>
+        Why is dropdown empty?
+      </div>
```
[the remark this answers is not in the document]
dmokllm: That was my thinking, with the UI mentioned in this PR.  However, with the new UI design, this may be unnecessary.  I think the text you mentioned about no files may be better.

apps/frontend/src/components/global/Sidebar.vue
```diff
@@ -12,39 +12,37 @@
...
+      <AboutModal>
+        <template v-slot:clickable="{on}">
+          <LinkItem key="about" text="About" icon="mdi-information" v-on="on" />
+        </template>
+      </AboutModal>
+      <HelpModal>
+        <template v-slot:clickable="{on}">
+          <LinkItem key="help" text="Help" icon="mdi-help-circle" v-on="on" />
+        </template>
+      </HelpModal>
```
rbclark, line 45: These were all moved to a new location in the master branch, this PR should not add them back in.

rbclark (approved):

aaronlippold:
Yay !! Cheer !!
--------
Aaron Lippold

lippold@gmail.com

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

PR #127, opened 2020-09-01
Adding Dockerbuild and deploy code

rx294:
Signed-off-by: Rony Xavier <rxavier@mitre.org>


Bialogs (changes requested):
Currently trying to build and test this locally.

Running in to two things:
* If the intent is to pull a build image from Docker Hub then it should be namespaced correctly as `mitre/heimdall2` (I pushed a change to address this.)
* Need to be able to run the `npx sequelize-cli db:seed:all` command when the user wants to preload an admin user into the application. This is not currently possible.
* Docker entrypoint script is overwriting JWT secret every time you start the container or execute a run command against it.
* Docker entrypoint script is continually adding the `.env` file with `>>`
* Migrations are run automatically as part of the entrypoint, this is pretty bad because users don't have control of when to migrate.

rx294:
@Bialogs thanks for the review...please see below

> * If the intent is to pull a build image from Docker Hub then it should be namespaced correctly as `mitre/heimdall2` (I pushed a change to address this.)

Thanks

> * Need to be able to run the `npx sequelize-cli db:seed:all` command when the user wants to preload an admin user into the application. This is not currently possible.

Sure I can update the the entrypoint when the `npx sequelize-cli db:seed:all` command is ready possible in a subsequent PR?

> * Docker entrypoint script is overwriting JWT secret every time you start the container or execute a run command against it.

`    export JWT_SECRET=${JWT_SECRET:-$(openssl rand -hex 64)}`
the above code in entrypoint will overwrite JWT_SECRET only if its not provided on docker-compose or .env

> * Docker entrypoint script is continually adding the `.env` file with `>>`
thanks fixed;

> * Migrations are run automatically as part of the entrypoint, this is pretty bad because users don't have control of when to migrate. Migrations should be tied to startup of a container after a new pull.

I don't understand, entrypoint script run only during startup, if you are talking about stability when new images are released, the recommended practice to resolve that is to use docker image version tags in docker-compose 

Also note that the docker .env in the entrypoint will be removed once @rbclark removes the dependency of .env file in the server

rx294:
@Bialogs made the updates we talked about
 - Updated to use yarn
 - Added `db:seed:all` command to entrypoint
 - Added toggles for migration and seedding
 - Ben removed the dependency on .env file for server.. removed that from entrypoint

Bialogs (dismissed):

PR #132, opened 2020-09-03
Monorepo cleanup

rbclark:
Add code-workspace for vscode

This works around lack of multi-root support in vetur: https://github.com/vuejs/vetur/issues/424#issuecomment-630986860

Create interfaces for all DTO's in heimdall-server 

By creating interfaces, these interfaces can be used in the Heimdall frontend and allow assurance that the data being passed back and forth is consistent

Deduplicate tsconfig files

Add additional testing steps to ensure that the backend and frontend package lists still work independently.

Upgrade version of PrismJS and update PrismJS Ruby regex override to new location (this is used to add inspec format support to Prism)


Bialogs (approved):
Needs the lint run.

Otherwise we've gone over this a few times and I'm in agreement.

PR #143, opened 2020-09-09
Add dropdown to Heimdall logo

### LINKED ISSUE #134: Move "Help" and "About" text to the Heimdall Logo
Clicking on the Heimdall Logo in the top right of the app currently does nothing.

Screen Shot 2020-09-03 at 3 08 06 PM

It should instead bring up a dropdown providing the user with a Help and About button. Once that is moved the "Help" and "About" buttons should be removed from the sidebar.

camdenmoors:
Closes #134, adds a dropdown to the Heimdall logo with help/about


Bialogs (changes requested):
- [ ] When hovering over the two new links they highlight in red/orange color. Should be gray and consistent with the rest of the app

- [ ] There are additional space (height, etc.) styles added to the two list items in the drop down. @rbclark and I feel like the drop-down looks better without that additional spacing.

aaronlippold:
The build is failing and there are merge conflicts

aaronlippold:
Given this change we should swap the light/dark icon and the drop down so it is on the far right side I think

rbclark (approved):

Issue #63, opened 2020-08-18
Update Vuetify version to latest


rbclark:
The app is currently on Vuetify 2.1.1. The latest Vuetify (at the time of writing this) is 2.3.9. The ColorHack module currently in the app is the biggest potential blocker to this upgrade. As part of this issue, any changes that need to be done to update/remove the ColorHack module should be performed.

PR #144, opened 2020-09-09
Upgrade vuetify

sjoshi10:
Updrading vuetify broke UploadNexus colors. Needed to specify background color  on v-list and v-tabs. 

Closes https://github.com/mitre/heimdall2/issues/63


aaronlippold:
There are merge conflicts here, please fix and we can merge

rbclark (changes requested):
It seems on the UploadNexus there is some black box behind it now, I noticed that the corners are a different color and when changing windows there's a flash of black now, is that just another element that the color needs to be changed on? Here is a screenshot of it mid-transition.

Screen Shot 2020-09-09 at 4 50 21 PM

PR #162, opened 2020-09-11
Update Results View Data

sjoshi10:
Make Results view data part of the Results Page. This will remove 3 srollbar.


rbclark (approved):
Looks good, will merge after CI finishes re-running.

Issue #89, opened 2020-08-04
Remove external property declarations from classes


rbclark:
We currently declare all of our properties separately which is not necessary since we have `vue-property-decorator`

https://github.com/mitre/heimdall-lite/blob/a2db66c09e2fa6b87984a8f8f3256ba246ee1c77/src/views/BaseView.vue#L35-L53

This chunk can be replaced for example with the following:

```typescript
@Component({
  components: {
    Sidebar,
    Topbar
  }
})
export default class Base extends Vue {
  @Prop({default: 'Heimdall Lite'}) readonly title!: string;
  /** Models if the drawer is open */
  drawer: boolean = false;
}
```

We should go through and do this for all of our Vue components.

See also: https://github.com/kaorun343/vue-property-decorator#-propoptions-propoptions--constructor--constructor---decorator

PR #181, opened 2020-09-16
Add support for loading results and profiles from the Heimdall2 backend

rbclark:
This is the initial pass at adding support for loading heimdall2 results into the frontend.

Issues:

- [x] The same evaluation can be saved multiple times and a new record will be created in the database for each save. (https://github.com/mitre/heimdall2/issues/201)
- [x] Clicking is not disabled once the load button is clicked or a result filename is clicked. This means a user can click it twice and will get 2 files imported. (https://github.com/mitre/heimdall2/issues/201)
- [x] This PR begins to fix https://github.com/mitre/heimdall2/issues/89 however it only fixes it on Props that got in my way when loading mixins.
- [x] This does not include tag editing or any filtering or grouping yet. This just adds the basic data-table containing results. (https://github.com/mitre/heimdall2/issues/200)
- [x] Add testing for bad backend responses. (https://github.com/mitre/heimdall2/issues/199)
- [x] This does not yet handle errors gracefully if the backend is down, need to just show an error and empty table at that point.


DEBUGGING.md
```diff
@@ -0,0 +1,15 @@
+# Debugging Tips
+
+## Backend E2E tests
```
Bialogs, line 3: These are not e2e tests these are integration tests.

Bialogs (changes requested):

Bialogs (approved):

PR #242, opened 2020-09-25
Remove password peek from all pages but signup

### LINKED ISSUE #225: Only allow users to "Peek Password" on the Signup page
If possible, don't provide the ability to see the entered password if auto-filled by browser. While I had my browser cache my password, I didn't need to be able to see if what was populated was indeed the correct password. And neither should anyone else! 

This was further exacerbated because my browser kept offering the same login credentials to the Splunk and S3 settings areas, all with ability to view the password again. A little too helpful.

For signing up, sure it is very helpful. But not much after that. If you can distinguish between hand-typed (allow display) vs auto-filled (don't allow display), that would be a fair compromise for now.

camdenmoors:
Closes #225, removes password peek from all pages except Signup


rbclark (changes requested):
I believe we made the same icon fixes on both of our branches, it seems this might need to be rebased now.

apps/frontend/src/components/global/upload_tabs/aws/AuthStepBasic.vue
```diff
@@ -11,11 +11,9 @@
       <v-text-field
         :value="secret_token"
         label="User Account Secret Token"
+        type="password"
         :rules="[req_rule]"
-        :append-icon="show_secret ? 'mdi-eye' : 'mdi-eye-off'"
-        :type="show_secret ? 'text' : 'password'"
         @input="change_secret_token"
-        @click:append="show_secret = !show_secret"
```
rbclark, line 18: Since `show_secret` is no longer used, the data field can be removed from this Component.

apps/frontend/src/views/Login.vue
```diff
@@ -28,12 +28,10 @@
                     id="password_field"
                     v-model="password"
                     :error-messages="passwordErrors"
+                    type="password"
                     name="password"
                     label="Password"
                     prepend-icon="mdi-lock"
-                    :type="showPassword ? 'text' : 'password'"
-                    :append-icon="showPassword ? 'mdi-eye' : 'mdi-eye-off'"
-                    @click:append="showPassword = !showPassword"
```
rbclark, line 36: Since `show_secret` is no longer used, the data field can be removed from this Component.

rbclark (changes requested):

apps/frontend/src/views/Login.vue
camdenmoors: I've removed showPassword from Login.vue already, are you referring to something else?

rbclark: It unfortunately won't let me comment on the correct line, on `AuthStep.vue` and `AuthStepBasic.vue` there is still a `show_secret: boolean = false;` that is unused.

https://github.com/mitre/heimdall2/blob/35bbbfc7980c10f60cd6173e759a6d694d3f4e8e/apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue#L68
https://github.com/mitre/heimdall2/blob/35bbbfc7980c10f60cd6173e759a6d694d3f4e8e/apps/frontend/src/components/global/upload_tabs/aws/AuthStepBasic.vue#L73

rbclark (approved):

PR #252, opened 2020-09-29
Switch to Vuetify Snackbar Component

### LINKED ISSUE #163: Remove Toast plugin in favor of Vuetify snackbar components
https://vuetifyjs.com/en/components/snackbars/

Backend tests will need to be updated.

camdenmoors:
Closes #163, adds Snackbar component and replaces vue-toasted.


apps/frontend/src/store/snackbar.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.error = false;
+    this.message = message;
+    this.show = true;
+  }
+  @Mutation
+  failure(message: string) {
+    this.error = true;
+    this.message = message;
+    this.show = true;
+  }
```
rbclark, line 31: These should be `@Action` not `@Mutation` and they should call the mutations and not set the values directly.

apps/frontend/src/components/global/UploadNexus.vue
```diff
@@ -105,7 +106,7 @@ export default class UploadNexus extends mixins(ServerMixin) {
   // Handles change in tab
   selected_tab(new_tab: string) {
     this.active_tab = new_tab;
-    this.$toasted.clear();
+    SnackbarModule.SET_VISIBILITY(false);
```
rbclark, line 109: There should probably be a `visible()` Action on the Vuex store, directly calling Mutations here seems like potentially a bad idea, not sure if I know all of the implications but the general idea is:

> Actions are similar to mutations, the differences being that:
> 
>     Instead of mutating the state, actions commit mutations.
>     Actions can contain arbitrary asynchronous operations.

rbclark (changes requested):

apps/frontend/src/components/global/Snackbar.vue
```diff
@@ -0,0 +1,59 @@
...
+import Component from 'vue-class-component';
+import Vue from 'vue';
+
+@Component({})
+export default class Snackbar extends Vue {
+  get show(): boolean {
+    return SnackbarModule.show;
+  }
+  set show(visibility) {
+    SnackbarModule.SET_VISIBILITY(visibility);
```
rbclark, line 50: This should use `SnackbarModule.visibility` and not the mutation directly.

apps/frontend/src/components/global/Snackbar.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <v-snackbar v-if="show" id="info-snackbar" v-model="show" timeout="10000">
+    <template v-if="error">
+      <template v-if="message" id="snackbar-message"
+        >ERROR: {{ message }}</template
```
rbclark, line 5: The old Toasted display was modified to not include `ERROR:` in front of everything. It might be best to remove the `ERROR:` from the message below as well.

apps/frontend/src/store/snackbar.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.context.commit('SET_ERROR', false);
+    this.context.commit('SET_MESSAGE', message);
+    this.context.commit('SET_VISIBILITY', true);
+  }
+  @Action
+  failure(message: string) {
+    this.context.commit('SET_ERROR', true);
+    this.context.commit('SET_MESSAGE', message);
+    this.context.commit('SET_VISIBILITY', true);
+  }
```
rbclark, line 37: I believe `this.context.commit('SET_ERROR', false);` can be replaced with `this.SET_ERROR(false);`, etc. This is how it is done in `server.ts` and it seems to function fine.

rbclark (changes requested):

apps/frontend/src/components/global/Snackbar.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <v-snackbar v-if="show" id="info-snackbar" v-model="show" timeout="10000">
```
rbclark, line 2: It may be better to display this on top of the page instead of the bottom to match the existing Toasted configuration.

rbclark (changes requested):
One additional comment, it is a little weird that the Login and Signup pages have a "Report Error" even though those components will very commonly display non server-related errors. It may be best to just have those two use cases not display the Report Error text.

Screen Shot 2020-09-30 at 4 06 22 PM
Screen Shot 2020-09-30 at 4 06 04 PM

rbclark (approved):

PR #260, opened 2020-09-30
Fix HelpFooter maximum width

### LINKED ISSUE #73: At certain widths, file select modal footer becomes not-full-width
Screen Shot 2019-11-05 at 6 32 31 PM

camdenmoors:
Closes #73. Changes HelpFooter maximum width to 100%.


rbclark (changes requested):
It is possible to solve this without a custom style, using `<v-container fluid`.

rbclark (approved):

PR #313, opened 2020-10-09
Create Collapsible Search-bar on Mobile

### LINKED ISSUE #254: [BUG] Searchbar does not collapse well on small displays
**Describe the bug**
The searchbar does not handle the small displays well:
Screen Shot 2020-09-29 at 12 54 11 PM

**To Reproduce**
1. Go to heimdall-lite on a small display

**Expected behavior**
The searchbar should behave closer to how https://piaf-vue.coloredstrategies.com/app/dashboards/default does on small displays. (become a button that when clicked takes over the whole topbar)

camdenmoors:
Closes #254, turns the search bar into a button when on displays with a width smaller than 600px.


Bialogs (changes requested):
- When collapsing to mobile view, the clear filter button goes away.

- When clicking on search on < 600px screens, you cannot exit the search dialog unless you click into it first and then click somewhere else. Expected behavior would be that the search will always collapse back into a button when clicking anywhere else (this would also apply the search filter if anything was typed).

Bialogs:
As we discussed:

- Provide method for clearing the search text filter without clearing all filters

- Display clear all filters button when search bar is not expanded.

Bialogs (approved):

aaronlippold:
On my mobile device the tree map does not seem to be loading in this PR and
also the hamburger menu when switching between the comparison view and the
results view seems to be acting a little funky on the mobile device. Lastly
the profile info model seems to be over scanning the display area and it's
going outside the bounds of the application.

PR #314, opened 2020-10-09
Fix dropdown arrow on help/about dropdown

camdenmoors:
Fixes the help/about dropdown displaying weird on smaller displays.


rbclark (changes requested):
This seems to cause the icon to push very hard to the right of the screen, at least in firefox. It may be better to fix this by adding the `white-space: nowrap;` style to the class containing the logo and the down arrow icon.

PR #330, opened 2020-10-13
Handle Errors on Registration Screen

### LINKED ISSUE #205: [BUG] Registration frontend form does not handle all errors from the backend correctly
**Describe the bug**
The backend sometimes returns `data.message` and other times `data.messages` to the signup form. This causes for less-than-helpful errors when a user tries to perform some actions such as registering with an email that is already taken.

**To Reproduce**
1. Go to the registration page
2. Register with an email that is already taken
3. Receive error `ERROR: An unidentified error has occured, if ...` since the backend responds with `messages` instead of `message.

**Expected behavior**
All the different errors returned from the backend should be consistently formatted so that the frontend can easily display them to users. It is currently returned sometimes as a string and sometimes as an array, and sometimes as an Object.

It would also be preferred if the backend returned in a more consistent manner for ease of error handling.

### Some example responses:

**When the signup form is completely blank:**
```
{"statusCode":400,"message":["email should not be empty","email must be an email","password should not be empty","passwordConfirmation should not be empty"],"error":"Bad Request"}
```

**When email is already taken:**
```
{"statusCode":500,"error":"Internal Server Error","messages":[{"email":"email must be unique"}]}
```

**When password confirmation does not match:**
```
{"statusCode":400,"message":"Passwords do not match","error":"Bad Request"}
```

camdenmoors:
Closes #205, Changes `UniqueConstraintError` to return `message` string array to match [other NestJS errors](https://docs.nestjs.com/exception-filters). This also allows the frontend to differentiate between an error array and an error string.


rbclark:
I'll defer to @Bialogs on reviewing this one once CI is passing.

PR #341, opened 2020-10-14
Change How Registration Errors are Handled

### LINKED ISSUE #205: [BUG] Registration frontend form does not handle all errors from the backend correctly

camdenmoors:
Closes #205, changes how error messages are sent for email uniqueness, allows the frontend to differentiate between error arrays and error strings returned by server, and removes a duplicate test for password complexity.

Moved to new PR due to multiple issues with linting/tests.


apps/backend/test/users.e2e-spec.ts
```diff
@@ -377,22 +374,6 @@ describe('/users', () => {
           .expect(HttpStatus.UNAUTHORIZED);
       });
 
-      it('should return 400 status when password does not meet complexity requirements', async () => {
```
Bialogs, line 380: So I'm fine with removing this test but shouldn't the API return *some* status if password does not meet the requirements? What I mean is we should still have a test for this scenario, right?

Bialogs (changes requested):

Bialogs (approved):

PR #344, opened 2020-10-16
Update the evalutions index route to exclude the actual evaluation da…

### LINKED ISSUE #331: Optimize result loading from the database
When the frontend of the application loads results from the backend, all results (including the full data field) are returned. This makes for a slow loading experience for users, and will only get significantly worse as the database size grows. With only 17 results the response is nearly 22MB. In order to speed this up the application needs to return all data EXCEPT the `data` field, and then the `data` should be loaded as a separate request when the user actually chooses a file.

Bialogs:
…ta for quicker loading.

Closes #331


rbclark (changes requested):
Seems to be failing tests.

PR #405, opened 2020-10-28
Fix regression where donut chart compliance percentages were not show…

### LINKED ISSUE #397: Replace compliance % to center of comparison status donut
In 2.3.1, this is what comparison status donuts look like. Note that the inner part doesn't have the count of total tests (controls), but rather the % compliance:
image

In 2.3.2, this inadvertently fell back showing the counts in the middle, which was not the intent for this view.

Bialogs:
…n on compare view.

Closes #397


rbclark (approved):

ejaronne:
Looks good, thanks!
image

PR #467, opened 2020-11-10
Fix showing "None" on some ControlRows

### LINKED ISSUE #399: [BUG] Inspec Profile view should use Severity instead of Impact for Results View Data
**Describe the bug**

**To Reproduce**
1. Load the attached json [output.txt](https://github.com/mitre/heimdall2/files/5453259/output.txt)
2. Search for `V-58157`
2. Notice that the severity shows as None. This is due to the way that the inspec json command parses profiles when generating a `profile.json`. 
3. Go to the `Code` tab for this control.
4. See that the impact is in fact what is being displayed (which does not work properly on profile.json files).

**Expected behavior**
The `tag "severity": "medium"` should instead be what is displayed for profiles, and impact should be displayed for results.

**Screenshots**
Screen Shot 2020-10-28 at 12 14 10 PM

Screen Shot 2020-10-28 at 12 14 38 PM

camdenmoors:
Closes #399, uses `control.data.tags.severity` as it is always defined.


rbclark (changes requested):
This changes the value for profiles as well as results. Only profiles should use the `severity` tag. This causes results that legitimately have an impact of 0 (N/A) to show with a higher severity. (Nginx Clean sample pictured)

Before:
Screen Shot 2020-11-10 at 10 01 00 AM

After:
Screen Shot 2020-11-10 at 10 01 31 AM

rbclark:
Additionally the Column title should change names from "Severity" -> "Impact" on this one when we are in results view. When in profile view it should still say "Severity". Basically the column should should be called Impact on Results view and display the impact, and called Severity on Profile view and display the severity.

ejaronne:
thinking about this... do overlay profile jsons work at all, and if so many times folks set impact to 0.0 on a control, but don't or don't know to touch the severity... Philosophically I wonder what should show. If we show the original severity folks trying to explain the design of their overlay might be confused/frustrated...  not sure what the correct approach should be...

rbclark:
> thinking about this... do overlay profile jsons work at all, and if so many times folks set impact to 0.0 on a control, but don't or don't know to touch the severity... Philosophically I wonder what should show. If we show the original severity folks trying to explain the design of their overlay might be confused/frustrated... not sure what the correct approach should be...

I don't see why overlay profile jsons wouldn't work. Once users actually run the profile and view their results this will still show their overridden severity. It may be necessary to train users to override the severity tag as well as the impact if they want the profile to display in inspec correctly, since there is no way for us to use the impact for a profile due to the way profiles are generated.

rbclark (approved):

ejaronne:
Seems to work thanks!

PR #468, opened 2020-11-10
Add Update Notifications

### LINKED ISSUE #374: Create Update Workflow
No

For the backend to check for new releases from [here](https://api.github.com/repos/mitre/heimdall2/releases/latest). Only admin users will be able to update.

None

Checks for updates on launch and will prompt the user to update (perhaps using a snackbar).

Requested by @aaronlippold

camdenmoors:
Closes #374, notifies the user using a snackbar when there is a newer version of Heimdall available.


rbclark:
Since administrators are the only ones who are able to update the application, it may not make sense to show this to everyone.

Also, is there a need to hit the backend for this or can the frontend just reach out to the Github API directly?

camdenmoors:
> Also, is there a need to hit the backend for this or can the frontend just reach out to the Github API directly?

Due to CORS and other protections it seems more robust to make the request on the backend

rbclark:
> > Also, is there a need to hit the backend for this or can the frontend just reach out to the Github API directly?
> 
> Due to CORS and other protections it seems more robust to make the request on the backend

On the Github API side they allow all origins. Since you're hitting an API I can't think of too many situations where it would be blocked.
 
Screen Shot 2020-11-10 at 10 37 13 AM

rbclark:
Also another huge advantage to doing this client side would be the ability to alert lite users that their version is out of date.

apps/frontend/src/store/snackbar.ts
```diff
@@ -20,25 +21,44 @@ export interface ISnackbarState {
   name: 'SnackbarModule'
 })
 export class Snackbar extends VuexModule {
+  customText: string | undefined = 'owo';
```
rbclark, line 24: Not sure I understand the point of the customText. It seems to only be set and never used.

camdenmoors: Oh sorry, that was for debugging and I never removed it.

camdenmoors: This was intended to add a button with a custom url to the releases page, customText being "View", and url being the link to the releases page

rbclark (approved):

PR #556, opened 2020-11-24
Add Admin Panel

### LINKED ISSUE #97: Component for the root user to reset users password
When in server mode, an admin role user can reset a normal user password through the update component.

rbclark:
This includes functionality for changing all of an existing users information (including changing their password and updating their role).

Closes #97

This PR only provides MVP, there are still 2 additional follow-up issues based on this implementation:
 - https://github.com/mitre/heimdall2/issues/563
 - https://github.com/mitre/heimdall2/issues/564

Todo:

- [x] Remove Load button from Admin page
- [x] Show "Edit" and "Delete on mouseover of Edit and Delete buttons in admin panel.
- [x] Redirect users so non-admins cannot see the admin panel.


camdenmoors:
I think that something should prevent the user from deleting the admin account if it's the only one available, to prevent the user from accidentally locking themselves out of the app and having to re-seed.

ejaronne:
Looks good, however having an active load button is not part of the use case, and would be better if it is possible to be removed. Also, it would be nice to have hover-text "Edit" for the pencil icon and "Delete User" for the trash icon.

Bialogs (approved):
API looks good

PR #562, opened 2020-11-25
Switch to Cypress Tests

### LINKED ISSUE #471: Switch from jest-puppeteer to Cypress.io
Our `jest-puppeteer` tests have becoming more and more unreliable. Additionally, jest-puppeteer is [currently not maintained](https://github.com/smooth-code/jest-puppeteer#%EF%B8%8F-looking-for-maintainers). https://www.cypress.io/ seems a lot cleaner, is not selenium based, and seems like a great alternative. 

I've pushed some very light initial WIP stuff to the [switchToCypress](https://github.com/mitre/heimdall2/tree/switchToCypress) branch.

Relevant links:

https://activeviam.com/blog/puppeteer-vs-cypress-best-end-end-testing/
https://www.cypress.io/blog/2019/11/20/drastically-simplify-your-testing-with-cypress-github-action/

camdenmoors:
Closes #471. Switches to [Cypress](cypress.io) testing suite for integration tests.


.eslintrc.js
```diff
@@ -0,0 +1,27 @@
...
+    node: true,
+  },
+  rules: {
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-unused-vars': ["warn", { "argsIgnorePattern": "^_" }],
+    'no-return-await': 'warn',
+    'no-throw-literal': 'warn'
+  },
+};
```
rbclark, line 27: Is this file still required? It looks the same as the one in the `test/` directory.

test/support/helpers/DatabaseHelper.ts
```diff
@@ -0,0 +1,12 @@
+export default class DatabaseHelper {
+  createAdmin(): string {
+    const re = /[\n\r].*New administrator password is:s*([^\n\r]*)/;
+    let adminPassword = '';
+    cy.exec(
+      'sudo systemctl restart postgresql; sleep 2; yarn backend sequelize-cli db:drop; sleep 2; yarn backend sequelize-cli db:create; sleep 2; yarn backend sequelize-cli db:migrate; sleep 2; yarn backend sequelize-cli db:seed:all'
```
rbclark, line 6: Shouldn't need to restart postgresql here. I'm fine with the database being created and migrated before the tests run, and having this just seed the database and get the admin account over and over for every test.

.eslintrc.js
camdenmoors: Not needed, removed

apps/frontend/src/components/global/Topbar.vue
```diff
@@ -41,11 +44,9 @@ export default class Topbar extends mixins(ServerMixin) {
...
   }
-}
-</script>
 
-<style scoped>
-.clamped {
-  overflow: hidden;
+  truncate(text: string, stop: number, clamp: string | null) {
+    return text.slice(0, stop) + (stop < text.length ? clamp || '...' : '');
+  }
```
rbclark, line 50: Can't this be achieved using `text-overflow: ellipsis;`?

.github/workflows/e2e-ui-tests.yml
```diff
@@ -45,12 +45,29 @@ jobs:
...
        run: |
         yarn backend sequelize-cli db:create
         yarn backend sequelize-cli db:migrate
+        yarn backend sequelize-cli db:seed:all
 
-     - name: Build frontend and server for E2E UI tests
-       run: yarn run build
+     - name: Build Heimdall
+       env:
+        NODE_ENV: development
```
rbclark, line 57: Does the environment need to be set for every stage? Couldn't it just be set at the top? Also is there a reason that the build needs to happen in dev mode?

rbclark (changes requested):
Does this remove the jest-puppeteer package since it should no longer be required?

.github/workflows/e2e-ui-tests.yml
camdenmoors: The tests need to be seeded in test ENV to seed the policies into the database, and the site needs to be built in development mode as building it in test mode gives this: 
image

Global env should work as test, with only dev as build, though.

apps/frontend/src/components/global/UserModal.vue
```diff
@@ -46,17 +48,28 @@
...
           </v-row>
           <v-row>
             <v-col>
-              <v-text-field id="title" v-model="userInfo.title" label="Title" />
+              <v-text-field
+                id="title"
+                v-model="userInfo.title"
+                name="title"
+                label="Title"
+              />
```
rbclark, line 66: If this text field already has an ID, what is the need for adding a name as well?

apps/frontend/src/components/global/UserModal.vue
```diff
@@ -7,27 +7,29 @@
...
+      <v-card-title id="userModalTitle" class="headline grey" primary-title>{{
         title
       }}</v-card-title>
       <v-card-text v-if="userInfo === null">
         <v-progress-linear indeterminate color="white" class="mb-0" />
       </v-card-text>
       <v-card-text v-else>
         <br />
-        <v-form>
+        <v-form name="userInfo">
```
rbclark, line 18: I see this uses name but other spots use `data-cy`, could you clarify on when and why you chose one versus the other?

apps/backend/src/guards/development.guard.ts
```diff
@@ -0,0 +1,8 @@
+import {CanActivate, Injectable} from '@nestjs/common';
+
+@Injectable()
+export class DevelopmentGuard implements CanActivate {
```
rbclark, line 4: I'm not sure how much the name makes sense now that it only runs on tests.

yarn.lock
```diff
@@ -1102,6 +1102,50 @@
     exec-sh "^0.3.2"
```
rbclark, line 1102: It looks like `yarn install` needs to be re-run, I'm not seeing the dependencies of jest-puppeteer removed.

test/integration/login.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  const loginPage = new LoginPage();
+  const loginPageVerifier = new LoginPageVerifier();
+  const registrationPage = new RegistrationPage();
+  const toastVerifier = new ToastVerifier();
+
+  // Run before each test
+  beforeEach(() => {
+    databaseHelper.clear();
+    registrationPage.register(CREATE_USER_DTO_TEST_OBJ);
+    cy.visit('127.0.0.1:3000/login');
```
rbclark, line 26: Is there a DRYer way to store the site address? It seems to be duplicated in a lot of locations, potentially making it overly difficult if the user needs to override the port or address for some reason or another.

rbclark (changes requested):

rbclark (approved):

PR #579, opened 2020-12-02
Display Topbar on Landing

### LINKED ISSUE #563: Add Topbar to Loading Modal
Currently it is difficult to access user information from the Upload Modal. A user has to load a sample before they can actually access their user information.

Displaying the Topbar on the Upload Modal page would provide users with the ability to access results without having to load a result.

This:
Screen Shot 2020-11-25 at 4 19 35 PM

Should display above:

Screen Shot 2020-11-25 at 4 20 09 PM

camdenmoors:
Closes #563, adds the top bar to the landing page.

### DIFF
--- a/apps/frontend/src/views/Landing.vue
+++ b/apps/frontend/src/views/Landing.vue
@@ -2,7 +2,7 @@
   <BaseView
     title="Load Sample"
     :show-toolbar="false"
-    :show-topbar="server_mode"
+    :show-topbar="serverMode"
     :minimal-topbar="true"
     :topbar-z-index="1000"
   >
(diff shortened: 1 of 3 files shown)


rbclark (changes requested):
This looks really good, one small suggestion would be to completely hide the Topbar in light mode and only show it in server mode.

apps/frontend/src/views/Landing.vue
```diff
@@ -1,22 +1,36 @@
...
+    UploadNexus,
+    BaseView
   }
 })
-export default class Landing extends Vue {}
+export default class Landing extends Vue {
+  get server_mode(): boolean {
+    return ServerModule.serverMode
+  }
+}
```
rbclark, line 35: Can you just use the ServerMixin instead of duplicating the function here?

rbclark (changes requested):

apps/frontend/src/views/Landing.vue
camdenmoors: Fixed

rbclark (approved):

Issue #334, opened 2020-10-13
Limit visibility of uploaded results to only the current user and provide user with ability to share with other users


rbclark:
Currently, heimdall2 makes results visible to all users whenever they are uploaded. This should instead be limited to only the user who saved the result and the user should then have the option to share with other users either by creating a group or setting it public.

PR #587, opened 2020-12-03
Add Users to Evaluations, Cleanup ABAC, Convert all PrimaryKeys to strings 

### LINKED ISSUE #21: Evaluations Controller
- [x] Evaluation DTO

- [x] Create Method

- [x] Read Methods (Index and by id)

- [x] Update Methods

- [x] Destroy Methods

- [x] New policy database seed files to allow users to perform actions on evaluations

- [x] Evaluation routes are behind authentication guards

- [x] Evaluation routes are behind authorization guards

- [x] Evaluation Module imported into Application module

rbclark:
This PR is necessary in order to make #334 and fixes a part of #21

As part of this PR, all of the Policies are removed from the database and are instead stored in a file as part of the application. This fixes type mismatches between the return values from the database and the application itself. The database for heimdall server uses BigInt for all of it's primary keys. A BigInt can be larger than a javascript number. Due to this, the type that comes from postgres for BigInt fields is string, not number. All of the types in the application stating that it is a javascript number are incorrect, and are fixed in this PR.

Still Todo:

- [x] Figure out how to deal with evaluations that already exist in users databases but do not have a user attached to them. (may need to add a 'public' field and mark all of these as public in order to avoid issues, or come up with another solution). Considering we already have SET NULL set for evaluations with users who are deleted, it may not be the worst thing to just let them be NULL.


apps/backend/test/evaluations.e2e-spec.ts
```diff
@@ -79,23 +80,25 @@ describe('/evaluations', () => {
...
 
       await login(app, LOGIN_AUTHENTICATION).then((response) => {
+        userId = response.body.userID;
         jwtToken = response.body.accessToken;
       });
     });
 
     describe('Create', () => {
-      it('should create an evaluation', async () => {
+      it.only('should create an evaluation', async () => {
```
Bialogs, line 96: Left an only in here.

apps/backend/test/users.e2e-spec.ts
```diff
@@ -594,6 +583,15 @@ describe('/users', () => {
           });
       });
 
+      it('should not be able to update themselves without providing the current password', async () => {
+        return update(
+          app,
+          editUserId,
+          UPDATE_USER_DTO_WITH_MISSING_CURRENT_PASSWORD_FIELD,
+          jwtToken
+        ).expect(HttpStatus.OK);
```
Bialogs, line 592: Why return OK when the test description/title is written like its expecting a failure?

Bialogs (changes requested):

apps/backend/src/users/user.model.ts
```diff
@@ -78,4 +78,8 @@ export class User extends Model<User> {
   @AllowNull(false)
   @Column(DataType.DATE)
   updatedAt!: Date;
+
+  get modelName(): string {
+    return 'User';
```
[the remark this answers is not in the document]
rbclark: I originally hoped for that as well, however at runtime everything in Javascript is just an object.

Screen Shot 2020-12-16 at 2 31 08 PM

https://casl.js.org/v4/en/guide/subject-type-detection#how-does-casl-detect-subject-type

Bialogs (approved):

PR #647, opened 2020-12-15
Evaluations Search, Edit Evaluations and Tags

### LINKED ISSUE #200: [Database Results] Tag editing, filtering, searching
Frontend components for tag editing, filtering and searching. Depends #181

The page for loading results from the database is currently very simple, it just displays a list of results. It should include the ability for users to add tags to results, and filter the results by name or tag.

camdenmoors:
Closes #200. Adds searching for evaluations and editing evaluation metadata (Filename and Tags). Blocked by #587, I will add types to backend once that is merged.


apps/frontend/src/store/snackbar.ts
```diff
@@ -23,22 +23,42 @@ export class Snackbar extends VuexModule {
...
   @Action
   failure(message: string) {
     this.context.commit('SET_ERROR', true);
     this.context.commit('SET_MESSAGE', message);
     this.context.commit('SET_VISIBILITY', true);
   }
+
+  @Action
+  HTTPFailure(error: any) {
+    console.log(error.response.data.message);
```
rbclark, line 43: Is the console.log necessary here?

apps/frontend/src/store/snackbar.ts
```diff
@@ -23,22 +23,42 @@ export class Snackbar extends VuexModule {
...
+    if (typeof error.response.data.message === 'object') {
+      this.notify(
+        error.response.data.message
+          .map(function capitalize(c: string) {
+            return c.charAt(0).toUpperCase() + c.slice(1);
+          })
+          .join(', ')
+      );
+    } else {
+      this.notify(error.response.data.message);
```
rbclark, line 53: Should this be using `failure` instead of `notify`? This seems like an error not just a notification.

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -1,16 +1,47 @@
 <template>
   <div>
-    <v-container>
+    <v-container class="ml-0 pl-0 mr-0 pr-0 pb-0 mb-0">
```
rbclark, line 3: Is the intent here to remove all margin and padding except the top? If so this could just be `my-0 py-0 pb-0 mb-0` or if this intended to be all directions `ma-0 pa-0`

apps/frontend/src/store/evaluations.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Action({rawError: true})
+  async updateEvaluation() {
+    return axios
+      .put(`/evaluations/${this.activeEvaluation.id}`, this.activeEvaluation)
+      .then((response) => {
+        this.context.commit('SET_ACTIVE_EVALUATION', response.data);
+      });
+  }
```
rbclark, line 70: Normally the Vuex store is used for storing items that need to have their state shared across multiple components. This module seems to really exist for the purpose of passing the file to the Edit page. It seems like there are a few alternative ways to implement this:

1. Could the full list of evaluations be synced to the edit component with an ID of the current one to edit? https://vuejs.org/v2/guide/components-custom-events.html#sync-Modifier
2. Instead of storing only the active evaluation in Vuex, why not just store an array of all evaluations in Vuex and then pass the ID of the one you want to edit over to the edit component?

I just don't think syncing only one evaluation to Vuex is the best way to approach this.

apps/frontend/src/store/evaluations.ts
```diff
@@ -0,0 +1,87 @@
...
+  VuexModule
+} from 'vuex-module-decorators';
+
+@Module({
+  namespaced: true,
+  dynamic: true,
+  store: Store,
+  name: 'EvaluationModule'
+})
+export class Evaluation extends VuexModule {
```
rbclark, line 18: I'm not sure on the naming of this, since profiles can be stored in the database in the same table as well. Although that may be a more systemic issue than just this one filename.

rbclark (changes requested):
I'll have to make another pass at this once I fix the deploy previews on Heroku. For now this is the code review comments.

apps/frontend/src/store/evaluations.ts
camdenmoors: Ended up syncing all evaluations to the store, and passing the ID.

apps/frontend/src/store/evaluations.ts
camdenmoors: That does seem more like a systemic issue, I think there should be a way to easily determine what's a profile, and what's an evaluation. Would an extra field in the data table work?

image

rbclark (changes requested):

apps/backend/src/evaluation-tags/evaluation-tags.controller.ts
```diff
@@ -43,7 +43,7 @@ export class EvaluationTagsController {
   async update(
     @Param('id') id: string,
     @Body() updateEvaluationTagDto: UpdateEvaluationTagDto
-  ): Promise<EvaluationTagDto> {
+  ): Promise<any> {
```
rbclark, line 46: Is it possible to avoid using <any> here?

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -35,26 +82,100 @@
...
+        result = true;
+      };
+    })
+    return result
+  }
+
+  get filteredFiles() {
+    let matches: any[] = []
+    if (this.search != '') {
+      (this.files as Array<any>).forEach(async (item: any) => {
```
rbclark, line 159: Can the use of <any> be avoided here?

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -35,26 +82,100 @@
...
+        if (this.filterEvaluationTags(item, this.search) || item.filename.toLowerCase().includes(this.search)) {
+          matches.push(item)
+        }
+      })
+    } else {
+      return this.files;
+    }
+    console.log(matches)
```
rbclark, line 167: Shouldn't leave in console.log during deployment.

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -35,26 +82,100 @@
...
+
+  editItem (item: IEvaluation) {
+    EvaluationModule.setActiveEvaluation(item);
+    this.editDialog = true;
+  }
+
+  deleteItem(item: any) {
+    EvaluationModule.setActiveEvaluation(item);
+    this.deleteDialog = true;
+  }
```
rbclark, line 135: It makes sense to me to store all of the Evaluations in the Evaluation Vuex store. It seems a little strange to save the active one there though. It makes a bit more sense to me for this to be stored in the LoadFileList component itself and then pass the active file to the EditEvaluationModal just like `this.editDialog` and `this.deleteDialog`.

test/support/verifiers/ResultsPageVerifier.ts
```diff
@@ -1,23 +1,23 @@
...
     ];
     const correctCards: string[] = [
-      'Passed: 0',
-      'Failed: 3',
-      'Not Applicable: 0',
-      'Not Reviewed: 0'
+      'Passed: 5',
+      'Failed: 60',
+      'Not Applicable: 1',
+      'Not Reviewed: 35'
```
rbclark, line 20: Could you elaborate on the reason for this change?

rbclark (changes requested):

test/support/verifiers/ResultsPageVerifier.ts
camdenmoors: The check for results data loads the first available sample, I changed the v-data-table to sort by filename automatically, so a different file is loaded during tests

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
camdenmoors: While possible, I don't think that this would work too well. It would mean directly mutating what's passed down to EditEvaluationModal (and the way around that would be using emitters, and then handling all edits in LoadFileList)

apps/frontend/src/components/global/upload_tabs/DatabaseReader.vue
```diff
@@ -87,6 +98,10 @@ export default class DatabaseReader extends mixins(ServerMixin) {
       });
   }
```
rbclark, line 99: This is probably more suited to live in the Evaluations store now since this.files was moved there.

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -35,26 +83,97 @@
...
+      (this.files as Array<IEvaluation | Sample>).forEach(async (item: any) => {
+        if (this.filterEvaluationTags(item, this.search) || item.filename.toLowerCase().includes(this.search)) {
+          matches.push(item)
+        }
+      })
+    } else {
+      return this.files;
+    }
+    return matches
+  }
```
rbclark, line 176:
```suggestion
  get filteredFiles() {
    let matches: Array<IEvaluation | Sample> = []
    if (this.search != '') {
      (this.files as Array<IEvaluation | Sample>).forEach(async (item: IEvaluation | Sample) => {
        if (this.filterEvaluationTags(item, this.search) || item.filename.toLowerCase().includes(this.search)) {
          matches.push(item)
        }
      })
    } else {
      return this.files;
    }
    return matches
  }
```

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -35,26 +83,97 @@
...
+
+  filterEvaluationTags(file: IEvaluation, search: string) {
+    let result = false;
+    file.evaluationTags?.forEach((tag) => {
+      if (tag.value.toLowerCase().includes(search.toLowerCase())) {
+        result = true;
+      };
+    })
+    return result
+  }
```
rbclark, line 153:
```suggestion
  filterEvaluationTags(file: IEvaluation | Sample, search: string) {
    let result = false;
    if('evaluationTags' in file)
    {
      file.evaluationTags?.forEach((tag) => {
        if (tag.value.toLowerCase().includes(search.toLowerCase())) {
          result = true;
        };
      })
    }
    return result
  }
```

apps/frontend/src/components/global/upload_tabs/EditEvaluationModal.vue
```diff
@@ -0,0 +1,187 @@
...
+        this.activeEvaluation.evaluationTags!.indexOf(tag),
+        1
+      )
+      SnackbarModule.notify("Deleted tag successfully.")
+    }).catch((error) => {
+      SnackbarModule.HTTPFailure(error)
+    });
+  }
+
+  async updateTag(tag: any) {
```
rbclark, line 165: For consistency here can the type of this be changed to `EvaluationTag` and all the places where it is used updated accordingly?

apps/frontend/src/components/global/upload_tabs/EditEvaluationModal.vue
```diff
@@ -0,0 +1,187 @@
...
+      text: 'Tag',
+      value: 'value'
+    },
+    {
+      text: 'Actions',
+      value: 'actions'
+    }
+  ];
+
+  async deleteTag(tag: any) {
```
rbclark, line 153: Ditto on the type here.

apps/frontend/src/components/global/upload_tabs/EditEvaluationModal.vue
```diff
@@ -0,0 +1,187 @@
...
+  @Prop({default: false}) visible!: boolean;
+  @Prop({type: Object, required: true}) readonly active!: IEvaluation;
+
+  newTagDialog: boolean = false;
+  deleteTagDialog: boolean = false;
+  awaitingFinishTyping: boolean = false;
+  activeTag: any = {
+    id: '-1',
+    value: ''
+  };
```
rbclark, line 139: Could this be converted to a full blown `EvaluationTag` type? (some fields will need to be added here but that's probably okay, I'd rather the extra fields over the `any` type)

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
```diff
@@ -35,26 +83,97 @@
...
 import Vue from 'vue';
 import Component from 'vue-class-component';
+import EditEvaluationModal from '@/components/global/upload_tabs/EditEvaluationModal.vue'
 
+import {SnackbarModule} from '@/store/snackbar';
+import {EvaluationModule} from '@/store/evaluations'
 import {IEvaluation} from '@heimdall/interfaces';
 import {Prop} from 'vue-property-decorator';
-import {Samples} from 'aws-sdk/clients/devicefarm';
+import {Sample, Samples} from 'aws-sdk/clients/devicefarm';
```
rbclark, line 92: I'm pretty sure this is the completely wrong sample altogether (pretty sure this one is on me, sorry). This should be:

```javascript
import {Sample} from '@/utilities/sample_util';
```

Once this is changed the filtered file search function gets simplified a bit.

libs/interfaces/evaluation/evaluation.interface.ts
```diff
@@ -2,9 +2,9 @@ import {IEvaluationTag} from '..';
 
 export interface IEvaluation {
   id: string;
-  readonly filename: string;
+  filename: string;
   readonly data: Record<string, any> | undefined;
-  readonly evaluationTags: IEvaluationTag[] | null;
+  evaluationTags: IEvaluationTag[] | null;
```
rbclark, line 7: This is most likely one of the many places where the null value should be removed.

rbclark (changes requested):

apps/frontend/src/components/global/upload_tabs/LoadFileList.vue
camdenmoors: That makes more sense, fixed now.

rbclark (approved):

PR #648, opened 2020-12-15
[WIP] First pass at creating a cloud.gov deployment

JJediny:
Blocked at Disk Space limit hitting the 2GB qouta - closing PR until it is ready to be re-opened. Will create an issue to track

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,4 @@
-release: yarn backend sequelize-cli db:migrate && yarn backend sequelize-cli db:seed:all 
+install: yarn install
+build: yarn run build
+db_migrate: yarn backend sequelize-cli db:migrate
+db_seed: yarn backend sequelize-cli db:seed:all
--- a/Procfile
+++ b/Procfile
@@ -2,3 +2,4 @@ install: yarn install
 build: yarn run build
 db_migrate: yarn backend sequelize-cli db:migrate
 db_seed: yarn backend sequelize-cli db:seed:all
+release: yarn backend sequelize-cli db:migrate && yarn backend sequelize-cli db:seed:all
--- a/package.json
+++ b/package.json
@@ -18,9 +18,9 @@
     "test:ui:open": "cypress open"
   },
   "workspaces": [
-      "apps/*",
-      "libs/*",
-      "test"
+    "apps/*",
+    "libs/*",
+    "test"
   ],
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^4.8.0",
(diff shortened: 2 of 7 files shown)


JJediny:
Recently attended a presentation by MITRE to [our team at GSA](https://handbook.tts.gsa.gov/). I'm part of the team that manages [our infrastructure](https://handbook.tts.gsa.gov/tech-portfolio/) and wanted to setup a quick deployment of the Heimdall2 server to http://cloud.gov, so we could try out your inspec tests and have somewhere to upload the results to track over time. 

I started working on a fork/branch but stuck, hoping someone would be willing to pair with me so I could commit back a new deployment/hosting option for the project?

Procfile
```diff
@@ -1 +1,4 @@
-release: yarn backend sequelize-cli db:migrate && yarn backend sequelize-cli db:seed:all
```
rbclark, line 1: This file is currently used by our Heimdall Server Heroku deployments, removing the release command will break our demo environments.

package.json
```diff
@@ -13,13 +17,6 @@
     "test:ui": "cypress run",
     "test:ui:open": "cypress open"
   },
-  "workspaces": {
-    "packages": [
-      "apps/*",
-      "libs/*",
-      "test"
-    ]
-  },
```
rbclark, line 22: This project uses yarn workspaces since it is a monorepo of 4 projects, 2 public (heimdall server and heimdall lite) and 2 private (Server E2E tests and Heimdall Interfaces). Removing the workspaces will cause issues with the build.

Procfile
JJediny: cloud foundry also uses a procfile https://docs.cloudfoundry.org/buildpacks/prod-server.html , could the `release` be added back to resolve or is that single command the only command allowed for the file? I can setup a cloud foundry specific bash script and cfignore the procfile as an alternative here

package.json
JJediny: this was removed to avoid a new bug we encountered. For any node application using the vscode workspace feature the cloud foundry buildpack will fail to parse the package.json file and will throw `**ERROR** Failed parsing package.json: json: cannot unmarshal object into Go struct field .workspaces of type []string`

Will revert its removal as a new commit

Procfile
rbclark: I believe Heroku will ignore commands that it does not recognize, so just adding the `release` command back should be fine.

package.json
rbclark: This shouldn't be a vscode workspace, this is utilizing yarn workspaces https://yarnpkg.com/features/workspaces. Any vscode workspace related information should be stored in the `heimdall2.code-workspace` file.

PR #653, opened 2020-12-16
Group dependencies into ones required for building and others

rbclark:
Only test dependencies are now included in the devDependencies. This allows for quicker and smaller installations of the application on build servers.


rbclark:
@JJediny this should fix a lot of the issues you were experiencing. A few notes from this PR:

* You can now build with the `--production` flag however it will require you to ignore or delete the `apps/frontend/tests`,  `apps/backend/test`, and `test` folders. For Docker I've implemented this using `.dockerignore`, your `.cfignore` should work for this purpose as well.
* I would also suggest ignoring `apps/frontend/public/heimdall-lite-2.0-demo-5fps.gif` since it is kind of large and unnecessary.

PR #663, opened 2020-12-18
Add Login with Github, Gitlab, Google, Okta

### LINKED ISSUE #480: Add Google, GitHub, GitLab support for login and signup

camdenmoors:
Closes #480.


apps/frontend/src/views/OauthCallbackHandler.vue
```diff
@@ -0,0 +1,33 @@
...
+import Vue from 'vue';
+
+@Component({})
+export default class OauthCallbackHandler extends Vue {
+  mounted() {
+    const params = new URLSearchParams(window.location.search);
+    if (params.get('code')){
+      ServerModule.LoginGithub(params.get('code')).then(() => {
+        setTimeout(function(){
+          window.location.href = "/"
```
rbclark, line 24: Can this be handled using `this.$router.push(route)`?

apps/frontend/src/views/Login.vue
```diff
@@ -123,5 +140,9 @@ export default class Login extends Vue {
         SnackbarModule.notify(error.response.data.message);
       });
   }
+
+  loginGithub() {
+    window.location.href = `http://${window.location.hostname}:${process.env.PORT || 3000}/authn/github`;
```
rbclark, line 145: Can this be handled using the vue router? `this.$router.push(route)`

Also does this actually need the full window location? `/authn/github` is being served by the backend which should always be on the same host.

apps/backend/src/authn/authn.service.ts
```diff
@@ -26,6 +28,33 @@ export class AuthnService {
...
+    try {
+      user = await this.usersService.findModelByEmail(email);
+    } catch {
+      const randomPass = crypto.randomBytes(128).toString();
+      const createUser: CreateUserDto = {
+        email: email,
+        password: randomPass,
+        passwordConfirmation: randomPass,
+        firstName: '',
+        lastName: '',
```
rbclark, line 42: Can this information be retrieved from github?

apps/backend/src/authn/authn.service.ts
```diff
@@ -26,6 +28,33 @@ export class AuthnService {
...
+        organization: '',
+        title: '',
+        role: ''
+      };
+      this.usersService.create(createUser);
+      throw new UnauthorizedException('Incorrect Username or Password');
```
rbclark, line 48: Won't this always just throw an exception?

apps/backend/src/authn/authn.service.ts
```diff
@@ -26,6 +28,33 @@ export class AuthnService {
...
+    }
+    if (user) {
+      this.usersService.updateLoginMetadata(user);
+      return user;
+    } else {
+      return null;
+    }
```
rbclark, line 55: For simplicity this could just be

```javascript
    if (user) {
      this.usersService.updateLoginMetadata(user);
    }

	return user;
```

rbclark (changes requested):

apps/frontend/src/views/OauthCallbackHandler.vue
camdenmoors: Because this is within it's own function it's unable to access this.$router

apps/frontend/src/views/Login.vue
camdenmoors: image
It doesn't seem to work on the frontend port, and router [doesn't support](https://stackoverflow.com/questions/44595929/vue2-navigate-to-external-url-with-location-href) visiting the page even if its on the same host but different port

rbclark: Is this only in dev mode that it doesn't work? The frontend should proxy unknown requests through to the backend, also the `http` scheme being hardcoded here is going to cause problems for people using https.

camdenmoors: Router doesn't work in production either, however, the http scheme is a problem. I've set it now to be relative to current location.

apps/frontend/src/store/server.ts
```diff
@@ -19,6 +19,8 @@ export interface IServerState {
   loading: boolean;
   token: string;
   banner: string;
+  enabledOAuth: string[];
+  userID: string;
```
rbclark, line 23: Instead of adding userID here, can the one in userInfo be used?

apps/frontend/src/store/server.ts
```diff
@@ -136,16 +141,41 @@ class Server extends VuexModule implements IServerState {
...
+    return axios.post('/authn/login', userInfo).then((response) => {
+      this.handleLogin(response);
       this.GetUserInfo();
     });
   }
 
+  @Action
+  public SetSessionValues(sessionInfo: {userID: string; accessToken: string}) {
+    this.context.commit('SET_USERID', sessionInfo.userID);
+    this.context.commit('SET_TOKEN', sessionInfo.accessToken);
```
rbclark, line 161: If handleLogin were parameterized, this code could leverage it as well.

apps/frontend/src/views/Login.vue
```diff
@@ -99,7 +112,30 @@ export default class Login extends Vue {
...
+    );
+    return match ? match[1] : null;
+  }
+
+  delete_cookies(names: string[]) {
+    names.forEach((name) => {
+      document.cookie =
+        name + '=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;';
+    });
+  }
```
rbclark, line 130: Would the vue cookies library work in place of this?

https://www.npmjs.com/package/vue-cookies

apps/frontend/src/views/Login.vue
```diff
@@ -99,7 +112,30 @@ export default class Login extends Vue {
...
+
   checkLoggedIn() {
+    const userID: string | null = this.getCookie('userID');
+    const accessToken: string | null = this.getCookie('accessToken');
+    if (userID && accessToken) {
+      ServerModule.SetSessionValues({userID: userID, accessToken: accessToken});
+      this.delete_cookies(['userID', 'accessToken']);
+    }
```
rbclark, line 138: This may be more appropriate to handle during the `CheckForServer` in `server.ts` in order to avoid the user being redirected from `/` to the `/login` view only to be redirected again back to `/`.

rbclark (changes requested):

apps/backend/src/main.ts
```diff
@@ -19,6 +22,11 @@ async function bootstrap() {
     })
   );
   app.use(json({limit: '50mb'}));
+  await app.use(
+    session({secret: generateDefault(), saveUninitialized: true, resave: true})
```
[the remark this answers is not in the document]
camdenmoors: This is [required](http://www.passportjs.org/docs/downloads/html/#middleware) as part of passport-okta-oauth, but now I've changed it to use the JWT Secret.

apps/backend/src/authn/github.strategy.ts
```diff
@@ -0,0 +1,57 @@
...
+    const userInfoResponse = await axios.get('https://api.github.com/user', {
+      headers: {Authorization: `token ${accessToken}`}
+    });
+    const firstName: string = userInfoResponse.data.name.substr(
+      0,
+      userInfoResponse.data.name.indexOf(' ')
+    );
+    const lastName: string = userInfoResponse.data.name.substr(
+      userInfoResponse.data.name.indexOf(' ') + 1
+    );
```
rbclark, line 40: This code seems basically the same as the chunk below, can you just make a `splitName` function?

apps/backend/src/main.ts
```diff
@@ -19,6 +22,15 @@ async function bootstrap() {
     })
   );
   app.use(json({limit: '50mb'}));
+  app.use(
+    session({
+      secret: configService.get('JWT_SECRET') || generateDefault(),
```
rbclark, line 27: Does the app even start without the JWT? According to the documentation it is a required value so there should be no need for a default.

rbclark (changes requested):

apps/backend/src/main.ts
camdenmoors: ConfigService will return String or Undefined, so while it won't start without it, it's still required to have a fallback:
image

apps/backend/src/authn/gitlab.strategy.ts
```diff
@@ -0,0 +1,39 @@
...
+    }
+  ): Promise<any> {
+    const email = profile.emails[0].value;
+    const firstName: string = profile._json.name.substr(
+      0,
+      profile._json.name.indexOf(' ')
+    );
+    const lastName: string = profile._json.name.substr(
+      profile._json.name.indexOf(' ') + 1
+    );
```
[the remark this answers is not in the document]
camdenmoors: My intention here was to have any middle names (or names with spaces) to appear in the firstName and just the last name in the lastName, but it seems I switched them up on accident.
image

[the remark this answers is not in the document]
camdenmoors: Switched to using
```js
const nameArray = profile._json.name.split(' ');
const email = profile.emails[0].value;
firstName: string = nameArray.slice(0, -1).join(' ');
lastName = nameArray[nameArray.length - 1];
```
Which appears to work as intended

apps/backend/.env-example
```diff
@@ -10,3 +10,18 @@ JWT_EXPIRE_TIME=<JSON Web Token Length of time before signature expires (if noth
...
+
+GITLAB_CLIENTID=<Github Application Client ID>
+GITLAB_CLIENTSECRET=<Github Application Client Secret>
+
+GOOGLE_CLIENTID=<Google Application Client ID (apps.googleusercontent.com)>
+GOOGLE_CLIENTSECRET="<Google Application Client Secret>"
+
+OKTA_DOMAIN="<yourdomain>.okta.com"
+OKTA_CLIENTID=<Your Client ID>
+OKTA_CLIENTSECRET=<Your Client Secret>
```
rbclark, line 27: Newline EOF

apps/frontend/src/views/Login.vue
```diff
@@ -39,16 +39,60 @@
                     @keyup.enter="login"
                     @blur="$v.password.$touch()"
                   />
-                  <v-btn
```
rbclark, line 42: Could this be updated to look as follows?

login

rbclark (changes requested):

rbclark (changes requested):
One small issue I encountered, user roles are not set by default for OAuth users. (The two in the middle were OAuth users).
Screen Shot 2021-01-14 at 6 36 06 PM

apps/backend/src/main.ts
```diff
@@ -36,6 +39,15 @@ async function bootstrap() {
...
+      saveUninitialized: true,
```
rbclark, line 45: According to the docs

> Forces a session that is "uninitialized" to be saved to the store. A session is uninitialized when it is new but not modified. **Choosing false is useful for implementing login sessions**, reducing server storage usage, or complying with laws that require permission before setting a cookie. Choosing false will also help with race conditions where a client makes multiple parallel requests without a session.

Is it possible to set this to false or does the app not work correctly then?

apps/backend/src/main.ts
```diff
@@ -36,6 +39,15 @@ async function bootstrap() {
...
+      resave: true
```
rbclark, line 46: Similar to above, based on the docs it may be useful to set this to false if possible since this causes the session to be saved over and over, which we may not need since we are just using this for login sessions.

apps/frontend/src/store/server.ts
```diff
@@ -136,16 +142,33 @@ class Server extends VuexModule implements IServerState {
         this.context.commit('SET_LOADING', false);
       });
   }
+  @Action
+  public async handleLogin(data: any) {
```
rbclark, line 146: Is it possible to avoid the use of any here?

apps/backend/src/authn/authn.controller.ts
```diff
@@ -13,4 +14,68 @@ export class AuthnController {
...
+  async loginToOkta(@Req() req: Request): Promise<any> {
+    return this.authnService.login(req.user as User);
+  }
+
+  @Get('okta/callback')
+  @UseGuards(AuthGuard('okta'))
+  async getUserFromOkta(@Req() req: Request): Promise<any> {
+    const session = await this.authnService.login(req.user as User);
+    await this.setSessionCookies(req, session);
+  }
```
rbclark, line 68: Please add the types for the return of all of these.

apps/backend/src/authn/authn.service.ts
```diff
@@ -26,6 +28,37 @@ export class AuthnService {
     }
   }
 
+  async oauthValidateUser(
+    email: string,
+    firstName: string,
+    lastName: string
+  ): Promise<any> {
```
rbclark, line 35: Please add return type here as well.

apps/backend/src/authn/github.strategy.ts
```diff
@@ -0,0 +1,53 @@
...
+      clientSecret: configService.get('GITHUB_CLIENTSECRET') || 'disabled',
+      scope: 'user:email',
+      passReqToCallback: true
+    });
+  }
+
+  async validate(
+    req: Record<string, unknown>,
+    accessToken: string
+  ): Promise<any> {
```
rbclark, line 25: Please add a return type here.

apps/backend/src/authn/gitlab.strategy.ts
```diff
@@ -0,0 +1,35 @@
...
+        `${configService.get('EXTERNAL_URL')}/authn/gitlab/callback` ||
+        'disabled'
+    });
+  }
+
+  async validate(
+    accessToken: string,
+    refreshToken: string,
+    profile: {
+      [key: string]: any;
```
rbclark, line 26: Instead of `any` here, can you just add the fields you expect to receive?

rbclark (changes requested):
Can you go through this and replace any usage of `any` with an appropriate type?

apps/backend/src/main.ts
camdenmoors: Both of these work when false

apps/backend/src/authn/gitlab.strategy.ts
camdenmoors: While I can't explicitly type all of the fields, I'll type the fields in use and leave the rest as any.

apps/backend/src/authn/authn.service.ts
```diff
@@ -26,6 +28,37 @@ export class AuthnService {
...
+      const randomPass = crypto.randomBytes(128).toString('hex');
+      const createUser: CreateUserDto = {
+        email: email,
+        password: randomPass,
+        passwordConfirmation: randomPass,
+        firstName: firstName,
+        lastName: lastName,
+        organization: '',
+        title: '',
+        role: 'user'
```
rbclark, line 49: Looks like this is supposed to have a default value on the model, not sure what was causing it to not be set (not saying to change this, just definitely weird behavior and may be worth investigating further since it popped up on the LDAP PR as well)

apps/backend/src/authn/github.strategy.ts
```diff
@@ -0,0 +1,75 @@
...
+import {PassportStrategy} from '@nestjs/passport';
+import axios from 'axios';
+import {Strategy} from 'passport-github';
+import {ConfigService} from '../config/config.service';
+import {User} from '../users/user.model';
+import {AuthnService} from './authn.service';
+
+interface GithubProfile {
+  name: string;
+  [key: string]: any;
```
rbclark, line 11: Are these `[key: string]: any;` types actually necessary? I tried removing one of them locally and it seems that everything still compiles successfully.

rbclark (changes requested):

rbclark:
This also needs the "Some Settings are managed by your organization" for users who try to edit their profile.

apps/backend/src/authn/github.strategy.ts
camdenmoors: They don't seem to be necessary, will remove them.

apps/backend/src/authn/authn.service.ts
camdenmoors: Yeah it definitely should be setting it by default, the only reason I can think of why it doesn't happen is if sequelize handles the AllowNull(false) decorator weirdly.

apps/backend/src/authn/github.strategy.ts
```diff
@@ -0,0 +1,84 @@
...
+    req: Record<string, unknown>,
+    accessToken: string
+  ): Promise<User> {
+    // Get user's linked emails from Github
+    const githubEmails = await axios
+      .get(
+        this.configService.get('CYPRESS_TESTING') === 'true'
+          ? 'http://127.0.0.1:3001/emails'
+          : 'https://api.github.com/user/emails',
+        {
```
rbclark, line 42: For more flexibility here, should the value come from the ENV? Basically if `CYPRESS_TESTING_CALLBACK` is set then use the URL stored in `CYPRESS_TESTING_CALLBACK`, else use `https://api.github.com/user/emails`

camdenmoors: To allow for enterprise GitHub instances, I'm adding GITHUB_ENTERPRISE.. INSTANCE_URL, TOKEN_URL, PROFILE_URL, and EMAIL_URL which will default to public github, and will be set to the mock server during testing.

.github/workflows/e2e-ui-tests.yml
```diff
@@ -21,52 +21,52 @@ jobs:
...
 
-     - name: Cypress run
-       uses: cypress-io/github-action@v2
-       with:
-         start: yarn start
-         wait-on: http://127.0.0.1:3000
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          start: yarn start, yarn run cypress-test mock-oauth
```
rbclark, line 64: Can you just say `yarn run start cypress-test mock-oauth` here?

camdenmoors: Tried this locally, doesn't seem to start the mock server

rbclark (approved):

PR #744, opened 2021-01-08
Add the ability to connect to Postgres with SSL

rbclark:
This may fix the seemingly random connection failures we are receiving on Heroku.


rbclark:
This option fixes the release step however it causes the actual application deployment to fail. Converted to Draft until I can fix that specific issue.

PR #764, opened 2021-01-14
Fix formatting of ControlRowCol's

### LINKED ISSUE #761: [BUG] Please restore word wrapping for `test` column for failed controls.
Please restore word wrapping for `test` column for failed controls.

image

Reproduce:
Load RedHat Eval results from samples and view the failed control `Test` panels.

rbclark:
The application was not properly wrapping text for "Test" field in ControlRowCol's. Fix the wrapping and respect newlines in Test and Result fields.

Fixes #761

### DIFF
--- a/apps/frontend/src/components/cards/controltable/ControlRowCol.vue
+++ b/apps/frontend/src/components/cards/controltable/ControlRowCol.vue
@@ -23,8 +23,8 @@
       <v-divider />
       <!-- HTML is sanitized with sanitize-html -->
       <!-- eslint-disable vue/no-v-html -->
-      <pre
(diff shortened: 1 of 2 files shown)


rbclark:
@aaronlippold @ejaronne @rx294 How married are we to the text justification? Without the text justification it is possible to do 
Screen Shot 2021-01-14 at 8 29 04 AM
instead of
Screen Shot 2021-01-14 at 8 29 18 AM
(The difference being it actually respects `\n`'s in the Test and Result box). Unfortunately the option that must be enabled to respect `\n`'s causes text justification to break.

PR #792, opened 2021-01-21
Add LDAP Authentication

### LINKED ISSUE #679: Add support for LDAP user authentication
**Add support for for LDAP and LDAPS authentication for heimdall2.**

In the initial iteration, the setup could be done over API calls. The API way has the benefit of avoiding plain-text files with LDAP password.

It also allows for setup through env vars for docker/k8s deployments.
For traditional deployments we could develop a script + yaml conf file to automate the setup.

We could make localhost based actions non-authenticated allowing for automated setups.

Later the API could be hooked View panels to allow UI based setup.

Example on how gitlab deployment allows for LDAP setup through env vars

https://github.com/sameersbn/docker-gitlab/blob/e95c3b744af2ed0302424c40c049717630522b67/README.md#available-configuration-parameters

camdenmoors:
Closes #679, adds the ability to login using LDAP. Configuration is set in .env as follows:

`LDAP_ENABLED`: Choose whether or not to use ldap (True/False)
`LDAP_HOST`: Your target LDAP server (Default 389)
`LDAP_BINDDN`: User account used to query account lookups
`LDAP_PASSWORD`: Password for user account used to query account lookups
`LDAP_SEARCHBASE`: The location to search within LDAP (AD Users would use `"OU=Users, DC=<yourdomain>, DC=local"`)
`LDAP_SEARCHFILTER`: LDAP search filter (AD Users would use sAMAccountName={{username}})
`LDAP_NAMEFIELD`: The field that contains the user's name (default is 'name', works for AD)
`LDAP_MAILFIELD`: The field that contains the user's email (default is 'mail', works for AD)


camdenmoors:
Converting to a draft until tests are added

apps/frontend/src/views/Login.vue
```diff
@@ -4,61 +4,126 @@
...
+                background-color="primary darken-1"
+                color="primary-visible"
+                show-arrows
+              >
+                <v-tab id="select-tab-standard-login" href="#login-standard"
+                  >Heimdall Login</v-tab
+                >
+                <v-tab id="select-tab-ldap-login" href="#login-ldap"
+                  >Organization Login</v-tab
+                >
```
rbclark, line 26: The actual code for these 2 pieces should be in their own component. There should be a LdapLogin component and a DatabaseLogin component.

apps/frontend/src/views/Login.vue
```diff
@@ -71,31 +136,47 @@ import Vue from 'vue';
...
 export default class Login extends Vue {
   email: string = '';
+  username: string = '';
   password: string = '';
+  active_tab: string = login_tab.get_default('logintab-standard')
 
   mounted() {
+    if(!ServerModule.ldap){
+      document.querySelectorAll('[role="tablist"]')[0].remove()
+    }
```
rbclark, line 179: Modifying the DOM directly like this seems like bad practice. Can you just add a `v-if` to the `v-tab` above to hide the LDAP tab if LDAP is not available?

rbclark (changes requested):

apps/frontend/src/views/Login.vue
camdenmoors: Unfortunately the tablist component isn't accessible, and only hiding v-tab still shows the container it sits on

rbclark: Adding `:height="ldapenabled ? '' : 0"` to the `v-show` seems to have the equivalent behavior without modifying the DOM.

camdenmoors: Changed to just always show "Heimdall Login" and "Organization Login" if supported.

camdenmoors (commented):
Did not mean to start a review. Hopefully commenting will let me remove this

camdenmoors:
> Is there an advantage to using this container over `rroemhild/test-openldap` which comes preseeded with LDAP information?

Nope, will switch to that.

apps/backend/src/authn/authn.service.ts
```diff
@@ -26,6 +30,37 @@ export class AuthnService {
     }
   }
 
+  async validateOrCreateUser(
+    email: string,
+    firstName: string,
+    lastName: string
+  ): Promise<any> {
```
rbclark, line 37: This will always return a User object correct? Can the type be updated accordingly?

apps/backend/src/authn/ldap.strategy.ts
```diff
@@ -0,0 +1,46 @@
...
+          user[configService.get('LDAP_NAMEFIELD') || 'name']
+        );
+        const email: string =
+          user[configService.get('LDAP_MAILFIELD') || 'mail'];
+
+        req.user = this.authnService.validateOrCreateUser(
+          email,
+          firstName,
+          lastName
+        );
```
rbclark, line 41: Role is missing from LDAP users on creation.

apps/frontend/src/views/Login.vue
```diff
@@ -69,31 +45,22 @@
...
-    }
+  components: {
+    LocalLogin,
+    LDAPLogin
   }
 })
 export default class Login extends Vue {
-  email: string = '';
-  password: string = '';
+  username: string = '';
```
rbclark, line 62: Is username still necessary here?

rbclark (changes requested):
There is also currently a bug where if an LDAP user tries to edit their user info, all of the fields are editable however the current password is randomly generated so the user will not know it. Perhaps it would be good to add a field to the database and save the name of the authentication provider (ldap, local, google, gitlab, etc). Then if a user browses to the edit page we can just greet them with a banner that says "Some settings are managed by your organization" and not allow them to edit their profile. Note that administrators should still be able to edit users (so that they can change their role, etc).

rbclark (approved):

PR #859, opened 2021-02-02
Prevent User from Accidentally Re-Uploading Evaluations

### LINKED ISSUE #218: Warn user when uploading the same file multiple times
A user can upload the same result multiple times with no warning that they have duplicated the results.

I would like the user to be warned if they try to upload the same `result.json` twice.

This would eliminate confusion for users who are uploading results and do not potentially realize that they have loaded the same file multiple times.

Related to https://github.com/inspec/inspec/issues/5262

camdenmoors:
Closes #218. Prevents users from uploading an existing file into the database.

### DIFF
--- a/apps/frontend/src/components/global/sidebaritems/SidebarFileList.vue
+++ b/apps/frontend/src/components/global/sidebaritems/SidebarFileList.vue
@@ -42,7 +42,6 @@ import {ServerModule} from '@/store/server';
 @Component
 export default class FileItem extends mixins(ServerMixin) {
   @Prop({type: Object}) readonly file!: EvaluationFile | ProfileFile;
-  clickToConfirm: boolean = false;
 
   select_file() {
     if (this.file.hasOwnProperty('evaluation')) {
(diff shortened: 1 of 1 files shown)


rbclark (changes requested):
Can we just check if the file has a database_id and if it does disable the save button completely?

apps/frontend/src/components/global/sidebaritems/SidebarFileList.vue
```diff
@@ -74,7 +74,10 @@ export default class FileItem extends mixins(ServerMixin) {
     let file = InspecDataModule.allFiles.find(
       (f) => f.unique_id === this.file.unique_id
     );
-    if (file) {
+    if (file?.database_id){
+      SnackbarModule.failure('This evaluation is already in the databse.')
```
rbclark, line 78: Can you have the icon update to be disabled and just have the mouseover text say "Saved" instead? I'd prefer to have a saved with checkbox icon but it seems that icon doesn't actually exist sadly.

rbclark (approved):

PR #897, opened 2021-02-05
Require Confirmation Before Deleting Tags

### LINKED ISSUE #890: Add "Are you Sure" prompt to deleting tags
It is currently very easy to delete tags.

easytodelete

Would be beneficial to show an "Are you sure" prompt first.

camdenmoors:
Closes #890, requires confirmation before deleting tags.


rbclark (changes requested):
Cancel button currently does not work.

rbclark (approved):

PR #898, opened 2021-02-05
Catch Alternate Authentication Errors

### LINKED ISSUE #892: Display non-api messages on authn callback routes
Currently all of the OAuth errors are just displayed as JSON responses. It would be great to decorate these in some manner (preferably without breaking the other API endpoints in the app). This could be as simple as setting up a proper 500 and 401 error pages in nest.

camdenmoors:
Closes #892, displays the attached error message when alternate authentication via callback fails.
image


rbclark (changes requested):
It would be better if this redirected back to the homepage and set an error message in a cookie that the frontend recognizes to more cleanly display the error to the user.

rbclark (approved):

PR #899, opened 2021-02-05
Fix alignment of OAuth Buttons

### LINKED ISSUE #891: Fix alignment of Oauth button when there is only 1 provider
Screen Shot 2021-02-04 at 4 23 13 PM

It should be either centered or left-aligned, not in-between.

camdenmoors:
Closes #891. Uses standard d-flex grid and center-aligns.


rbclark (changes requested):
I believe my merging of openID connect caused a merge conflict on this one.

rbclark (approved):

Issue #765, opened 2021-01-14
Intermittent Test Failure in E2E Backend Tests (loginCount not incrementing)


rbclark:
There has been an intermittent failure in the tests that is causing the test suite to fail. The error is: 

```
 FAIL test/users.e2e-spec.ts (84.083 s)
  ● /users › Functions that require authentication › Read › should return 200 status when user is returned

    expect(received).toEqual(expected) // deep equality

    Expected: 1
    Received: 0

      251 |             );
      252 |             expect(response.body.lastLogin).toBeDefined();
    > 253 |             expect(response.body.loginCount).toEqual(1);
          |                                              ^
      254 |             expect(response.body.organization).toEqual(
      255 |               CREATE_USER_DTO_TEST_OBJ.organization
      256 |             );

      at users.e2e-spec.ts:253:46
```

For some reason the `loginCount` is not always incrementing, or is incrementing but is not being properly reflected in the return data.

https://github.com/mitre/heimdall2/runs/1702301802

PR #932, opened 2021-02-12
Convert E2E Tests to Controller Tests

### LINKED ISSUE #849: Rewrite most/all e2e tests as controller tests
Currently the Heimdall Backend has both `controller` and `e2e` tests in it. In my opinion, the `e2e` tests could be much better served to be rewritten as `controller` tests for two reasons:

1. The mocking of controllers seems mostly useless since our controllers require objects that require significant setup, it is much easier to just setup the database as expected and see if the controller responds appropriately.
2. The `e2e` tests are significantly harder to debug than the controller tests.
3. (Bonus reason) This should make debugging and fixing errors such as #765 much easier.

camdenmoors:
Closes #849 and #765. Converts end to end tests to controller tests (except Authn tests, as the controller takes `@Req() req: Request` as an argument). Keeping as draft until I finish comments. evaluationsController e2e tests should be converted as part of the groups update (#794).


apps/backend/src/users/users.controller.spec.ts
```diff
@@ -29,191 +48,525 @@ import {UsersService} from './users.service';
...
+          await usersController.remove(
+            adminUser.id,
+            {user: adminUser},
+            DELETE_USER_DTO_TEST_OBJ
+          );
+        }).rejects.toThrow(ForbiddenException);
+      });
+
+      // Admins should be able to remove their account if there is another administrator
+      it('should test remove function with admin user and there is another admin', async () => {
```
Bialogs, line 552: keep this test but rewrite and move to users.service.spec.ts

apps/backend/src/users/users.controller.spec.ts
```diff
@@ -29,191 +48,525 @@ import {UsersService} from './users.service';
...
+
+      // Admins should be able to remove other users without their password
+      it('should test remove function with admin user and a dto that has no password', async () => {
+        expect(
+          await usersController.remove(basicUser.id, {user: adminUser}, {})
+        ).toEqual(new UserDto(basicUser));
+      });
+
+      // Admins should not be able to remove their account if they are the only administrator
+      it('should test remove function with admin user that is the only admin', async () => {
```
Bialogs, line 539: keep this test but rewrite and move to users.service.spec.ts

apps/backend/src/users/users.controller.spec.ts
```diff
@@ -29,191 +48,525 @@ import {UsersService} from './users.service';
...
+          await usersController.remove(
+            basicUser.id,
+            {user: basicUser},
+            DELETE_USER_DTO_TEST_OBJ_WITH_MISSING_PASSWORD
+          );
+        }).rejects.toThrow(ForbiddenException);
+      });
+
+      // Admins should be able to remove other users without their password
+      it('should test remove function with admin user and a dto that has no password', async () => {
```
Bialogs, line 532: keep this test but rewrite and move to users.service.spec.ts

Bialogs (changes requested):

Bialogs:
You can keep the additional expec.assertions(#) as well but I think most other tests in users.controller.spec.ts are duplicates.

Bialogs:
The github action that runs the e2e tests also needs to be removed since we won't have any of those anymore.

Bialogs (approved):

PR #979, opened 2021-02-23
Add an Expand All Button

### LINKED ISSUE #942: Add expand all button on results view

camdenmoors:
Closes #942, adds an expand all button only if single expand is disabled.

### DIFF
--- a/apps/frontend/src/components/cards/controltable/ControlTable.vue
+++ b/apps/frontend/src/components/cards/controltable/ControlTable.vue
@@ -3,7 +3,12 @@
     <!-- Toolbar -->
     <v-row>
       <v-col cols="12">
-        <v-switch v-model="single_expand" label="Single Expand" class="mt-2" />
+        <v-switch
+          v-model="single_expand"
+          label="Single Expand"
+          class="mt-2"
+          @change="$emit('toggleExpandAll', single_expand)"
+        />
       </v-col>
     </v-row>
 
@@ -101,6 +106,7 @@ interface ListElt {
 export default class ControlTable extends Vue {
   @Prop({type: Object, required: true}) readonly filter!: Filter;
   @Prop({type: Boolean, required: true}) readonly showImpact!: boolean;
+  @Prop({type: Boolean, required: false, default: false}) readonly expandAll!: boolean;
   // Whether to allow multiple expansions
   single_expand: boolean = true;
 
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/cards/controltable/ControlTable.vue
```diff
@@ -133,6 +139,16 @@ export default class ControlTable extends Vue {
...
 
+ /** Expands all elements within items asynchronously */
+  async expandAllEvaluations(): Promise<void> {
+    this.items.forEach(async (item) => {
+      const i = this.expanded.indexOf(item.key);
+      if (i < 0) {
+        this.expanded.push(item.key);
+      }
+    })
+  }
```
rbclark, line 150: Instead of doing a check for every single element if it is already in the array, can we save ourselves the check and just replace the whole `this.expanded` list?

```
this.expanded = this.items.map((items) => items.key);
```

apps/frontend/src/views/Results.vue
```diff
@@ -162,10 +162,25 @@
...
+                <v-col
+                  v-if="show_expand_all_button"
+                  cols="auto"
+                  class="text-right"
+                >
+                  <v-btn color="primary" text @click="expand_all"
+                    >Expand All</v-btn
+                  ></v-col
+                >
+              </v-row>
```
rbclark, line 178: It would be better if this lived with the Single Expand toggle (maybe as a toggle as well?)

apps/frontend/src/views/Results.vue
```diff
@@ -456,6 +473,16 @@ export default class Results extends Vue {
...
+
+  // Expand all visible controls
+  expand_all() {
+    (this.$refs.controlTable as Vue & { expandAllEvaluations: () => void }).expandAllEvaluations();
+  }
+
+  // Toggles whether or not to show expand all button
+  set_show_expand_all(single_expand: boolean) {
+    this.show_expand_all_button = !single_expand;
+  }
```
rbclark, line 485: No need to hide the button, just make the button change the state of Single Expand.

apps/frontend/src/views/Results.vue
```diff
@@ -456,6 +473,16 @@ export default class Results extends Vue {
       this.eval_info = index;
     }
   }
+
+  // Expand all visible controls
+  expand_all() {
+    (this.$refs.controlTable as Vue & { expandAllEvaluations: () => void }).expandAllEvaluations();
+  }
```
rbclark, line 480: Moving the 'Expand All' toggle to ControlRow should should remove the need for this.

rbclark (changes requested):

rbclark:
@ejaronne Can you take a look at the netlify deploy preview on this one and tell me if you still want this one merged even though Vue is lazily rendering the page which is causing the print-ability to be inconsistent?

ejaronne:
@rbclark I see the lag too. Even before this, if one loads many results files, the unexpanded table lags as it optimizes memory and shows only what is in your immediate window. When going to print preview, you see countless blank pages showing what is really going on above/below the current view. Now, even with one result set I'm seeing the lag you see, and the print preview is bad as well.  Note however this expand all is also to not have to open each of 200+ rows to find out what's going on. Not sure why Heimdall-lite 1.0 did this so much faster. Let's talk, please.

rbclark (approved):

PR #990, opened 2021-02-25
Enable ESLint no-explicit-any

### LINKED ISSUE #793: Enable '@typescript-eslint/no-explicit-any' for the frontend and backend
It is bad practice to use `any` inside of typescript unless it is critically necessary. Currently the `'@typescript-eslint/no-explicit-any'` rule is disabled, allowing some instances of this to slip by. In order to enable this rule, someone will need to first analyze the 15 instances of `any` used in the backend and then `70` instances in the frontend. Once those are fixed it will be trivial to enable this rule.

In some instances it should be very simple to fix these, just mousing over in VSCode for the backend login function for example shows the return type of the related service, so this is just a matter of copy pasting the related service return type onto this function.

Screen Shot 2021-01-21 at 12 04 38 PM

camdenmoors:
Closes #793, enables `@typescript-eslint/no-explicit-any` in the frontend and backend, and removes as many instances of any as possible.

Remaining instances:
 - OAuth Strategies: Callback is a function and cannot be typed within the function declaration.
 - UtilColorGenerator: Color objects are typed as readonly, so we must declare our own object (`l`) and then return it `as Color`
 - On_resize: Resize events (handled by [vue-resize-directive](https://github.com/David-Desmaisons/Vue.resize)) don't have their own type other than Event, which won't have clientWidth
 - HTTPFailure: It seems catches from axios requests don't have their own type, so HTTPFailure will continue to take any and hope that `error.response.data.message` exists


apps/frontend/src/components/global/upload_tabs/aws/S3Reader.vue
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
rbclark, line 1: Does this disable for the whole file?

rbclark:
Can we take advantage of the `unknown` typescript keyword for any of these? For example, for HTTPFailure it could look like this:

```typescript
  @Action
  HTTPFailure(error: unknown) {
    const nestedError = _.get(error, 'response.data.message');
    if (typeof nestedError === 'object') {
      this.failure(
        nestedError
          .map(function capitalize(c: string) {
            return c.charAt(0).toUpperCase() + c.slice(1);
          })
          .join(', ')
      );
    } else {
      this.failure(nestedError || error);
    }
  }
```

Which no longer requires `any`.

rbclark (approved):

PR #974, opened 2021-02-22
Add Sharing to Evaluations

camdenmoors:
Adds a route `/results/:id` to the frontend which loads the passed comma separated evaluation(s).

PR #1003, opened 2021-03-01
Add Sharing to Evaluations

### LINKED ISSUE #853: Add ability for direct linking to results
It currently is not possible to provide a link directly to an evaluation (something such as https://mitre-heimdall.herokuapp.com/results/1) which would take a user to a result with the database ID 1. This would be very beneficial for sharing results between teams.

camdenmoors:
Adds a route `/results/:id` to the frontend which loads the passed comma separated evaluation(s). Also returns a share URL on evaluation creation. Closes #853, and supersedes #974.


apps/backend/src/evaluations/evaluations.controller.ts
```diff
@@ -85,13 +87,18 @@ export class EvaluationsController {
...
+    );
     return new EvaluationDto(
-      // Do not include userId on the DTO so we can set it automatically to the uploader's id.
-      await this.evaluationsService.create(
-        updatedEvaluationDto,
-        serializedDta,
-        request.user.id
-      )
+      createdDto,
+      false,
```
rbclark, line 98: If the user just created an evaluation then shouldn't editable be true since they are the creator?

rbclark (changes requested):

apps/frontend/src/store/evaluations.ts
```diff
@@ -35,6 +37,38 @@ export class Evaluation extends VuexModule {
...
 
+  @Action
+  findEvaluationsByIds(evaluationIds: string[]): IEvaluation[] {
+    return this.allEvaluations.filter((evaluation) =>
+      evaluationIds.includes(evaluation.id)
+    );
+  }
+
+  @Action
+  async load_results(evaluations: IEvaluation[]): Promise<(FileID | void)[]> {
```
rbclark, line 48: If the type here is loosened down from a whole `IEvaluation[]` down to `{id: string}[]` then it helps in some other areas.

apps/frontend/src/views/Results.vue
```diff
@@ -282,6 +285,26 @@ export default class Results extends Vue {
...
+        return
+      }
+      await EvaluationModule.getAllEvaluations();
+      const evaluationsToLoadArray: IEvaluation[] = await EvaluationModule.findEvaluationsByIds(this.$route.params.id.split(','));
+      if (evaluationsToLoadArray.length !== 0) {
+        EvaluationModule.load_results(evaluationsToLoadArray)
+      } else {
+        SnackbarModule.failure(`Heimdall was passed the following evaluations to open, but couldn't find any of them: ${this.$route.params.id}`);
+        this.$router.push('/results');
+      }
```
rbclark, line 304: Instead of doing the check against all of the evaluations, if the change below is made you can just go ahead and make a call to `EvaluationModule.load_results(this.$route.params.id.split(','))`.

apps/frontend/src/views/Results.vue
```diff
@@ -282,6 +285,26 @@ export default class Results extends Vue {
...
+      if(this.$route.params.id === 'results') {
+        return
+      }
+      await EvaluationModule.getAllEvaluations();
+      const evaluationsToLoadArray: IEvaluation[] = await EvaluationModule.findEvaluationsByIds(this.$route.params.id.split(','));
+      if (evaluationsToLoadArray.length !== 0) {
+        EvaluationModule.load_results(evaluationsToLoadArray)
+      } else {
+        SnackbarModule.failure(`Heimdall was passed the following evaluations to open, but couldn't find any of them: ${this.$route.params.id}`);
+        this.$router.push('/results');
```
rbclark, line 303: Is it necessary to push to results here? Aren't we already on results?

apps/frontend/src/components/global/upload_tabs/DatabaseReader.vue
```diff
@@ -93,30 +88,18 @@ export default class DatabaseReader extends mixins(ServerMixin) {
...
+      if(evaluation.hasOwnProperty('id')){
+        stringresult += `${(evaluation as IEvaluation).id},`;
+      }
     });
+    return stringresult.slice(0, -1);
+  }
+
+  load_results(evaluations: IEvaluation[]): void {
+    this.$router.push(`/results/${this.joinEvaluationsByIds(evaluations)}`);
   }
```
rbclark, line 103: If the files aren't loaded all at once then this is not working correctly. (watch URL bar)

sharingURL

rbclark (changes requested):

apps/frontend/src/views/Results.vue
camdenmoors: load_results takes a list of IEvaluations, not just IDs, so unfortunately we have to find each one that's passed before loading

rbclark: That was my reason for suggesting the change of the type of `load_results` from a list of evaluations to just an ID. It will require a few other small changes to `load_results` but all of the data should be available as part of the response from ```axios.get<IEvaluation>(`/evaluations/${evaluation.id}`)```

apps/frontend/src/views/Results.vue
```diff
@@ -282,6 +284,24 @@ export default class Results extends Vue {
   /** Determines if we should make the search bar colapseable */
   show_search_mobile: boolean = false;
 
+  /**
+   * On mount, check if we have been passed an evaluation
+   * to load from Vue router
+   */
+  async mounted() {
+    if(this.$route.params.id){
+      if(this.$route.params.id === 'results') {
```
rbclark, line 293: Does this cause problems if the route is `profiles` with no parameters?

apps/backend/src/evaluations/evaluations.controller.ts
```diff
@@ -78,23 +78,26 @@ export class EvaluationsController {
   async create(
     @Body() createEvaluationDto: CreateEvaluationDto,
     @UploadedFile() data: Express.Multer.File,
-    @Request() request: {user: User}
+    @Request() request: {user: User; headers: {origin: string}}
```
camdenmoors, line 81: This requires the Origin header to be set by anything interacting with the API to set an Origin header if the caller wants a non-relative shareURL (and if another web-app interacts with Heimdall's API this might set the share URL to the wrong host)

ejaronne (approved):
pretty cool: when I save a result set, copy the url, delete it from the db, then try the link I get a 404 (not found). When I go to numbers "in between" the ones I see in my list, I get a 403 (forbidden), which means something is there but I do not have access.

rbclark:
One small final bug on this before merging that I'm working to fix is the backend URL is returning `undefined/results/...` when uploading via the API instead of `<URL>/results/...`. Once that is fixed this will be ready to go.

rbclark (approved):

PR #1046, opened 2021-03-08
Ask for Name on Signup

### LINKED ISSUE #1021: Ask user for name on registration
The lack of this piece of user information is not ideal on the user and groups pages. Asking all new users would at least fix this problem going forward:

Screen Shot 2021-03-03 at 10 35 05 AM

camdenmoors:
Closes #1021, adds a requiredname field to Signup.vue, and lets the Topbar Dropdown work with just a first name.


rbclark (changes requested):
This should include firstName and lastName, not just firstName.

rbclark (changes requested):
Instead of trying to split one field, can you just ask for firstName and lastName?

apps/frontend/src/components/global/TopbarDropdown.vue
```diff
@@ -118,6 +118,8 @@ export default class TopbarDropdown extends mixins(ServerMixin) {
         this.userInfo.firstName.charAt(0) +
         this.userInfo.lastName.charAt(0)
       );
+    } else if (this.userInfo.firstName){
+      return this.userInfo.firstName.charAt(0);
```
rbclark, line 122: I'd be fine with just getting the first 2 characters if we know we only have a firstname.

apps/frontend/src/views/Signup.vue
```diff
@@ -163,6 +208,14 @@ export default class Signup extends Vue {
...
   }
 
+  splitName(fullName: string): {firstName: string; lastName: string} {
+    const nameArray = fullName.split(' ');
+    return {
+      firstName: nameArray.slice(0, -1).join(' '),
+      lastName: nameArray[nameArray.length - 1]
+    };
+  }
+
```
rbclark, line 218: This doesn't seem to be called anywhere.

rbclark (approved):

PR #1065, opened 2021-03-09
Always check the root HDF when hiding objects for compare view

rbclark:
This fixes an issue where incorrect data was being shown for overlayed profiles.


ejaronne (commented):
Works.

Bialogs (approved):

PR #1068, opened 2021-03-10
Cleanup Status Counts

### LINKED ISSUE #80: Inconsistent test counts
using this sample json: [t100_08102020.zip](https://github.com/mitre/heimdall-lite/files/5054011/t100_08102020.zip)

from this profile: https://github.com/CMSgov/inspec-profile-disa_stig-el7

Shouldn't the numerators add up to the total tests below?
image

...AND is a goal to be consistent with the CLI counts? Not sure if that is the intent, so please correct me if I'm wrong:
image

rbclark:
This clarifies the status counts along the top of Heimdall, removes the total count, and makes the count of passing and failing individual controls more clear in the failed box. It also stops the counting of "FailedOutOf", "NotApplicableTests", "NotReviewedTests", "ErroredOutOf", "ErroredTests", and "TotalTests" since those counts are no longer used anywhere and counting them would be a waste of CPU cycles.

Closes #80

### DIFF
--- a/apps/frontend/src/components/cards/StatusCardRow.vue
+++ b/apps/frontend/src/components/cards/StatusCardRow.vue
@@ -69,46 +69,43 @@ export default class StatusCardRow extends Vue {
       {
         icon: 'check-circle',
         title: 'Passed',
-        subtitle: `Has ${StatusCountModule.countOf(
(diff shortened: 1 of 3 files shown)


ejaronne (approved):
Looks good. Thanks!
image

PR #1072, opened 2021-03-10
WIP: Automatically expand all controls whenever a user opens the system print dialog

rbclark:
Currently only works in chrome. This could potentially be fixed by adding a `Print` button for other browsers and requiring users to click on it.


ejaronne (commented):
I'm not sure it's worthwhile if we provide later an export to different report formats. The time required to create the preview is several minutes for 200+ controls. Not sure folks will bother with trying it after a few times of experiencing that lag.

PR #1085, opened 2021-03-12
Show Loading Bar when Loading Samples

### LINKED ISSUE #117: Switch to async pattern, add in process indicator for multiple file loading, with success and fail

camdenmoors:
Possibly closes #117, adds a loading bar while loading samples.


rbclark (changes requested):
If loading the load button should also probably be disabled.

rbclark (approved):

PR #1121, opened 2021-03-19
Sort Results File List Ascending by start_time

### LINKED ISSUE #1053: Sort Results View the same as Comparison View
Currently one sorts by date and the other does not. Results view should sort loaded files the same way comparison view does now.

camdenmoors:
Closes #1053, sorts results control table ascending by start_time (same as compare view).


camdenmoors:
Superseded by #1124

Issue #887, opened 2021-02-04
Display filename tags on the right side of the Profile Info


rbclark:
Currently tags are only displayed on the Database load tab. Having them displayed when the file itself is loaded as well would be useful.

![Screen Shot 2021-02-04 at 2 45 51 PM](https://user-images.githubusercontent.com/2308869/106947075-ed5a3780-66f7-11eb-901c-076a9ffb8808.png)

rbclark:
Blocked until #794 is merged.

PR #1124, opened 2021-03-19
Show Evaluation Tags on EvaluationInfo, sort EvaluationInfo Components by start_time

### LINKED ISSUE #1053: Sort Results View the same as Comparison View

camdenmoors:
Closes #1053 and #887, supersedes #1121, shows the Evaluation tags of evaluations in the database.


apps/frontend/src/views/Results.vue
```diff
@@ -362,6 +354,10 @@ export default class Results extends mixins(RouteMixin) {
     this.search_term = '';
   }
 
+  mounted() {
+    EvaluationModule.getAllEvaluations()
+  }
```
rbclark, line 359: I believe this is causing a 404 error in Heimdall lite mode.

apps/frontend/src/views/Results.vue
```diff
@@ -472,4 +439,7 @@ export default class Results extends mixins(RouteMixin) {
   top: 4px;
   z-index: 5;
 }
+span + span {
+  margin-left: 10px;
+}
```
rbclark, line 444: I'm not quite sure what this is adding.

apps/frontend/src/views/Results.vue
```diff
@@ -362,6 +355,12 @@ export default class Results extends mixins(RouteMixin) {
     this.search_term = '';
   }
 
+  mounted() {
+    if(this.serverMode) {
+      EvaluationModule.getAllEvaluations()
+    }
+  }
```
rbclark, line 362: This shouldn't be necessary if the evaluation already has the tags included when it is loaded.

rbclark (changes requested):

PR #1155, opened 2021-03-26
Add Edit Profile Info to Results View

camdenmoors:
Closes #886, adds an edit tag to the parent component of EvaluationInfo.vue. Depends on #1124.

PR #1169, opened 2021-03-30
Add Edit Profile Info to Results View

### LINKED ISSUE #886: Link to Result Edit page from Individual Result
Currently the only way to edit an evaluation is by doing so on Database load tab. Having the edit results button appear in the top right corner would be very beneficial.

Screen Shot 2021-02-04 at 2 45 51 PM

camdenmoors:
Closes #886 and supersedes #1155, adds an edit tag to the parent component of EvaluationInfo.vue.


test/integration/database-results.spec.ts
```diff
@@ -9,6 +9,15 @@ import DataTableVerifier from '../support/verifiers/DataTableVerifier';
...
 import ToastVerifier from '../support/verifiers/ToastVerifier';
 
+// There seems to be an issue with v-slide-group that causes it to throw
+// "Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined"
+// On a null-obj
+Cypress.on('uncaught:exception', (err, _runnable) => {
+  console.log(err);
+  // Returning false here prevents Cypress from failing the test
+  return false;
+});
```
rbclark, line 19: This should be slightly less aggressive about catching every possible exception, it should only catch the 'getBoundingClientRect' error.

apps/frontend/src/components/global/sidebaritems/SidebarFileList.vue
```diff
@@ -116,17 +118,16 @@ export default class FileItem extends mixins(ServerMixin, RouteMixin) {
     }
     axios
       .post('/evaluations', formData)
-      .then((response) => {
+      .then(() => {
         SnackbarModule.notify('File saved successfully');
-        file.database_id = parseInt(response.data.id);
```
rbclark, line 121: How do we get the ID of the file if it is no longer stored here?

apps/frontend/src/components/global/sidebaritems/SidebarFileList.vue
```diff
@@ -116,17 +118,16 @@ export default class FileItem extends mixins(ServerMixin, RouteMixin) {
...
       })
       .catch((error) => {
         SnackbarModule.failure(error.response.data.message);
       }).finally(() => {
         this.saving = false;
+        this.saved = true;
```
rbclark, line 128: Is it actually saved if the request failed?

rbclark (changes requested):

apps/frontend/src/components/global/sidebaritems/SidebarFileList.vue
camdenmoors: This was originally to prevent the user from editing their local copy of the uploaded profile as it would should empty fields when you tried to edit the thing you just uploaded, however, adding the created IEvaluation just lets the user edit it after saving.

apps/frontend/src/views/Results.vue
```diff
@@ -289,6 +319,16 @@ export default class Results extends mixins(RouteMixin, ServerMixin) {
     return this.is_result_view ? this.evaluationFiles : this.profiles;
   }
 
+  getEvaluation(file: SourcedContextualizedEvaluation): IEvaluation | undefined {
+    let result: IEvaluation | undefined;
+    EvaluationModule.allEvaluations.forEach((e) => {
```
rbclark, line 324: Use Array.find() here instead.

apps/frontend/src/views/Results.vue
```diff
@@ -65,6 +65,32 @@
                     @click="toggle_profile(file)"
                   >
                     <EvaluationInfo :file="file" />
+                    <v-card-actions>
+                      <div
+                        v-if="
+                          file.from_file.database_id &&
+                          getEvaluation(file).editable
+                        "
+                        class="top-right"
```
rbclark, line 74: Does vuetify already have any builtin CSS for this?

test/integration/database-results.spec.ts
```diff
@@ -9,6 +9,17 @@ import DataTableVerifier from '../support/verifiers/DataTableVerifier';
...
+// There seems to be an issue with v-slide-group that causes it to throw
+// "Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined"
+// On a null-obj
+Cypress.on('uncaught:exception', (err, _runnable) => {
+  if (err.message.includes('getBoundingClientRect')) {
+    // Returning false here prevents Cypress from failing the test
+    return false;
+  }
+  return true;
+});
```
rbclark, line 21: This is in `index.ts` on master, it shouldn't be necessary here.

apps/frontend/src/views/Results.vue
```diff
@@ -355,6 +395,12 @@ export default class Results extends mixins(RouteMixin, ServerMixin) {
     this.search_term = '';
   }
 
+  mounted() {
+    if(this.serverMode) {
+      EvaluationModule.getAllEvaluations()
+    }
+  }
```
rbclark, line 402: I'm still a bit iffy on this change, I'd rather at least investigate if we can load only the evaluations we need instead of all of them.

rbclark (changes requested):

rbclark (approved):

PR #1178, opened 2021-04-01
Move UploadNexus to Full Page

### LINKED ISSUE #1022: Restructure database view to be a dedicated page (like Heimdall Server 1.0)
The database tab in the Upload Modal is quickly outgrowing it's current location.

Screen Shot 2021-03-03 at 10 43 01 AM

With tags added and a reasonable size filename, the view is pretty much out of space. Now that groups are added, it is going to make this view even more squished together. It has gotten to a point where it may as well be a dedicated page (similar to how the groups management page looks)

**A few things to consider:**
 * When the load button is clicked, should the user be taken to basically a fullscreen version of the current upload modal, or should some other approach be taken?
 * May be worthwhile to revisit how this works on Heimdall 1 and see what worked and what didn't with that design.

Better ability to quickly identify information about an evaluation.

camdenmoors:
Closes #1022, converts UploadNexus into a full page.


rbclark (changes requested):
For this one I would really prefer to at least test out using a proper transition to fullscreen for that modal for that tab.

rbclark (changes requested):
(Disregard, comment was on wrong issue)

rbclark (approved):

PR #1182, opened 2021-04-02
Bucket 'Waived' Controls as Waived as we do for 'Profile Errors'

### LINKED ISSUE #410: Bucket 'Waived' Controls as Waived as we do for 'Profile Errors'
Using the HDF backend, filter the controls array base on the waived boolean prop and bucket and filter any controls that have been Waived as such rather than putting them as Not Applicable.

Not Applicable is a different business decision so I want to clearly distinguish between the two if an end user chooses to Waive controls.

camdenmoors:
Closes #410. 

Adds: 
- Taking waived profiles into account when calculating compliance
- Filter by Waived Tests
- A skip message to ControlRowCol results


apps/frontend/src/components/cards/ComplianceChart.vue
```diff
@@ -93,7 +93,8 @@ export default class ComplianceChart extends Vue {
       passed +
       StatusCountModule.countOf(this.filter, 'Failed') +
       StatusCountModule.countOf(this.filter, 'Profile Error') +
-      StatusCountModule.countOf(this.filter, 'Not Reviewed');
+      StatusCountModule.countOf(this.filter, 'Not Reviewed') +
+      StatusCountModule.countOf(this.filter,  'WaivedTests');
```
rbclark, line 97: Just call this `Waived` instead.

apps/frontend/src/components/cards/StatusCardRow.vue
```diff
@@ -52,7 +76,7 @@ import {Filter} from '@/store/data_filters';
 import {Prop} from 'vue-property-decorator';
 
 interface CardProps {
-  icon: string;
+  icon?: string;
```
rbclark, line 79: This should no longer be necessary.

apps/frontend/src/components/cards/controltable/ControlRowCol.vue
```diff
@@ -15,9 +15,9 @@
     </v-col>
     <v-col
       cols="12"
-      :sm="result.message ? 6 : 12"
-      :lg="result.message ? 5 : 10"
-      :class="result.message ? 'left' : 'right'"
+      :sm="result.message || result.skip_message ? 6 : 12"
+      :lg="result.message || result.skip_message ? 5 : 10"
+      :class="result.message || result.skip_message ? 'left' : 'right'"
```
rbclark, line 20: Add a computed property that gets `result.message | result.skip_message`.

apps/frontend/src/views/Results.vue
```diff
@@ -262,6 +263,11 @@ export default class Results extends mixins(RouteMixin) {
    */
   status_filter: ControlStatus | null = null;
 
+  /**
+   * Determines if we should filter to only waived controls
+   */
+  waived_filter = false;
```
rbclark, line 269: Instead change the type of `status_filter` to `ControlStatus | 'Waived' | null = null`.

rbclark (changes requested):

apps/frontend/src/views/Results.vue
```diff
@@ -135,8 +136,8 @@
                 <ComplianceChart :filter="all_filter" />
               </v-card-actions>
               <v-card-text style="text-align: center"
-                >[Passed/(Passed + Failed + Not Reviewed + Profile Error) *
-                100]</v-card-text
+                >[Passed/(Passed + Failed + Not Reviewed + Profile Error +
+                Waived) * 100]</v-card-text
```
rbclark, line 140: Only show the words `+ Waived` if there are any Waived controls.

apps/frontend/src/components/cards/StatusCardRow.vue
```diff
@@ -112,19 +136,29 @@ export default class StatusCardRow extends Vue {
...
       number: StatusCountModule.countOf(filter, 'Profile Error')
     };
   }
+
+  get waivedProfiles(): CardProps | null {
+    // Want to ignore existing status filter
+    const count = StatusCountModule.countOf(this.filter, 'WaivedTests');
+    return {
+      title: 'Waived Tests',
+      subtitle: `Has ${count} waived tests.`,
```
rbclark, line 158: `Consider using an overlay or manual attestation to properly address this control.`

apps/frontend/src/components/cards/StatusCardRow.vue
camdenmoors: It seems this is still used by the Passed/Failed/Not Applicable/Not Reviewed cards

rbclark (changes requested):
Getting

```
[Vue warn]: Property or method "card" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.
```

on this one and the waivers and profile errors are missing.

apps/frontend/src/components/cards/StatusCardRow.vue
rbclark: In that case I'd rather just pass the icon in as well for consistency.

rbclark (changes requested):
This has an issue where filtering to `Waived` behaves differently than filtering by `Profile Errors`. Filtering by Profile Errors hides the waived warning (which makes sense since they are filtered out). Filtering by Waived on the other hand doesn't disable the button to show that it is filtered, and still shows the Profile Error message:

issueWithWaived

rbclark (approved):

PR #1190, opened 2021-04-05
Refactor ExportCAAT

### LINKED ISSUE #1107: CAAT doesn't distinguish multiple result sets
If more than one of the same type of result set are loaded for viewing (having common control ids), when exporting to CAAT only one set of information for each control id is listed, as opposed to each. This can be seen by loading the NGINX samples or red hat samples.

To distinguish each result set in the CAAT, it is safe to use the (as yet unused by us) 4th column "D": "Finding ID" to populate with the following information:

<filename> - Test <control_id>

e.g.,: my_rhel7_03012021.json - Test V-71999

camdenmoors:
Closes #1107, separates multiple profiles into different excel workbooks.


ejaronne:
How is the unique identifier(?) determined?
image

ejaronne:
problem: 
- Worksheet names cannot be more than 31 characters
- If you export only one result set it simply truncates the filename for the worksheet name
- If you export more than one result set AND the filenames are the same or the first 30 letters are the same, I think it attempts to use the full filename and tack on the unique identifier e.g., a9e1
- You can see this by loading up the "OWASP ZAP Zero_WebAppSecurity Heimdall_tools Sample" twice and attempting to output the CAAT:
image

- Maybe using the approach done by the "NIST SP 800-53 Security Control Coverage" Export:
image

But this still presents a problem: If I have a run against more than one target and/or dates, environments, systems, the beginning 30 characters of my results file json may all be the same, only the characters after that may be different:
image

Therefore:
Let the tabs be "OWASP_ZAP_Zero_WebAppSecurity_H" and "OWASP_ZAP_Zero_WebAppSecuri (2)"

But *PLEASE* populate the worksheet's 4th column "D": "Finding ID" with the following information:

<filename> - Test <control_id>

e.g.,: OWASP_ZAP_Zero_WebAppSecurity_Heimdall_tools_04092021_Dev_App1 - Test V-71999

Thanks!

ejaronne (changes requested):
See comments
[longfilenametestsamples.zip](https://github.com/mitre/heimdall2/files/6288925/longfilenametestsamples.zip)

rbclark (approved):
As long as Eugene is fine with the output then this seems fine.

ejaronne (approved):
[CAAT-04-15-2021_test.xlsx](https://github.com/mitre/heimdall2/files/6322177/CAAT-04-15-2021_test.xlsx)
Looks good. Thanks!

PR #1143, opened 2021-03-23
Vertically Slim UI, Enhance Focus, Pin ResponsiveRowSwitch and ControlRowHeader

camdenmoors:
Closes #104, changes:

- [x] Pin = For example, when I expand a row for a result (e.g., V-#), "pin" the row header at the top of the results view. That way, if I scroll down to look at test results or details or especially that very tall code block, I can still recall wth V-# I'm reviewing. 
- [x] Vertically slim = reduce the vertical padding. everywhere. to almost nothing. change the paradigm to "the functionality is there, now let's add a little padding here and there".

- [x] Focus = for example, when binking on the results view table,  slide the top of the table to the top of the page. If I'm going to look at stuff there, I don't plan to work down in 1/4 of the bottom half of the page. Slide it up so I can see what I'm after!

- [x] Datatable - Remove "Single Expand" toggle (Uncertain if we are still doing this). The "Status, ID, Severity, Title, and 800-53 Controls & CCIs" should stick to the top of the page as the user scrolls, and the currently expanded control heading should also stick if it is expanded.

PR #1196, opened 2021-04-06
Remove all references to Babel as a dependency

### LINKED ISSUE #1171: NOTE: babel-eslint is now @babel/eslint-parser and has moved into the Babel monorepo.
Need to update this dependency to point to the new location.

rbclark:
Since IE is no longer supported, there is not much use in including babel.

Fixed #1171


camdenmoors (approved):

camdenmoors:
This, #1197, and #1203 seem to be failing due to a cypress bug that only occurs when running via headless tests introduced in #1143, trying to find a fix for it.

PR #1225, opened 2021-04-13
Allow for Disabling User Registration

camdenmoors:
Adds:

 - `DISABLE_REGISTRATION` .env variable
 - `registrationEnabled` to StartupSettingsDto
 - `/authn/signup` (unauthenticated) vs `/users` (authenticated) endpoints
 - Creating users from admin panel

Keeping as draft until I've written tests.


apps/backend/src/users/users.controller.spec.ts
```diff
@@ -113,7 +113,8 @@ describe('UsersController Unit Tests', () => {
       expect.assertions(1);
 
       const createdUser = await usersController.create(
-        CREATE_USER_DTO_TEST_OBJ_2
+        CREATE_USER_DTO_TEST_OBJ_2,
+        {user: adminUser}
```
rbclark, line 117: Does this mean that the test suite always runs with registration disabled now?

camdenmoors: Registration is not disabled for the test suite, the problem here is that the only way I seem to be able to get the user's information from the request (to check if they are an admin) is to use `JwtAuthGuard`, however, using that means requests have to be authenticated. So I added `POST /authn/signup` which is the non-authenticated version of `POST /users`, which will be disabled when configured in .env.

apps/backend/src/users/users.controller.ts
```diff
@@ -71,8 +75,27 @@ export class UsersController {
...
+        ForbiddenError.from(abac)
+          .setMessage(
+            'User registration is disabled. Please ask your system administrator to create the account.'
+          )
+          .throwUnlessCan(Action.ReadAll, User);
+        return new UserDto(await this.usersService.create(createUserDto));
+      }
+      throw new ForbiddenException(
+        'User registration is disabled. Please ask your system administrator to create the account.'
+      );
```
rbclark, line 97: Can this be restructured to not duplicate the call to `usersService.create` and the `ForbiddenException`?

rbclark (changes requested):

rbclark (approved):

PR #1240, opened 2021-04-19
Add Title and Role to Group Users Table

### LINKED ISSUE #1026: Show users group role on the edit page
It is currently impossible to tell a users role in a Group

Screen Shot 2021-03-03 at 11 17 03 AM

A column should be added called `Role` which shows the user's role in the group.

Note that the backend doesn't currently return this data so this will require backend changes as well.

camdenmoors:
Closes #1026. Adds the Title and Role of all users to the "Update Group" modal.
Screen Shot 2021-05-18 at 9 50 15 AM


apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -123,8 +131,11 @@ export default class Users extends Vue {
 
     ServerModule.allUsers.forEach((user) => {
       if(!currentUserIds.includes(user.id)) {
+        const text = user.title !== null
+          ? `${user.firstName || ''} ${user.lastName || ''} (${user.email}, ${user.title})`
+          : `${user.firstName || ''} ${user.lastName || ''} (${user.email})`
```
rbclark, line 136: Instead of building the string twice here, can you do:

```javascript
let text = `${user.firstName || ''} ${user.lastName || ''} (${user.email}, ${user.title})`
if(user.title !== null) {
	text += `, ${user.title})`;
}
```

apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -121,10 +129,14 @@ export default class Users extends Vue {
     const currentUserIds: string[] = this.currentUsers.map((user) => user.id);
     const users: IVuetifyItems[] = [];
 
-    ServerModule.allUsers.forEach((user) => {
+    ServerModule.allUsers.forEach(async (user) => {
       if(!currentUserIds.includes(user.id)) {
+        let text = `${user.firstName || ''} ${user.lastName || ''} (${user.email})`
+        if (user.title !== null) {
+          text = text.slice(0, -1) + `, ${user.title})`;
```
rbclark, line 136: Isn't `text.slice(0, -1) === text`?

rbclark (changes requested):
Title seems to be a new database field here correct? I don't see an accompanying migration that actually adds it.

camdenmoors:
> Title seems to be a new database field here correct? I don't see an accompanying migration that actually adds it.

This doesn't require any changes to the database as it just extends the GroupsDTO.

rbclark (approved):

PR #1246, opened 2021-04-22
Use Postgres Session Store

### LINKED ISSUE #1020: [BUG] connect.session() MemoryStore is not designed for a production environment
**Describe the bug**

```
server_1    | Warning: connect.session() MemoryStore is not
server_1    | designed for a production environment, as it will leak
server_1    | memory, and will not scale past a single process.
```

**To Reproduce**
1. Start Heimdall in Docker
2. View the logs
3. See error

**Expected behavior**
I believe sessions are only used currently for passing information to the frontend. The issue is, if anyone is running multiple frontend containers or multiple backend processes, the session information will be lost.

This will most likely only cause problems in the short-term if someone deploys Heimdall in a load-balanced manner.

**Possible solutions**
There are multiple compatible session stores: https://www.npmjs.com/package/express-session#compatible-session-stores

Since Heimdall server is already using postgres, that may be an attractive alternative.

camdenmoors:
Closes #1020, stores session information in database instead of MemoryStore.


apps/backend/config/app_config.ts
```diff
@@ -69,6 +69,15 @@ export default class AppConfig {
...
   }
 
+  getDbURL() {
+    let url = this.get('DATABASE_URL');
+    if (url === undefined) {
+      const dbConfig = this.getDbConfig();
+      url = `pg://${dbConfig.username}:${dbConfig.password}@${dbConfig.host}:${dbConfig.port}/${dbConfig.database}`
+    }
+    return url
+  }
```
rbclark, line 79: I would prefer if we didn't have to have a function to figure out the DATABASE_URL and could instead standardize on using the `getDbConfig` object everywhere if it is supported.

apps/backend/src/main.ts
```diff
@@ -47,7 +48,12 @@ async function bootstrap() {
   app.use(
     session({
       secret: configService.get('JWT_SECRET') || generateDefault(),
-      saveUninitialized: false,
+      store: new (postgresSessionStore(session))({
+        conString: configService.getDbUrl(),
```
rbclark, line 52: The `postgresSessionStore` apparently takes a `conObject` option as well. Could you see if it is compatible with our `getDbConfig` object?

rbclark (changes requested):

apps/backend/src/config/config.service.ts
```diff
@@ -30,6 +30,10 @@ export class ConfigService {
     return this.appConfig.getDbConfig();
   }
 
+  getDbUrl(): string {
+    return this.appConfig.getDbURL();
+  }
```
rbclark, line 35: This seems to no longer be needed.

apps/backend/config/app_config.ts
```diff
@@ -69,6 +71,15 @@ export default class AppConfig {
...
```
rbclark, line 81: This seems to no longer be needed.

rbclark (changes requested):

PR #1274, opened 2021-04-27
Export HTML Reports for Admins, Managers and Users

### LINKED ISSUE #1242: HTML Export
An export option to have a single deliverable page would be beneficial. There should be 3 different options:

An Executive Report:
 - Profile Info
 - Statuses
 - Status Doughnuts
 - Heatmap (Static)

A Manager Report:
 - Everything from executive
 - Results
 - Collapsible by row
 - Test results and details (no code, side by side)

An Administrator Report:
 - Everything from manager
 - Test results, details (side by side), and code (below results and details)
---
It should:
 - Be done within the frontend (Selenium or similar could be used to create a report in CI runs). 
 - Have an `Expand All` button to expand all results sets and results to make it easy to print for Manager/Administrator reports .

camdenmoors:
Closes #1242, adds HTML export to frontend.


camdenmoors:
I had linted the projected to get CI to pass, however it seems that cci_utils was updated along with it, and now SonarCloud is complaining that all the mapped lines are too long. Can you add `cci_util.ts` to the ignore list @rbclark?

apps/frontend/public/static/export/style.css
```diff
@@ -0,0 +1,7 @@
+/*
+      * Bootstrap v5.0.0-beta3 (https://getbootstrap.com/)
+      * Copyright 2011-2021 The Bootstrap Authors
+      * Copyright 2011-2021 Twitter, Inc.
+      * Licensed under MIT (https://github.com/twbs/bootstrap/blob/main/LICENSE)
+*/
```
rbclark, line 6: Is this file still necessary if we're getting bootstrap from a package?

rbclark (changes requested):

rbclark (changes requested):
There seems to be a lot of duplication between the different reports here, and this doesn't seem to be using a templating engine to actually generate the resulting file. It would be a bit cleaner to use something such as [mustache.js](https://mustache.github.io/) to generate the reports.

aaronlippold:
I would have expected a set of template 'sectoions' or stubs for sub-parts written in mustash .. something like:
```
stubs/header
stubs/footer
stubs/counts
stubs/parts/reults/row .. 
```
etc

aaronlippold:
So that with the base part of the template, then v1, v2 and v3, of the report would just add more stubs for the more advanced reports.

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,294 @@
...
+    <div class="container-fluid pt-4" id="statusCards">
+      <div class="row">
+        <div class="col-md">
+          <div class="card bg-success tall-card">
+            <div class="card-body">
+              <div class="card-title">
+                <div class="h4">
+                  <svg style="width:24px; height:24px" viewBox="0 0 24 24">
+                    <path fill="white" d="M12 2C6.5 2 2 6.5 2 12S6.5 22 12 22 22 17.5 22 12 17.5 2 12 2M10 17L5 12L6.41 10.59L10 14.17L17.59 6.58L19 8L10 17Z" />
+                  </svg>
```
rbclark, line 129: Can we just bundle @mdi/font similar to how we are bundling bootstrap so that it is more obvious for future developers that these are icons?

rbclark (changes requested):
Really liking how this is looking with the new template setup!

apps/frontend/vue.config.js
```diff
@@ -12,6 +12,17 @@ const changelog = parsed.changelog || '';
 const branch = parsed.branch || '';
 const issues = parsed.issues || '';
 
+// This grabs the css for Bootstrap to allow for HTML export
+fs.copyFile(
+  require.resolve('bootstrap/dist/css/bootstrap.css'),
```
rbclark, line 17: Should this be the `bootstrap.min.css`?

apps/frontend/vue.config.js
```diff
@@ -12,6 +12,17 @@ const changelog = parsed.changelog || '';
 const branch = parsed.branch || '';
 const issues = parsed.issues || '';
 
+// This grabs the css for Bootstrap to allow for HTML export
+fs.copyFile(
```
rbclark, line 16: Would this method work for the icons as well? It looks like `@mdi/font` is just a webfont and a CSS file.

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,294 @@
...
+          </div>
+        </div>
+        <div class="col-md">
+          <div class="card bg-warning tall-card">
+            <div class="card-body">
+              <div class="card-title">
+                <div class="h4">
+                  <svg style="width:24px;height:24px" viewBox="0 0 24 24">
+                    <path fill="white" d="{{icons.circleAlert}}" />
+                  </svg>
```
rbclark, line 174: If the whole webfont is available as part of the page that may simplify actually adding the icons to this page. It would be nice if this was something similar to `<i class="{{icons.circleAlert}}" />`. If that isn't possible with the current state, it looks like the CSS may make something like `<i class="mdi-icon-name" />` possible.

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,294 @@
...
+        <a class="navbar-brand" href="#">Heimdall Report</a>
+        <button class="navbar-toggler" type="button" data-bs-toggle="collapse" data-bs-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation"> <span class="navbar-toggler-icon"></span> </button>
+        <div class="collapse navbar-collapse" id="navbarSupportedContent">
+          <ul class="navbar-nav me-auto mb-2 mb-lg-0">
+            <li class="nav-item"> <a class="nav-link active" aria-current="page" href="#">Profile Info</a> </li>
+          </ul>
+          <div class="d-flex">
+            <a id="generate" onclick="$(this).attr('href', 'data:text/plain;base64,' + utf8_to_b64($('html').clone().find('#generate').remove().end()[0].outerHTML));" download="report.html">
+              <svg style="width:24px;height:24px" viewBox="0 0 24 24" onclick="savePage()">
+                <path fill="currentColor" d="M5,20H19V18H5M19,9H15V3H9V9H5L12,16L19,9Z" />
```
rbclark, line 96: Is this icon part of `@mdi/font`?

rbclark (changes requested):

apps/frontend/vue.config.js
camdenmoors: There doesn't seem to be a way to embed the webfont inside of the HTML. I've been trying to embed the icons as SVG in CSS but that isn't working out so far.

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,284 @@
...
+      .tall-card {
+      height: 190px;
+      }
+      }
+      @media (min-width: 1886px) {
+      .tall-card {
+      height: 130px;
+      }
+      }
+    </style>
```
rbclark, line 44: Can this CSS be formatted to be more readable?

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,284 @@
...
+    </script>
+  </head>
+  <body>
+    <nav class="navbar fixed-top navbar-expand-lg navbar-light bg-light">
+      <div class="container-fluid">
+        <a class="navbar-brand" href="#">Heimdall Report</a>
+        <button class="navbar-toggler" type="button" data-bs-toggle="collapse" data-bs-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation"> <span class="navbar-toggler-icon"></span> </button>
+        <div class="collapse navbar-collapse" id="navbarSupportedContent">
+          <ul class="navbar-nav me-auto mb-2 mb-lg-0">
+            <li class="nav-item"> <a class="nav-link active" aria-current="page" href="#">Profile Info</a> </li>
```
rbclark, line 91: I'm not sure what the intent of this link is.

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,284 @@
...
+          </ul>
+          <div class="d-flex">
+            <a id="generate" onclick="savePage()" download="report.html">
+              {{{icons.download}}}
+            </a>
```
rbclark, line 96: Is it possible to hide the downloaded link on the actual downloaded version?

rbclark, line 96: The download link doesn't show the appropriate cursor when it is hovered over.

rbclark (changes requested):
Expanding and collapsing is not currently working

clicknotworking

apps/frontend/public/static/export/template.html
camdenmoors: This was intended to jump to the top alongside the ability to jump between control sets, however, the library to do this seems to be causing the copied Apex charts to disappear.
image

aaronlippold:
Update the PR title with the actual feature - such as 'Export HTML Reports for Admins, Managers and Users'

rbclark (changes requested):
This doesn't seem to be working correctly at the moment. On the server preview I am getting 

Screen Shot 2021-06-14 at 11 06 15 AM

and on the deploy preview I am getting 

Screen Shot 2021-06-14 at 11 11 07 AM

camdenmoors:
I've hit a little bit of a roadblock here. This works fine locally due to the content security policy not kicking in. However once actually running on a webserver, inline javascript will be blocked (needed for collapsing and page downloading) and it can't be re-enabled as we can't set headers for a locally generated page.

@rbclark Would it work to have the page download as soon as you click the button in Heimdall, instead of keeping the new window open?

image

ejaronne:
Admin report is not showing code for overlays: For example, using triple overlay sample:
In heimdall:
image
versus in the report:
image

ejaronne:
What is this thing in the lower left corner?
image
It shows up in the locally saved version as this sad icon:
image

camdenmoors:
> What is this thing in the lower left corner?

This is part the deploy preview service and won't be in the final version

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -0,0 +1,333 @@
...
+  exportHTML(): void {
+    this.resetOutputData();
+    if(this.filter.fromFile.length === 0){
+      return SnackbarModule.failure('No files have been loaded.')
+    }
+    // Download the template
+    axios.get(`/static/export/template.html`).then(({data}) => {
+      axios.get(`/static/export/style.css`).then((styleResponse) => {
+        axios.get(`/static/export/bootstrap.js`).then((bootstrapJsResponse) => {
+          axios.get(`/static/export/jquery.js`).then((jqueryResponse) => {
```
rbclark, line 297: Does something like what is described here: https://www.storyblok.com/tp/how-to-send-multiple-requests-using-axios make this downloading easier? I'd imagine it will also happen multithreaded if that approach is followed.

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,221 @@
...
+				<ul class="navbar-nav me-auto mb-2 mb-lg-0"> {{#showControlSets}}
+					<li class="nav-item"> <a class="nav-link active" aria-current="page" href="#">Profile Info</a> </li>
+					<li class="nav-item dropdown"> <a class="nav-link dropdown-toggle" href="#" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+              Control Sets
+            </a>
+						<div class="dropdown-menu" aria-labelledby="navbarDropdown"> {{#controlSets}} <a class="dropdown-item" href="#resultSetAccordion-{{fileID}}">{{filename}}</a> {{/controlSets}} </div>
+					</li> {{/showControlSets}} </ul>
+				<div class="d-flex"> <a id="generate" onclick="savePage()">
+              {{{icons.download}}}
+            </a> </div>
```
rbclark, line 70: This is no longer necessary correct?

rbclark (changes requested):

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -0,0 +1,335 @@
...
+
+  /** Converts a string to an array buffer */
+  s2ab(s: string) {
+		const buf = new ArrayBuffer(s.length); //convert s to arrayBuffer
+		const view = new Uint8Array(buf); //create uint8array as viewer
+		for(let i = 0; i < s.length; i++) {
+			view[i] = s.charCodeAt(i) & 0xff; //convert to octet
+		}
+		return buf;
+	}
```
rbclark, line 285: This function seems to be duplicated in every Export function, could it get de-duplicated and moved into the `utilities` folder instead?

apps/frontend/public/static/export/template.html
```diff
@@ -0,0 +1,218 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+	<!-- Required meta tags -->
```
rbclark, line 5: I believe this whole project is space-delimited except this file.

rbclark (changes requested):

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -0,0 +1,327 @@
...
+
+  exportHTML(): void {
+    this.resetOutputData();
+    if(this.filter.fromFile.length === 0){
+      return SnackbarModule.failure('No files have been loaded.')
+    }
+    const templateRequest = axios.get(`/static/export/template.html`)
+    const bootstrapCSSRequest = axios.get(`/static/export/style.css`)
+    const bootstrapJSRequest = axios.get(`/static/export/bootstrap.js`)
+    const jqueryRequest = axios.get(`/static/export/jquery.js`)
```
rbclark, line 287: Can these be all labeled with the same names that they have from the packages so it is clear what they actually are (including the min part) so `style.css should be `bootstrap.min.css`. Also these files should be `.gitignore`'d so they don't accidentally get committed to the repository.

rbclark (changes requested):
Is it possible to test if popups are allowed to avoid this somewhat confusing behavior?

popups

Expand/collapse is expanding/collapsing all headings at once

expandcollapse

rbclark (changes requested):
Getting some weird behavior with this in Firefox with multiple results.

1. The 2nd Result overlaps the first when you use the 'Jump to Result Set' functionality on the top bar.

https://user-images.githubusercontent.com/2308869/124629873-deff7b80-de4f-11eb-8fa2-5526bfc16ab0.mov

2. The last highlighted ApexChart title keeps showing up below the Apex Charts.

https://user-images.githubusercontent.com/2308869/124631097-0c98f480-de51-11eb-9676-4665431b4003.mov

rbclark (changes requested):
This is still loading weird in firefox, loading the report over and over gives a completely different result every time:

https://user-images.githubusercontent.com/2308869/124805054-fbbbb200-df28-11eb-9ee2-74776c4e5e40.mov

camdenmoors:
> This is still loading weird in firefox, loading the report over and over gives a completely different result every time:
> 
>  Screen.Recording.2021-07-07.at.1.39.14.PM.mov

Can you try this without any extensions or in an incognito window? I can't seem to replicate this on my side.

rbclark (approved):

PR #1342, opened 2021-05-13
Better (console) error reporting when inspecjs parsing has an error

Bialogs:
I know his leaves a console statement in there, but this would have saved us much more time debugging a recent parser issue if this was present.

If you can think if a better way to still log the error please just change the PR.

One thing is I don't think the user cares about the in-the-weeds details so I'd like to just leave the useless parser error message that we had before.


rbclark (changes requested):
I’d really prefer this as a console warning at least since it is notifying of a failure.

rbclark (approved):

PR #1346, opened 2021-05-14
Show all subdescriptions on the Descriptions page

### LINKED ISSUE #1305: Show all subdescriptions on the Descriptions page
Not all of the data in the `c.hdf.descriptions` is currently shown on the UI, currently the only data displayed is as follows: https://github.com/mitre/heimdall2/blob/master/apps/frontend/src/components/cards/controltable/ControlRowDetails.vue#L142-L190

The order of the elements in ControlRowDetails matters, so the existing sub-descriptions (`Caveat`, `Rationale`, etc) must stay in their current locations. All of the additional sub-descriptions

This allows showing any sub-descriptions that Heimdall tools may add.

camdenmoors:
Closes #1305, adds all `c.hdf.descriptions` to the details tab.

### DIFF
--- a/apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
+++ b/apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
@@ -138,6 +138,7 @@ export default class ControlRowDetails extends mixins(HtmlSanitizeMixin) {
   }
 
   get details(): Detail[] {
+    const hiddenDescriptions = ["default", "check", "fix"]
     let c = this.control;
     const details: Detail[] = [
       {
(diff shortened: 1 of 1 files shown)


aaronlippold:
There seems to be a small error in that the loops for the remaining sub-descriptions are not skipping if it is the 'default', 'check' or 'fix' description. Something like:

```ruby
descriptions.each do |x|
  next if x.match?(/default|check|fix/)
  add row
end
```
Seems to be needed

aaronlippold:
the other way this could be handled is that we make a stack of descriptions. Then just pop {key, values} off the stack as we add them to the descriptions tab then when we do the last loop to print the remaining {key,value} pairs we won't have to worry about the skip cases, given we have already pop'd them off the stack.

rbclark (approved):

Issue #1364, opened 2021-05-18
[BUG] Adding yourself to the group when creating it marks you ask a member


camdenmoors:
**Describe the bug**
If you invite yourself to a group while creating it, your status will be updated to member and you will loose your owner permissions over it.

**To Reproduce**
1. Go to group management
2. Create a new group
3. Add yourself in the users list
4. Refresh
5. Try to edit the group

**Expected behavior**
You should be able to edit the group.

PR #1381, opened 2021-05-21
Fixed issue where adding yourself to a group would remove ownership r…

samuelsunvold:
…ole #1364

### DIFF
--- a/apps/frontend/src/components/global/groups/Users.vue
+++ b/apps/frontend/src/components/global/groups/Users.vue
@@ -130,7 +130,7 @@ export default class Users extends Vue {
     const users: IVuetifyItems[] = [];
 
     ServerModule.allUsers.forEach(async (user) => {
-      if(!currentUserIds.includes(user.id)) {
+      if(!currentUserIds.includes(user.id) && user.id !== ServerModule.userInfo.id) {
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -166,9 +167,18 @@ export default class GroupModal extends Vue {
...
+        addUserDto = {
+          userId: user.id,
+          groupRole: 'owner'
+        }
+      }
+      else {
+        addUserDto = {
+          userId: user.id,
+          groupRole: 'member'
+        }
```
camdenmoors, line 181: Since users are automatically added to their group, we should be able to hide them from the user select dropdown.

e.g in Users.vue, L133:

```js
if(!currentUserIds.includes(user.id)) {
```
->
```js
if(!currentUserIds.includes(user.id) && user.id !== ServerModule.userInfo.id) {
```

samuelsunvold: I was thinking the same. Kind of redundant to add yourself to group if you just get added when the group is created

rbclark (approved):

PR #1382, opened 2021-05-21
Advanced Search and Interactive Filtering

### LINKED ISSUE #75: Cleanup application filter logic - store all filter data in the Vuex store
Heimdall-lite currently only stores what filters are applied to the application inside of the `Results.vue` component. This makes it significantly harder for other Vuex store components to access the current filters since they have to be passed in from the `Results.vue` component instead of being directly accessible out of the Vuex store. In order to fix this, the application should be refactored to store the active filters as part of a `Store` module. The most likely candidate location to store this is the `FilteredDataModule`, which currently only stores `selected_file_ids`, not filter data. I would prefer extending the `EvaluationFile` type to include a `selected` boolean to identify whether it is currently an active file, instead of storing a list of active file IDs.

### LINKED ISSUE #524: expand search to include all desc fields, tags, code, results and search on multiple items at once
Search seems to be limited to certain fields. Users creating overlays, for instance, may add additional sub-descriptions. Can we make the search cover any part of the test, details, or code text?

Can it also search multiple strings, quoted phrases, and do things like OR and AND?

camdenmoors:
Closes #524, closes #75, uses `search-query-parser` to allow queries like: `SSH status:Failed severity:"high,critical"`.


apps/frontend/src/store/data_filters.ts
```diff
@@ -158,12 +182,24 @@ export class FilteredData extends VuexModule {
     this.SELECT_EVALUATIONS([fileID]);
   }
 
+  @Action
+  public select_exclusive_evaluations(fileIDs: FileID[]): void {
+    this.CLEAR_ALL_EVALUATIONS();
+    this.SELECT_EVALUATIONS(fileIDs);
+  }
```
rbclark, line 189: Instead of duplicating this function can you just pass an array with one item for the spots that use `select_exclusive_evaluation`?

apps/frontend/src/store/data_filters.ts
```diff
@@ -158,12 +182,24 @@ export class FilteredData extends VuexModule {
...
   public select_exclusive_profile(fileID: FileID): void {
     this.CLEAR_ALL_PROFILES();
     this.SELECT_PROFILES([fileID]);
   }
 
+  @Action
+  public select_exclusive_profiles(fileIDs: FileID[]): void {
+    this.CLEAR_ALL_PROFILES();
+    this.SELECT_PROFILES(fileIDs);
+  }
```
rbclark, line 201: Same comment as `select_exclusive_evaluations`

apps/frontend/src/views/Results.vue
```diff
@@ -297,6 +330,58 @@ export default class Results extends mixins(RouteMixin, ServerMixin) {
...
+
+  @Watch('isTyping')
+  onDoneTyping() {
+    this.clear()
+    const searchResult: SearchQuery[] = parse(this.searchTerm)[0];
+    searchResult.forEach((result) => {
+      for (const prop in result) {
+        const include = result[prop].include || ''
+        if(prop === 'status') {
+          this.statusFilter.push(this.capitalizeMultiple(include) as ControlStatus & 'Waived')
```
rbclark, line 352: Isn't this the same as `ExpandedControlStatus`?

apps/frontend/src/views/Results.vue
```diff
@@ -297,6 +330,58 @@ export default class Results extends mixins(RouteMixin, ServerMixin) {
...
+  /** If we are currently showing the search help modal */
+  showSearchHelp = false;
+
+  @Watch('searchTerm')
+  onUpdateSearch(_newValue: string) {
+    if (this.typingTimer) {
+      clearTimeout(this.typingTimer);
+    }
+    this.typingTimer = setTimeout(this.onDoneTyping, 1000);
+  }
```
rbclark, line 342: What was the reason for adding the delay here instead of just using computed properties? Was it too slow with computed properties?

apps/frontend/src/views/Results.vue
```diff
@@ -297,6 +330,58 @@ export default class Results extends mixins(RouteMixin, ServerMixin) {
...
+
+  @Watch('isTyping')
+  onDoneTyping() {
+    this.clear()
+    const searchResult: SearchQuery[] = parse(this.searchTerm)[0];
+    searchResult.forEach((result) => {
```
rbclark, line 348: It may be easier to just have computed properties for these, especially if all the data is going to be contained in the search bar anyway if this is updated to automatically update the search bar when users click on different filters on the page.

rbclark (changes requested):
It looks like the `search-query-parser` package would work better than the `search-parser` package for our needs.

apps/frontend/src/views/Results.vue
camdenmoors: Yeah it was lagging behind when it ran the search every time you pressed a key. Now it just waits until you're finished typing to search.

rbclark:
@camdenmoors Does this also fix https://github.com/mitre/heimdall2/issues/75?

camdenmoors:
> @camdenmoors Does this also fix #75?

Yes it does

apps/frontend/src/components/cards/SeverityChart.vue
```diff
@@ -57,12 +57,26 @@ export default class SeverityChart extends Vue {
...
+        return 'low'
+      case 'medium':
+        return 'medium'
+      case 'high':
+        return 'high'
+      case 'critical':
+        return 'critical'
+    }
+    return 'none'
+  }
```
rbclark, line 72: This code seems rather redundant, I think it exists in at least 1 other place and the list of valid types exists in other places as well: https://github.com/mitre/heimdall2/blob/8c2a8b30c6161769dd2454d3e49a99d5de379141/apps/frontend/src/store/search.ts#L42

It would be great to deduplicate the list of possible severities into 1 location and then import it everywhere else, then have something like this in all of the locations this happens:

```
    const severityTypes: Severity[] = ['none', 'low', 'medium', 'high', 'critical'];
  valueToSeverity(severity: string): Severity {
    if(severity in severityTypes) {
      return severity as Severity;
    } else {
      return 'none';
    }
  }
```

rbclark (changes requested):

apps/frontend/src/store/data_filters.ts
```diff
@@ -355,16 +376,192 @@ export const FilteredDataModule = getModule(FilteredData);
 export function filter_cache_key(f: Filter) {
   // fix the search term
   let new_search: string;
-  if (f.search_term !== undefined) {
-    new_search = f.search_term.trim();
+  if (f.searchTerm !== undefined) {
+    new_search = f.searchTerm.trim();
   } else {
     new_search = '';
   }
```
rbclark, line 383: I believe this is equivalent to `new_search = f.search_term?.trim() || ''`

apps/frontend/src/views/Results.vue
camdenmoors: I believe it require be a lot of duplication to have similar parsers for each filter type, and now that search has its own store I don't think it will affect as much.

apps/frontend/src/store/data_filters.ts
```diff
@@ -353,18 +374,186 @@ export const FilteredDataModule = getModule(FilteredData);
...
+
+export function lowercaseAll(input: string | string[]): string | string[] {
+  if (typeof input === 'string') {
+    return input.toLowerCase();
+  } else {
+    return input.map((string) => {
+      return string.toLowerCase();
+    });
+  }
+}
```
rbclark, line 393: This function doesn't seem to be used anywhere

rbclark (changes requested):
Status filtering breaks whenever a search term is entered.

https://user-images.githubusercontent.com/2308869/125301768-4c505800-e2f9-11eb-9ce2-165132fac041.mov

apps/frontend/src/store/data_filters.ts
```diff
@@ -353,18 +374,186 @@ export const FilteredDataModule = getModule(FilteredData);
...
+              statusFilter.toLowerCase()
+          )
+        );
+      }
+    });
+    return foundControls.filter((c, index) => {
+      return foundControls.indexOf(c) === index;
+    });
+  } else {
+    return [...controls];
```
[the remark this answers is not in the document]
camdenmoors: This was originally done to avoid the read-only type of the original `controls` array, however, typing these filters as returning readonly also works.

apps/frontend/src/store/data_filters.ts
```diff
@@ -353,18 +374,186 @@ export const FilteredDataModule = getModule(FilteredData);
...
+    filter.titleSearchTerms?.forEach((term) => {
+      foundControls.push(
+        ...controls.filter((control) => {
+          return control.hdf.wraps.title?.toLowerCase().includes(term);
+        })
+      );
+    });
+    return foundControls.filter((c, index) => {
+      return foundControls.indexOf(c) === index;
+    });
```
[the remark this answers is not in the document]
camdenmoors: This originally removed duplicated controls but this doesn't apply anymore.

apps/frontend/src/store/data_filters.ts
camdenmoors: This is used in search.ts, moving to there instead.

apps/frontend/src/components/cards/StatusCardRow.vue
```diff
@@ -161,11 +170,29 @@ export default class StatusCardRow extends Vue {
   get waivedProfiles(): CardProps | null {
     return {
       icon: 'alert-circle',
-      title: 'Waived Tests',
```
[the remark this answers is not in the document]
camdenmoors: Adding this back into the card, seems a little off without it.

apps/frontend/src/components/global/SearchBar.vue
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @Watch('searchTerm')
+  onUpdateSearch(_newValue: string) {
+    if (this.typingTimer) {
+      clearTimeout(this.typingTimer);
+    }
+    this.typingTimer = setTimeout(this.onDoneTyping, 250);
+  }
+}
```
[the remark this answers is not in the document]
camdenmoors: Unfortunately even with all the optimizations it doesn't look real-time updating is going to work, especially on a low-power machine. Lowering this delay to 100ms makes it feel closer to real-time without causing the app to lag.

apps/frontend/src/store/data_filters.ts
```diff
@@ -353,18 +374,126 @@ export const FilteredDataModule = getModule(FilteredData);
...
+      return control.hdf.raw_nist_tags?.some((tag) =>
+        filter.cciIdFilter?.some((term) =>
+          tag.toLowerCase().includes(term.toLowerCase())
+        )
+      );
+    });
+  } else {
+    return controls;
+  }
+}
```
rbclark, line 484: All of these functions are basically exactly the same, can you create a helper function that takes 3 parameters, the item to filter and the array result to call it on, and just call it for all of these?

Basically the body of this could become

```javascript
return filter_controls_by('full_code', controls, filter.codeSearchTerms)
```

rbclark (changes requested):
Screen Shot 2021-07-12 at 12 59 28 PM

I kept getting confused how all results were filtered out when it was showing not applicable and not reviewed controls as still active. Perhaps if the filters have removed all of a type of control it should also grey out? Open for suggestions on this one.

apps/frontend/src/store/data_filters.ts
```diff
@@ -353,18 +368,67 @@ export const FilteredDataModule = getModule(FilteredData);
...
+            item.toLowerCase().includes(term.toLowerCase())
+          );
+        } else {
+          return newFilters[filter]?.some((term) =>
+            item.some((value) =>
+              value.toLowerCase().includes(term.toLowerCase())
+            )
+          );
+        }
+      });
```
rbclark, line 410: The check here seems inverted, for both code paths you call `return newFilters[filter]?.some((term) =>`, only the inner statement requires an if else.

rbclark (changes requested):

apps/frontend/src/components/cards/StatusCardRow.vue
```diff
@@ -101,44 +110,44 @@ export default class StatusCardRow extends Vue {
...
-          this.filter,
+          {fromFile: this.filter.fromFile},
           'FailedTests'
         )} failed out of ${StatusCountModule.countOf(
-          this.filter,
+          {fromFile: this.filter.fromFile},
           'PassingTestsFailedControl'
         ) + StatusCountModule.countOf(
-          this.filter,
+          {fromFile: this.filter.fromFile},
```
rbclark, line 132: Was `omit_overlayed_controls` passed in here previously? Should it still be passed in now? Can you just build a object with the data you need from `this.filter` and pass that to all of these instead of creating slightly different objects for each?

apps/frontend/src/components/global/SearchBar.vue
```diff
@@ -0,0 +1,107 @@
...
+  typingTimer = setTimeout(() => {return}, 0);
+
+  /**
+   * Handles focusing on the search bar
+   */
+  showSearch(): void {
+    this.showSearchMobile = true;
+    this.$nextTick(() => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (this.$refs.search as any).focus();
```
rbclark, line 76: This shouldn't need to be `any`, Vue has a type for elements.

apps/frontend/src/components/global/tags/TagRow.vue
```diff
@@ -58,7 +58,7 @@ import DeleteDialog from '@/components/generic/DeleteDialog.vue';
   }
 })
 export default class TagRow extends Vue {
-  @Prop({required: true, type: Object}) readonly evaluation!: IEvaluation;
```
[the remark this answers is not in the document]
rbclark: It looks like the other changes were reverted here but evaluation is no longer marked as required, if all TagRow's are wrapped in v-ifs then you can make this required again.

apps/frontend/src/store/search.ts
```diff
@@ -0,0 +1,550 @@
...
+    this.parseSearch();
+  }
+
+  @Action
+  removeSeveritySearch(severity: Severity) {
+    this.removeSeverity(severity);
+    if (this.severityFilter.length !== 0) {
+      // If we still have any severity filters
+      const newSearch = this.searchTerm.replace(
+        /severity:"(.*?)"/gm,
```
rbclark, line 290: `(.*?)` === `(.*)`, the `*` means "0 or more"

apps/frontend/src/store/search.ts
```diff
@@ -0,0 +1,550 @@
...
+
+  /** Adds or replaces severity search in the search bar */
+  @Action
+  addSeveritySearch(severity: Severity) {
+    // If we already have search filtering
+    if (this.searchTerm.trim() !== '') {
+      // If our current filters include severity
+      if (this.searchTerm.toLowerCase().indexOf('severity:') !== -1) {
+        const newSearch = this.searchTerm.replace(
+          /severity:"(.*?)"/gm,
```
rbclark, line 265: `(.*?)` === `(.*)`, the `*` means "0 or more"

apps/frontend/src/store/search.ts
```diff
@@ -0,0 +1,550 @@
...
+  @Action
+  addDescriptionSearch(description: string) {
+    if (this.searchTerm.trim() !== '') {
+      this.context.commit(
+        'SET_SEARCH',
+        `${this.searchTerm} AND description:${description}`
+      );
+    } else {
+      this.context.commit('SET_SEARCH', `description:${description}`);
+    }
```
rbclark, line 467: Checking to see if the current searchTerm is set seems like something that could be handled generically once and not like something that should have different code for every part of the search.

apps/frontend/src/store/search.ts
```diff
@@ -0,0 +1,550 @@
...
+
+  /** Adds or replaces code filter in the search bar */
+  @Action
+  addCodeSearch(code: string) {
+    if (this.searchTerm.trim() !== '') {
+      this.context.commit('SET_SEARCH', `${this.searchTerm} AND code:${code}`);
+    } else {
+      this.context.commit('SET_SEARCH', `code:${code}`);
+    }
+  }
```
rbclark, line 505: A lot of these filters could generically be parameterized if the options to these functions were changed to be

```options: {field: string, value: string}```

**Review by rbclark — CHANGES_REQUESTED** [review]

File: apps/frontend/src/store/search.ts (hunk 18 above)
Comment (camdenmoors) [reply apps/frontend/src/store/search.ts:290]: The ? here is to only match to the end of the first result, because if there is a field after the one currently being edited it will match to the end as well, e.g [this](https://regex101.com/r/7YoLqj/1) vs [this](https://regex101.com/r/oQawzZ/1)

File: apps/frontend/src/components/cards/StatusCardRow.vue (hunk 11 above)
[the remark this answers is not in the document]
Comment (rbclark) [reply apps/frontend/src/components/cards/StatusCardRow.vue:164]: The word 'Tests' is still missing here but your comment seems like you intended to add it back.

File: apps/frontend/tests/unit/Compare.spec.ts [hunk 22]
```diff
@@ -6,6 +6,7 @@ import {shallowMount, Wrapper} from '@vue/test-utils';
 import 'jest';
 import Vue from 'vue';
 import Vuetify from 'vuetify';
+import {SearchModule} from '../../src/store/search';
```
rbclark, line 9: Probably better practice here to just do `@/store/search`

rbclark (approved):

PR #1402, opened 2021-05-25
Add Profile Inputs Tab

### LINKED ISSUE #598: Add Inputs Display to the Profile and Results Views
Now that our profiles can support `sensitive` inputs - we should render the profile input keys and values in the Heimdall UX.

This would be very useful when I am talking to end-users about overlays, inputs in general, and doing walkthroughs.

As an end-user, for a run, I would like to see all the inputs and their values for that scan, and I would also like to be able to filter my results or controls by their use of inputs.

In profiles that have parent and child profiles, it would be very useful to know, if possible, in which profile or dependency the value of the input came from.

camdenmoors:
Closes #598, adds a profile inputs tab alongside the profile info tab.

### DIFF
--- a/apps/frontend/src/components/cards/ProfileInfo.vue
+++ b/apps/frontend/src/components/cards/ProfileInfo.vue
@@ -81,9 +81,9 @@ import _ from 'lodash';
 import TagRow from '@/components/global/tags/TagRow.vue';
 
 interface Attribute {
-  name?: string;
-  options?: {
-    value?: unknown;
+  name: string;
+  options: {
+    value: unknown;
   }
 }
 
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/cards/ProfileInfo.vue
```diff
@@ -48,6 +80,13 @@ import {Prop} from 'vue-property-decorator';
 import _ from 'lodash';
 import TagRow from '@/components/global/tags/TagRow.vue';
 
+interface Attribute {
+  name?: string;
+  options?: {
+    value?: unknown;
+  }
+}
```
rbclark, line 88: Are all of these actually optional or are some of them always present?

rbclark (changes requested):

rbclark (approved):

PR #1419, opened 2021-05-31
Do not hoist backend dependencies in order to slim server image

rbclark:
All of the backend dependencies are required to execute the server image. By not hoisting them it is possible to isolate the packages required and only install those on the server image, limiting the number of dependencies installed and slimming ~300MB from the production image.


camdenmoors (approved):
Builds and runs backend+frontend fine on my end.

PR #1436, opened 2021-06-02
Display group users on 'My Groups' page

samuelsunvold:

Setup back for showing displaying in groups.


apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -132,6 +187,63 @@ export default class GroupManagement extends Vue {
...
+        return true;
+      }
+      else if(window.innerWidth >= 0 && numberOfMembers > 1) {
+        return true;
+      }
+      else {
+        return false;
+      }
+    }
+  }
```
rbclark, line 218: Don't worry about different screen sizes, it makes this a lot more complex and most people will just be using it on a desktop. It collapses down well enough that I can deal with it displaying 4 chips all the time if it means less logic here.

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -62,40 +107,50 @@
...
+    GroupModal,
+    GroupUsers
   }
 })
 export default class GroupManagement extends Vue {
   @Prop({type: Boolean, default: false}) readonly allGroups!: boolean;
 
   editedGroup: IGroup | null = null;
+  selectedGroupUsers: ISlimUser[] | null = [];
+  test: boolean[] = [];
```
rbclark, line 141: This seems to be unused.

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -35,6 +35,51 @@
...
+          {{ getMemberNames(item.users)[2].identifier }}
+        </v-chip>
+        <v-chip
+          :active="displayTotalMembersChip(item.users.length)"
+          pill
+          color="primary"
+          outlined
+          class="ma-1"
+          v-on="on"
+          @click="displayMembersDialog(item.users)"
```
rbclark, line 78: I'm fine with just showing the `displayMembersDialog` all the time instead of filtering the displayMembersDialog down to the specific user they clicked on.

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -35,6 +35,51 @@
...
+          v-if="item.users.length >= 3"
+          :active="displayChip(1200, 3, item.users)"
+          pill
+          color="primary"
+          class="ma-1"
+          v-on="on"
+          @click="displayMembersDialog(item.users[2])"
+        >
+          {{ getMemberNames(item.users)[2].identifier }}
+        </v-chip>
```
rbclark, line 70: It should be possible to de-duplicate this by manipulating the data a little bit more before passing it to the component. If you have an array containing

`['Name 1', 'Name 2', 'Name 3', '+x total']`

then it should be possible to iterate it here instead of re-declaring the `v-chip` every time. There are a few examples where we do this in other parts of the application.

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -132,6 +187,63 @@ export default class GroupManagement extends Vue {
...
+    if(!Array.isArray(members)) {
+      const singleMember: ISlimUser[] = [members];
+      this.selectedGroupUsers = [...singleMember];
+      this.dialogDisplayUsers = true;
+    }
+    else {
+      this.selectedGroupUsers = [...members];
+      this.dialogDisplayUsers = true;
+    }
+  }
```
rbclark, line 245: Always show all the users. You also may not need to make a copy since you aren't changing the data in the popup.

apps/frontend/src/components/global/groups/GroupUsers.vue
```diff
@@ -0,0 +1,77 @@
+<template>
```
rbclark, line 1: This seems very similar to `src/components/global/groups/Users.vue`. Could the same component be reused with some deduplication?

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -132,6 +187,63 @@ export default class GroupManagement extends Vue {
...
+    for(let i = 0; i < users.length; i++) {
+      if(!users[i].firstName && !users[i].lastName) {
+        result[i] = {identifier: users[i].email};
+      }
+      else {
+        result[i] = {identifier: `${users[i].firstName} ${users[i].lastName}`};
+      }
+    }
+    return result;
+  }
```
rbclark, line 232: This should probably be a computed property, also use a map here instead of a for loop.

rbclark (changes requested):

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -35,6 +35,29 @@
...
+            pill
+            color="primary"
+            outlined
+            class="ma-1"
+            @click="displayMembersDialog(item.users)"
+          >
+            {{ `+${item.users.length} total` }}
+          </v-chip>
+        </span>
+      </template>
```
rbclark, line 60: Using the iterator here to display the +x total here feels a little wrong, is it possible to just do `item.users.slice(0, 3)` and then just put an if statement outside of the span but before the end of the template that shows `+x more`?

apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -90,14 +93,24 @@ export default class Users extends Vue {
...
+      this.headers.push({
+        text: 'Actions',
+        value: 'actions',
+        sortable: false,
+      });
+      return this.headers;
+    }
+    else {
+      return this.headers;
+    }
```
rbclark, line 111: This is redundant, you can just always `return this.headers` outside of the if/else statement.

rbclark (approved):

PR #1449, opened 2021-06-07
Add Generic Export to CSV

### LINKED ISSUE #1408: generic csv results export
1st iteration: Provide a generic csv results export with all fields.

2nd iteration: allow user to select which fields and in what order.

3rd iteration: allow choice for excel, to include separate worksheets with each profile info and run information (including inputs)

camdenmoors:
Closes #1408, adds a generic CSV export.


rbclark:
I will defer to @ejaronne to review the output format of this before I perform a code review. It seems a little strange to store JSON in CSV however if this is what Eugene requested then I can deal.

ejaronne:
> I will defer to @ejaronne to review the output format of this before I perform a code review. It seems a little strange to store JSON in CSV however if this is what Eugene requested then I can deal.

Not sure where anyone got the idea of embedding the json within the csv. I did not ask for that. The field output looks pretty good for the first iteration. Please test with both the red hat 7 baseline sample and triple overlay. Make sure the check and fix text populates. Remember check and fix can be tags or sub descriptions. Also respect overlay content for caveats, changes to fields, and showing the code block like we do in Heimdall.

rbclark:
@camdenmoors once this is ready for review again can you tag Eugene and request him to review the output CSV?

camdenmoors:
@ejaronne Can you re-review the csv output please?

aaronlippold:
Please update the PR to describe the actual feature

ejaronne:
When loading the Triple Overlay example, Column J (Code) isn't showing the "stacked" code of the overlays the way Heimdall does: 
image
versus:
image

ejaronne:
Seems to be working okay now, thanks! I did notice a "feature"? When exporting more than one result set at a time (outputs as a zip), I find it won't let me hit the export button more than once. Single results sets yes. Is this to prevent confusion with multiple files being downloaded?

camdenmoors:
> Seems to be working okay now, thanks! I did notice a "feature"? When exporting more than one result set at a time (outputs as a zip), I find it won't let me hit the export button more than once. Single results sets yes. Is this to prevent confusion with multiple files being downloaded?

Ah whoops, fixed. It's supposed to export and then close the modal.

apps/frontend/src/components/global/ExportCSVModal.vue
```diff
@@ -0,0 +1,324 @@
...
+          this.files.forEach((csvFile) => {
+            const buffer = Buffer.from(csvFile.data);
+            zipfile.addBuffer(buffer, csvFile.filename);
+          })
+          zipfile.outputStream.pipe(
+            concat({encoding: 'uint8array'}, (b: Uint8Array) => {
+              saveAs(new Blob([b]), 'exported_csvs.zip');
+            })
+          );
+          zipfile.end();
```
rbclark, line 314: All of this code being inside a for loop doesn't make much sense. The only code that needs to be in the `for` loop is the code that actually populates the `files` array. Once that is populated, write the logic to check if your new array is 1 file or multiple files in length and then provide the user with an appropriate download. Also the files being global to the component is unnecessary since it is only used in this function.

apps/frontend/src/components/global/ExportCSVModal.vue
```diff
@@ -0,0 +1,324 @@
...
+            data: csvString
+          })
+        }
+        // If we have all of our files converted we can save them as a .zip file
+        if(this.filter.fromFile.length !== 1 && this.files.length === this.filter.fromFile.length){
+          const zipfile = new ZipFile();
+          this.files.forEach((csvFile) => {
+            const buffer = Buffer.from(csvFile.data);
+            zipfile.addBuffer(buffer, csvFile.filename);
+          })
```
rbclark, line 308: The Export as JSON component already does this behavior (download 1 file if it is just a json, download a zip if it is an array of files). The code for these 2 components should be de-duplicated, and we should just have a generic helper function that takes an array of files and generates the appropriate output.

rbclark (changes requested):

apps/frontend/src/components/global/ExportCSVModal.vue
camdenmoors: This is in a loop due to objects-to-csv only providing an async function, and javascript doesn't seem to have a callback method on forEach loops. Switched to promises now.

rbclark (changes requested):
This adds a blank line to the export dropdown and seems to break Export JSON:

Screen Shot 2021-07-13 at 8 19 01 AM

The tendency of the dropdown to scroll all the way back to the top of the page makes it quite difficult to quickly uncheck columns:

https://user-images.githubusercontent.com/2308869/125450436-f7ca6a1f-2ba6-4297-80ef-dce78806d643.mov

camdenmoors:
> The tendency of the dropdown to scroll all the way back to the top of the page makes it quite difficult to quickly uncheck columns

This appears to be a bug with Vuetify, having all fields in the v-select empty by default with the data-table having all fields does help with things. The user can then select the fields they want without encountering the scroll bug (until they deselect fields).

rbclark:
> > The tendency of the dropdown to scroll all the way back to the top of the page makes it quite difficult to quickly uncheck columns
> 
> This appears to be a bug with Vuetify, having all fields in the v-select empty by default with the data-table having all fields does help with things. The user can then select the fields they want without encountering the scroll bug (until they deselect fields).

I have a feeling it is an interaction with something else we are doing since this behavior doesn't exhibit in their demo dropdown: https://vuetifyjs.com/en/components/selects/#multiple

rbclark:
The weird redraw behavior seems like it may only happen when the table updates, is there a way to not redraw the table data until the dropdown is closed? Or can we at least verify that the table is the problem?

camdenmoors:
> The weird redraw behavior seems like it may only happen when the table updates, is there a way to not redraw the table data until the dropdown is closed? Or can we at least verify that the table is the problem?

Committing items on closing was the plan but there doesn't seem to be a reliable way to do that. It also seems it's not caused by the table (e.g trying removing any v-model/updaters) still causes it to glitch when you select all items, close and re-open the list, and then uncheck some at the bottom.

The way it is currently avoids this behavior unless you select all items and then deselect some.

rbclark:
Relevant issue: https://github.com/vuetifyjs/vuetify/issues/13959

rbclark (dismissed):
Functionally this now has some weird behavior where it shows 1 row but says it is showing 5, until you click on 5 at which time it displays 5 rows:

https://user-images.githubusercontent.com/2308869/127411606-f8eeb9d3-b495-48fc-b50a-eae5fcd50032.mov

Also, it shows by default all of the rows even though none are enabled:

https://user-images.githubusercontent.com/2308869/127411653-5949a4f6-2d85-4faa-be95-76782d76c126.mov

If it's too much of a pain to fix the 2nd issue I'm fine with just checking all fields at first and waiting on the v-select issue to be addressed by vuetify.

apps/frontend/src/components/global/ExportCSVModal.vue
```diff
@@ -62,14 +62,13 @@ import LinkItem from '@/components/global/sidebaritems/IconLinkItem.vue';
...
-import {HDFControlSegment} from 'inspecjs';
-import {ContextualizedControl} from 'inspecjs/dist/context';
 import {EvaluationFile, ProfileFile} from '../../store/report_intake';
 import {Prop} from 'vue-property-decorator';
 import {InspecDataModule} from '../../store/data_store';
-import {ControlDescription} from 'inspecjs/dist/generated_parsers/v_1_0/exec-json';
 import _ from 'lodash';
 import {saveSingleOrMultipleFiles} from '@/utilities/export_util'
+import {HDFControlSegment, ContextualizedControl} from 'inspecjs';
+import {ControlDescription} from 'inspecjs/src/versions/v_1_0/exec_json';
```
rbclark, line 71: All imports of `inspecjs` should happen directly from `inspecjs` and not from `src`. The `src` path will not exist on the production build of `inspecjs` and even though it would work in this case I'd like to get into the habit of us always using the external interface to inspecjs. In this can you can access it by importing `ExecJSON` and then the type is `ExecJSON.ControlDescription`.

apps/frontend/src/utilities/export_util.ts
```diff
@@ -1,3 +1,39 @@
...
+  filename: string;
+  data: string;
+};
+export async function saveSingleOrMultipleFiles(
+  files: File[],
+  filetype: string
+) {
+  if (files.length === 1) {
+    const blob = new Blob([files[0].data], {
+      type: 'application/csv'
```
georgedias, line 15: The filetype when called from the ExportJson.vue (line 55) is json - shouldn't the type here be set accordantly?

georgedias (changes requested):

georgedias (approved):

PR #1451, opened 2021-06-07
Show Groups on Results View

### LINKED ISSUE #1023: Show groups like we do tags on the database view
Depends on #1022

It is currently impossible to identify why a result is showing up in a users database result list.

A 'Groups' column should be added to the database view.

camdenmoors:
Closes #1023. Adds groups to the EvaluationInfo card within results view.

image


apps/backend/src/evaluations/evaluations.controller.ts
```diff
@@ -72,6 +72,15 @@ export class EvaluationsController {
       abac.can(Action.Read, evaluation)
     );
 
+    // Ensure user has the ability to read associated groups
+    evaluations.forEach((evaluation) => {
+      evaluation.groups = evaluation.groups.filter(
+        (group) =>
+          abac.can(Action.AddEvaluation, group) &&
+          abac.can(Action.RemoveEvaluation, group)
```
rbclark, line 80: Shouldn't this be `Action.Read, group`?

camdenmoors: It seems that Action.Read doesn't filter out access to groups, which groupsForEvaluation seems to avoid as well.

rbclark: If it doesn't it sounds like a bug, I'd rather fix that if it is the case.

camdenmoors: The tests were assuming that public groups could not be read, which I don't think is the correct way to go about things. I've changed it to check Action.Read instead for both `groupsForEvaluation` and `findAll`.

apps/frontend/src/components/cards/EvaluationInfo.vue
```diff
@@ -63,6 +85,21 @@ export default class EvaluationInfo extends Vue {
...
   }
 
+  convertGroupsToIVuetifyItems(groups: IGroup[]): IVuetifyItems[] {
+    return groups.map((group) => {
+      return {
+        text: group.name,
+        value: group.id
+      }
+    })
+  }
```
rbclark, line 95: This code should be de-duplicated with the code in `EditEvaluationModal`, either by putting it in a mixin or using the composition API.

apps/frontend/src/components/cards/EvaluationInfo.vue
```diff
@@ -63,6 +85,21 @@ export default class EvaluationInfo extends Vue {
...
+
+  updateGroupsForEvaluation(): void {
+    if(this.evaluation?.id){
+      axios.get<IGroup[]>(`/evaluations/${this.evaluation?.id}/groups`).then(({data}) => this.groups = this.convertGroupsToIVuetifyItems(data));
+    }
+  }
```
rbclark, line 101: Screen Shot 2021-06-09 at 9 29 46 AM

This ends up causing the request to be duplicated for every page load since very similar code is in the EditEvaluationModal. As groups are added to more locations it may make sense to get ahead of this and store groups in the vuex store on evaluations as well.

Looking back on the original implemention of this, it seems that it would make more sense to just return the group as part of the `EvaluationDto`. With these changes we are requiring the groups in order to group the data by evaluations anyway, we may as well just include it all the time and avoid the need for this call altogether, leveraging the `evaluations.ts` store instead.

apps/frontend/src/components/global/upload_tabs/GroupFileList.vue
```diff
@@ -0,0 +1,228 @@
...
+                    })
+                })
+            } else {
+              if(file.userId === ServerModule.userInfo.id){
+                this.convertedFiles.push({...file, group: 'My Files'});
+              } else if(file.public) {
+                this.convertedFiles.push({...file, group: 'Public Files'});
+              } else {
+                this.convertedFiles.push({...file, group: 'No Group'});
+              }
```
rbclark, line 175: It currently looks a little strange having the word `'group'` in front of My Files, Public Files, and 'No Group', is there any way to remove that for these?

apps/frontend/src/views/Results.vue
```diff
@@ -468,6 +475,13 @@ export default class Results extends mixins(RouteMixin, ServerMixin) {
       this.evalInfo = file;
     }
   }
+
+  onUpdateGroups(newValue: string) {
+    this.updateGroups = newValue;
+    this.$nextTick(() => {
+      this.updateGroups = false;
+    })
+  }
```
rbclark, line 484: Groups being stored and updated in the vuex store would remove the need for this as well I believe.

apps/frontend/src/components/global/upload_tabs/GroupFileList.vue
```diff
@@ -0,0 +1,228 @@
...
+      @confirm="deleteItemConfirm"
+    />
+    <v-data-table
+      v-model="selectedFiles"
+      :headers="headers"
+      :loading="loading"
+      :items="filteredFiles"
+      group-by="group"
+      show-select
+      dense
```
rbclark, line 23: Groups are currently closable which is a little weird, is that possible to disable?

Screen Shot 2021-06-09 at 9 50 37 AM

apps/backend/src/evaluations/evaluations.controller.ts
```diff
@@ -55,10 +55,8 @@ export class EvaluationsController {
   ): Promise<GroupDto[]> {
     const abac = this.authz.abac.createForUser(request.user);
     let evaluationGroups = await this.evaluationsService.groups(id);
-    evaluationGroups = evaluationGroups.filter(
-      (group) =>
-        abac.can(Action.AddEvaluation, group) &&
-        abac.can(Action.RemoveEvaluation, group)
+    evaluationGroups = evaluationGroups.filter((group) =>
+      abac.can(Action.Read, group)
```
rbclark, line 59: Instead of a separate endpoint for groups lets just merge it into the findById for simplicity.

apps/backend/src/evaluations/evaluations.controller.ts
```diff
@@ -72,6 +70,13 @@ export class EvaluationsController {
...
+    // Ensure user has the ability to read associated groups
+    evaluations.forEach((evaluation) => {
+      evaluation.groups = evaluation.groups.filter((group) =>
+        abac.can(Action.Read, group)
+      );
+    });
+
     return evaluations.map(
       (evaluation) =>
         new EvaluationDto(evaluation, abac.can(Action.Update, evaluation))
```
rbclark, line 82: May be able to save a loop over all of the data by performing the filtering in the map below instead of a separate foreach loop.

rbclark (changes requested):

rbclark:
@camdenmoors The anonymous function that sonarcloud is flagging on does actually seem to be a duplicate. Can you go ahead and create a function and call it from both those locations?

apps/frontend/src/components/cards/EvaluationInfo.vue
```diff
@@ -9,32 +9,39 @@
       <div v-if="duration"><strong>Duration:</strong> {{ duration }}</div>
       <div v-if="start_time"><strong>Start Time:</strong> {{ start_time }}</div>
       <div v-if="evaluation" class="d-flex flex-nowrap">
-        <strong class="pt-1 pr-1">Tags:</strong>
+        <strong class="pt-2 pr-1">Tags:</strong>
         <TagRow :evaluation="evaluation" />
       </div>
+      <div v-if="groups.length !== 0" class="d-flex flex-nowrap">
```
rbclark, line 15: Wouldn't `groups.length` be already truthy without the redundant `!== 0`?

apps/frontend/src/components/global/tags/TagRow.vue
```diff
@@ -116,7 +117,9 @@ export default class TagRow extends Vue {
       SnackbarModule.notify("Deleted tag successfully.")
       EvaluationModule.getAllEvaluations().then(() => {
         this.syncEvaluationTags();
+        this.$emit('update')
       });
+      this.$emit('update')
```
rbclark, line 122: One of these seems redundant.

apps/frontend/src/components/global/upload_tabs/EditEvaluationModal.vue
```diff
@@ -136,29 +137,19 @@ export default class EditEvaluationModal extends Vue {
...
   async getGroupsForEvaluation(): Promise<void> {
-    return axios.get<IGroup[]>(`/evaluations/${this.active.id}/groups`).then(({data}) => {
-      this.originalGroups = this.groups = this.convertGroupsToIVuetifyItems(data);
-    });
+    this.originalGroups = this.groups = this.convertGroupsToIVuetifyItems(this.active.groups);
   }
 
   async update(): Promise<void> {
-    Promise.all([EvaluationModule.updateEvaluation(this.activeEvaluation), this.updateGroups()]).then(() => {
+    await Promise.all([EvaluationModule.updateEvaluation(this.activeEvaluation), this.updateGroups()]).then(() => {
```
rbclark, line 149: Awaiting here is going to delay the visibility change. Can you just put the emit in the `then` block?

libs/interfaces/evaluation/evaluation.interface.ts
```diff
@@ -1,10 +1,12 @@
 import {IEvaluationTag} from '..';
+import {IGroup} from '../group/group.interface';
 
 export interface IEvaluation {
   id: string;
   filename: string;
   readonly data?: Record<string, any>;
   evaluationTags: IEvaluationTag[];
+  groups?: IGroup[];
```
rbclark, line 9: Are there any instances left where we don't return the groups with evaluations? I'm just wondering if we can make this no longer optional.

rbclark (changes requested):

rbclark (approved):

PR #1482, opened 2021-06-15
Add API Key Support

### LINKED ISSUE #234: Create API tokens for uploading results via cURL
Heimdall currently uses JWT tokens for authentication between the frontend and backend. These are fine for user accounts, however due to their short expiration time they are not ideal for services such as CI pipelines that upload results back to Heimdall.

An API token which is long standing and very scoped to only allow specific actions (limiting the damage if such a token is leaked).

Documentation on a way to achieve this without an API token.

This will allow users to automate the upload of results into Heimdall server.

camdenmoors:
Closes #234.
 - Allow users to generate API tokens with the edit user modal
 - Adds `APIKeyOrJwtAuthGuard` (currently only on POST /evaluations)`
 - Changes the README to use API keys for file upload via curl


apps/backend/src/authn/apikey.strategy.ts
```diff
@@ -0,0 +1,26 @@
...
+      (apikey: string, done: any) => {
+        const user = User.findOne<User>({where: {apikey: apikey}});
+        if (!user) {
+          return done(null, false);
+        } else {
+          return done(null, user);
+        }
+      }
+    );
+  }
```
rbclark, line 25:
```suggestion
  constructor() {
    super(
      {header: 'Authorization', prefix: 'Api-Key '},
      false,
      (
        apikey: string,
        done: (arg1: null, arg2: Promise<User | null>) => unknown
      ) => {
        const user = User.findOne<User>({where: {apikey: apikey}});
        return done(null, user || false);
      }
    );
  }
```

apps/backend/src/users/user.model.ts
```diff
@@ -50,6 +50,10 @@ export class User extends Model {
   @Column(DataType.STRING)
   encryptedPassword!: string;
 
+  @AllowNull(true)
+  @Column(DataType.STRING)
+  apikey!: string;
```
rbclark, line 55: @Bialogs API keys are currently stored unencrypted in the database. Will this be sufficient or is there wording in the SRG that recommends against this?

If the data is encrypted then it will not be able to simply be hashed, unless we only show the user their API key on generation and tell them to make note since it is not recoverable.

apps/frontend/src/components/global/UserModal.vue
```diff
@@ -124,6 +124,18 @@
...
+              id="apiKey"
+              ref="apiKey"
+              v-model="user.apikey"
+              label="API Key"
+              readonly
+              type="text"
+              append-icon="mdi-refresh"
+              @click:append="changeAPIKey"
+            />
+          </div>
```
rbclark, line 138: Is it possible to provide a nice 'copy' button on this field?

rbclark (changes requested):

apps/backend/src/users/user.model.ts
rbclark: @camdenmoors The actual API keys will need to be updated to only be returned to the frontend once, and then hashed using bcrypt just as a password is.

apps/backend/migrations/20210615141642-add_api_token.js
```diff
@@ -0,0 +1,35 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.createTable('ApiKeys', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.BIGINT
+      },
+      userId: {
```
Bialogs, line 12: This userId is not a foreign key. It needs to reference the users table. Right now its just a number.

Look at `migrations/20201201195932-add_user_id_to_evaluations.js` for an example

aaronlippold:
I don't see the README in this PR

apps/backend/src/apikeys/apikey.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+    const apiKeyToDelete = await this.apiKeyService.findByPkBang(id);
+    const apiKeyOwner = await this.usersService.findById(apiKeyToDelete.userId);
+    const abac = this.authz.abac.createForUser(request.user);
+    ForbiddenError.from(abac).throwUnlessCan(Action.Update, apiKeyOwner);
+    return this.apiKeyService.delete(id);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Patch('/:id')
+  async updateAPIKey(
```
Bialogs, line 62: it is common practice to provide your password when updating properties of your API key

apps/backend/src/apikeys/apikey.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+    const apiKeyToUpdate = await this.apiKeyService.findByPkBang(id);
+    const apiKeyOwner = await this.usersService.findById(apiKeyToUpdate.userId);
+    const abac = this.authz.abac.createForUser(request.user);
+    ForbiddenError.from(abac).throwUnlessCan(Action.Update, apiKeyOwner);
+    return this.apiKeyService.update(apiKeyToUpdate.id, updateApiKeyDto);
+  }
+
+  @UseGuards(JwtAuthGuard)
+  @Put('/:id')
+  async regenerateAPIKey(
```
Bialogs, line 76: what is the purpose of regeneration versus just removing and creating a new one?

Bialogs (changes requested):

apps/backend/src/apikeys/apikey.controller.ts
camdenmoors: This is intended to be a shortcut for if you forgot to copy the key, or the key got leaked and needs to be renewed.

Bialogs:
Once we get approval on how the frontend looks, can we add to the README about API keys on how to generate?

rbclark (changes requested):
The current frontend implementation of this is very confusing. The API Keys field should not be dependent at all on the 'Please provide the current password' field below. Showing the list of API keys without the actual key being displayed is fine, prompting for the current password when the user clicks the `+` button would make more sense. Also:

https://user-images.githubusercontent.com/2308869/124803394-0bd29200-df27-11eb-9bff-8f4cb7eddbd7.mov

apps/backend/src/apikeys/apikey.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      {keyId: id, createdAt: new Date()},
+      this.configService.get('JWT_SECRET') || generateDefault()
+    );
+    const JWTSignature = newJWT.split('.')[2];
+    apiKey.apiKey = await hash(JWTSignature, 14);
+    await apiKey.save();
+    return newJWT;
+  }
+
+  async remove(id: string): Promise<APIKeyDto> {
```
Bialogs, line 104: Why is there remove and delete functions? remove does not seem to be used in controller.

apps/backend/src/apikeys/dto/create-apikey.dto.ts
```diff
@@ -0,0 +1,10 @@
+import {ICreateApiKey} from '@heimdall/interfaces';
+import {IsOptional, IsString} from 'class-validator';
+
+export class CreateApiKeyDto implements ICreateApiKey {
+  readonly name!: string;
+
+  @IsOptional()
+  @IsString()
+  readonly currentPassword!: string;
```
Bialogs, line 9: why have currentPassword defined with `!` but also have it be optional?

apps/backend/src/apikeys/dto/create-apikey.dto.ts
```diff
@@ -0,0 +1,10 @@
+import {ICreateApiKey} from '@heimdall/interfaces';
+import {IsOptional, IsString} from 'class-validator';
+
+export class CreateApiKeyDto implements ICreateApiKey {
+  readonly name!: string;
```
Bialogs, line 5: No validator on this prop.

apps/backend/src/apikeys/dto/apikey.dto.ts
```diff
@@ -0,0 +1,16 @@
+import {IApiKey} from '@heimdall/interfaces';
+import {ApiKey} from '../apikey.model';
+
+export class APIKeyDto implements IApiKey {
+  readonly id!: string;
+  readonly name!: string;
+  createdAt!: Date;
+  updatedAt!: Date;
```
Bialogs, line 8: should be readonly.

apps/backend/src/apikeys/dto/delete-apikey.dto.ts
```diff
@@ -0,0 +1,8 @@
+import {IDeleteApiKey} from '@heimdall/interfaces';
+import {IsOptional, IsString} from 'class-validator';
+
+export class DeleteAPIKeyDto implements IDeleteApiKey {
+  @IsOptional()
+  @IsString()
+  readonly currentPassword!: string;
```
Bialogs, line 7: defined with ! but is optional?

apps/backend/src/apikeys/dto/regenerate-apikey.dto.ts
```diff
@@ -0,0 +1,8 @@
+import {IRegenerateAPIKey} from '@heimdall/interfaces';
+import {IsOptional, IsString} from 'class-validator';
+
+export class RegenerateAPIKeyDto implements IRegenerateAPIKey {
+  @IsOptional()
+  @IsString()
+  readonly currentPassword!: string;
```
Bialogs, line 7: defined with ! but is optional?

apps/backend/src/apikeys/dto/update-apikey.dto.ts
```diff
@@ -0,0 +1,10 @@
+import {IUpdateAPIKey} from '@heimdall/interfaces';
+import {IsOptional, IsString} from 'class-validator';
+
+export class UpdateAPIKeyDto implements IUpdateAPIKey {
+  readonly name!: string;
+
+  @IsOptional()
+  @IsString()
+  readonly currentPassword!: string;
```
Bialogs, line 9: defined with ! but is optional?

apps/backend/src/apikeys/dto/update-apikey.dto.ts
```diff
@@ -0,0 +1,10 @@
+import {IUpdateAPIKey} from '@heimdall/interfaces';
+import {IsOptional, IsString} from 'class-validator';
+
+export class UpdateAPIKeyDto implements IUpdateAPIKey {
+  readonly name!: string;
```
Bialogs, line 5: no validators present; needs validators

apps/backend/src/evaluations/evaluations.controller.ts
```diff
@@ -80,13 +85,17 @@ export class EvaluationsController {
...
 
+  @UseGuards(APIKeyOrJwtAuthGuard)
   @Post()
   @UseInterceptors(FileInterceptor('data'), CreateEvaluationInterceptor)
   async create(
     @Body() createEvaluationDto: CreateEvaluationDto,
     @UploadedFile() data: Express.Multer.File,
     @Request() request: {user: User}
   ): Promise<EvaluationDto> {
+    if (!request.user) {
```
Bialogs, line 96: Why was this check added? What behavior changed between API or JWT guard and simple JWT guard?

apps/backend/src/users/users.service.ts
```diff
@@ -157,7 +157,10 @@ export class UsersService {
     }
   }
 
-  async testPassword(updateUserDto: UpdateUserDto, user: User): Promise<void> {
+  async testPassword(
+    updateUserDto: {currentPassword?: string},
+    user: User
+  ): Promise<void> {
```
Bialogs, line 163: @rbclark 

how do you feel about reuse of testPassword function here? Currently testPassword exists on user.service but now it is also being called as part of apiKey.service. Because of its use in apiKey.service update-api-key.dto includes a currentPassword prop. Now we change the definition of testPassword to not take a user related DTO but just object that has a currentPassword prop.

To me this seems like this more generic implementation should be pulled up to authn service

apps/backend/src/apikeys/apikey.controller.ts
Bialogs: could remove duplicated code on the back end and just call the API in sequence from the front end.

Bialogs (changes requested):
The controller and service methods should follow the pattern established in Users, Groups, Evaluations, etc. where the Dto is only created in the controller and the service only deals with Models.

camdenmoors (commented):
Github is bugging out and making me make a review. Please ignore this message.

apps/backend/src/evaluations/evaluations.controller.ts
camdenmoors: The passport-jwt library seems to throw UnauthorizedException by itself, however the lookup for API keys is handled on our side. The handling for this seems better suited for APIKeyOrJwtAuthGuard though.

apps/backend/src/apikeys/apikey.service.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  async count(): Promise<number> {
+    return this.apiKeyModel.count();
+  }
+
+  async checkKey(apikey: string): Promise<User | null> {
+    try {
+      const jwtPayload = jwt.verify(
+        apikey,
+        this.configService.get('JWT_SECRET') || generateDefault()
```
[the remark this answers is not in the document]
camdenmoors: API keys will now be disabled if JWT_SECRET is not set.

apps/backend/src/authn/apikey.strategy.ts
```diff
@@ -0,0 +1,24 @@
...
+) {
+  constructor(private readonly apiKeyService: ApiKeyService) {
+    super(
+      {header: 'Authorization', prefix: 'Api-Key '},
+      false,
+      async (
+        apikey: string,
+        done: (arg1: null, arg2: Promise<User | null> | boolean) => unknown
+      ) => {
+        return done(null, this.apiKeyService.checkKey(apikey));
```
Bialogs, line 20: instead of setting error to null, this should be set to an unauthorized exception that is raised based on checking the return value of `checkKey` function. This would probably remove the need to have a `handleRequest` method in the guard. If you take a look at local strategy you can see this design in action.

Bialogs (changes requested):

apps/backend/src/authn/apikey.strategy.ts
camdenmoors: Returning an error to to the callback does not cause a request exception, the request just continues with a null User. However, throwing exceptions inside of `checkKey` does seem to cause a request exception.

apps/backend/src/authn/apikey.strategy.ts
```diff
@@ -1,23 +1,23 @@
...
+  constructor(private readonly authnService: AuthnService) {
     super(
       {header: 'Authorization', prefix: 'Api-Key '},
       false,
       async (
         apikey: string,
         done: (arg1: null, arg2: Promise<User | null> | boolean) => unknown
       ) => {
-        return done(null, this.apiKeyService.checkKey(apikey));
+        return done(null, this.authnService.validateApiKey(apikey));
```
Bialogs, line 20: Still passing null to the exception parameter of `done`. What happens when you pass an exception based on the return value of `authnService.validateApiKey`?

apps/backend/src/authn/authn.service.ts
```diff
@@ -28,6 +34,27 @@ export class AuthnService {
     }
   }
 
+  async validateApiKey(apikey: string): Promise<User | null> {
+    try {
+      const jwtPayload = jwt.verify(
+        apikey,
+        this.configService.get('JWT_SECRET') || 'disabled'
```
Bialogs, line 41: This is passing the string literal to the key verification step so everything will fail if its disabled. But if you can check if the key is disabled, why is this function getting called & polluting logic within it ? You can raise a better error if you don't let the program get this far and raise something before this step.

Bialogs (changes requested):

apps/backend/src/users/users.service.ts
Bialogs: @rbclark

apps/backend/src/authn/apikey.strategy.ts
camdenmoors: It continues the request without throwing an an error even if you provide an error to `done`.

apps/backend/src/apikeys/apikey.service.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  async count(): Promise<number> {
+    return this.apiKeyModel.count();
+  }
+
+  async create(
+    user: User,
+    createApiKeyDto: CreateApiKeyDto
+  ): Promise<{id: string; name: string; apiKey: string}> {
+    const JWTSecret = this.configService.get('JWT_SECRET');
```
Bialogs, line 34: instead of using this JWT_SECRET i think we just need to have something in the config service or app module that is required (app will fail if not set) for API_KEYS_SECRET or whatever

apps/backend/src/apikeys/apikey.service.ts
```diff
@@ -0,0 +1,87 @@
...
+        userId: user.id,
+        name: createApiKeyDto.name
+      });
+      await newApiKey.save();
+      const newJWT = jwt.sign(
+        {keyId: newApiKey.id, createdAt: new Date()},
+        JWTSecret
+      );
+      // Since BCrypt has a 72 byte limit only hash the JWT signature
+      const JWTSignature = newJWT.split('.')[2];
```
Bialogs, line 48: Can you explain this

Bialogs (dismissed):

apps/backend/src/interceptors/api-keys-allowed-interceptor.ts
```diff
@@ -0,0 +1,27 @@
...
+import {ConfigService} from '../config/config.service';
+
+@Injectable()
+export class ApiKeysAllowedInterceptor implements NestInterceptor {
+  private readonly configService: ConfigService;
+  constructor(configService: ConfigService) {
+    this.configService = configService;
+  }
+
+  intercept(_context: ExecutionContext, next: CallHandler): Observable<void> {
```
Bialogs, line 18: With my other suggestions this would effectively change to a server error if the new API secret env is not set.

Bialogs: I actually think this file may be removed

rbclark (changes requested):

apps/backend/src/apikeys/apikey.service.ts
rbclark: Instead of `API_KEY_SECRET`, I'd argue that we just use `JWT_SECRET` for API keys and just generate `JWT_SECRET` as a random string on startup.

apps/backend/src/apikeys/apikey.service.ts
camdenmoors: BCrypt is limited in how many characters it can hash. If the entire JWT token is hashed with BCrypt, only the first 71 characters would be hashed.

The signature is unique to both the payload being signed and the signing key, so hashing the signature should be unique to each JWT token.

apps/backend/src/apikeys/apikey.service.ts
camdenmoors: I think that we should just stick with requiring `JWT_SECRET`. If we generated a new random string on startup, it would cause all existing API keys to break. (The app won't start without a JWT_SECRET anyways)

image

Bialogs: What existing API keys? This feature hasn't been released yet

rbclark: The issue here is not whether the app starts or not, it is just that in general reusing secrets for multiple purposes is generally bad. This is made even worse by JWT's since the only way to expire a users session is to change the secret for all users, meaning expiring a single session would currently also expire all API keys. Having a different secret for API keys would be better. With that said, we don't really have a method of notifying users that they need to rerun `setup-docker-secrets.sh` to generate a new key, so my initial suggestion was to repurpose `JWT_SECRET` to be used for API keys and generate a new random secret for users on startup. With that said, maybe we just go for it and instead add a `API_KEY_SECRET` and have the feature only enable when `API_KEY_SECRET` is set, or just throw an error on startup when `API_KEY_SECRET` isn't set just like we do for JWT secrets.

camdenmoors: I've moved it to now check for `API_KEY_SECRET` and throw a ForbiddenException if not set on API-key related actions.

apps/backend/src/authn/apikey.strategy.ts
Bialogs: I just tested this and that is not the case

apps/frontend/src/components/global/UserModal.vue
```diff
@@ -203,6 +323,12 @@ export default class UserModal extends Vue {
   passwordConfirmation = '';
   buttonLoading = false;
 
+  mounted() {
+    if(!this.admin) {
+      this.getAPIKeys()
+    }
```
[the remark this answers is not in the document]
camdenmoors: This prevents the administrator from managing other users API keys (not their own), should admins be allowed to manage other user's keys?

apps/backend/src/authn/apikey.strategy.ts
camdenmoors: Tried it again as well, it throws a 500 server error. Throwing exceptions within `validateApiKey` returns a more specific error

apps/frontend/src/components/global/UserModal.vue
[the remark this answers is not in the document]
rbclark: Is there anything else that we restrict admins from doing?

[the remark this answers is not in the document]
camdenmoors: No, there is not

apps/backend/src/authn/apikey.strategy.ts
Bialogs: That's now what I saw. It throws whatever error you pass into the first parameter. So if validateApiKey returns null, return create a new forbidden exception and pass it to the first parameter.

rbclark (dismissed):

georgedias (changes requested):

apps/backend/.env-example
```diff
@@ -7,6 +7,7 @@ DATABASE_PASSWORD=<Password to authenticate to the database (if nothing is provi
 DATABASE_NAME=<Name of the database (if nothing is provided, defaults to heimdall-server-NODE_ENV)>
 JWT_SECRET=<JSON Web Token Secret (no default, must be set)>
 JWT_EXPIRE_TIME=<JSON Web Token Length of time before signature expires (if nothing is provided, defaults to 60s)>
+API_KEY_SECRET=<API Key Token Secret (no default, API keys are disabled if this is not set)>
```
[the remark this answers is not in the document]
camdenmoors: We had settled upon the API being disabled if this secret is not set, because one could enable the API-Key but not set the secret.

georgedias (approved):

PR #1486, opened 2021-06-15
Add Control Run Time to Control Table

### LINKED ISSUE #55: Add control runtime information to the datatable to identify long running controls

camdenmoors:
Closes #55, adds a control run time column to the ControlTable to help identify long running controls.

image


apps/frontend/src/components/cards/controltable/ResponsiveRowLarge.vue
```diff
@@ -31,6 +31,12 @@
         <slot name="tags" />
       </v-card>
     </v-col>
+
+    <v-col class="pa-0" :cols="0.5">
+      <v-card tile flat class="fill-height">
+        <slot name="runTime" />
+      </v-card>
+    </v-col>
```
rbclark, line 39: I'm extremely wary of using partial columns. Last time we merged a PR with partial columns we had to revert it since it breaks word wrap. I don't believe it is a tested/supported feature of Vuetify, it just happens to work sometimes until it doesn't. Last time the bug was with the fact that word wrap doesn't work on columns with half width, and the Result Set row was busting out of its column and breaking the rest of the table. What is it possible to do with this PR and full size columns?

rbclark (changes requested):

rbclark:
Screen Shot 2021-06-29 at 1 26 44 PM

The alignment for the new row seems a bit off, also will there be any problems with this row being added and a 'Viewed' row being added https://github.com/mitre/heimdall2/pull/1528

aaronlippold:
Are we at a point where we may want to add an 'expanded view' to the
results table that hides the top two sections and minimizes the sidebar?
--------
Aaron Lippold

***@***.***

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

aaronlippold:
https://github.com/mirari/vue-fullscreen perhaps or even just a new view
that gives us the space we need?
--------
Aaron Lippold

***@***.***

260-255-4779

twitter/aim/yahoo,etc.
'aaronlippold'

rbclark:
I'm not sure how fullscreen would help since the issue is vertical space not horizontal, and fullscreen really only eliminates the tab and navigation bar.

We likely need to have a discussion about what fields are actually important to have, or revisit how we are slicing up space between columns.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -82,21 +82,31 @@
...
 import _ from 'lodash';
+import {context} from 'inspecjs';
+
+export function getControlRunTime(control: context.ContextualizedControl): number {
+  let time = 0;
+  control.hdf.segments?.forEach((segment) => {
+    time += segment.run_time || 0
+  })
+  return time
+}
```
rbclark, line 109: Since `segments` are an array can't you just do
```
control.hdf.segments?.reduce((total, segment) => segment.run_time + total, 0)

rbclark (changes requested):
See comment regarding alignment

rbclark (approved):

PR #1487, opened 2021-06-15
Force v-group-slider to recalculate position on switching tabs

### LINKED ISSUE #115: Sample selected bar in wrong position
When you click on Samples it shows the space above it as selected
Screen Shot 2020-06-25 at 9 33 08 AM

camdenmoors:
Closes #115. Forces v-group-slider to recalculate its position on switching tabs.

### DIFF
--- a/apps/frontend/src/components/global/UploadNexus.vue
+++ b/apps/frontend/src/components/global/UploadNexus.vue
@@ -11,7 +11,6 @@
       </v-banner>
       <v-tabs
         :vertical="$vuetify.breakpoint.mdAndUp"
-        active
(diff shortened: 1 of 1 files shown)


camdenmoors:
> Before merging this it would be nice to create an upstream ticket with Vuetify with a simple reproduction. The crux of the issue here seems to be a collection of `v-tabs` with a `<v-spacer />` that is dynamically changing size since the Window is also resizing.

I've opened an issue [here](https://github.com/vuetifyjs/vuetify/issues/13806)

rbclark (approved):

PR #1490, opened 2021-06-16
Lock to old version of Dart sass

rbclark:
The dart sass developers have decided that hiding deprecation warnings, even if they are out of our control, should not be allowed: https://github.com/sass/libsass/issues/2822#issuecomment-482914373

They have recommended locking to a version that does not have the warnings so this PR does just that.


camdenmoors (approved):

sesheikholeslam (commented):
Cool, no more getting spammed with warnings on boot up.

Issue #1027, opened 2021-03-03
Allow owners to change users group role


rbclark:
**Is your feature request related to a problem? Please describe.**
It currently isn't possible for an owner to edit a users role.

Once #1026 is resolved, the ability to click on that role should be added, and a dropdown should appear allowing the owner to change this setting.

samuelsunvold:
8d125471f37768267eb8b1b1cfa4bb212f4d9663

Amndeep7:
The associated PR needs to be reviewed and then merged

PR #1494, opened 2021-06-16
Allow owners to change users group role

samuelsunvold:
Closes [#1027](https://github.com/mitre/heimdall2/issues/1027). Allow owners to change users group role.


camdenmoors (dismissed):

rbclark (changes requested):
When adding a new user, by default the role is shown as blank even though it defaults to member after save

https://user-images.githubusercontent.com/2308869/124807488-ce243800-df2b-11eb-9ba9-8f6eb341e690.mov

It is currently possible to demote all members of a group to members, making the group unmanageable by anyone.

https://user-images.githubusercontent.com/2308869/124807585-ed22ca00-df2b-11eb-99bd-5fa419d1baf0.mov

apps/backend/src/groups/groups.controller.ts
```diff
@@ -165,6 +166,19 @@ export class GroupsController {
     );
   }
 
+  @Put(':id/updateUserGroupRole')
+  async updateGroupUserRole(
```
Bialogs, line 170: The route and method should match

libs/interfaces/user/slim-user.interface.ts
```diff
@@ -2,7 +2,7 @@ export interface ISlimUser {
   readonly id: string;
   readonly email: string;
   readonly title?: string;
-  readonly groupRole?: string;
+  groupRole?: string;
```
Bialogs, line 5: Why is a return-only interface having readonly removed?

apps/backend/src/groups/groups.service.ts
```diff
@@ -49,6 +51,18 @@ export class GroupsService {
     });
   }
 
+  async updateUserGroupRole(
+    group: Group,
+    updateGroupUser: AddUserToGroupDto
```
Bialogs, line 56: Seems like there should be an `updateGroupUser` dto instead of repurposing addusertogroup

apps/backend/src/groups/groups.service.ts
```diff
@@ -49,6 +51,18 @@ export class GroupsService {
...
+  ): Promise<GroupUser | undefined> {
+    return group.users
+      .find((userToUpdate) => updateGroupUser.userId === userToUpdate.id)
```
Bialogs, line 59: We have a user id from the DTO and we have the group id. It is more efficient to query for the GroupUser directly using SQL and not iterate a potentially large list of users in a group.

apps/backend/src/groups/groups.service.ts
```diff
@@ -49,6 +51,18 @@ export class GroupsService {
...
+      ?.GroupUser.update({
+        role: updateGroupUser.groupRole,
+        updatedAt: new Date()
```
Bialogs, line 62: Are you sure `updatedAt` is not set automatically?

Bialogs (changes requested):
Missing backend tests for this new function.

aaronlippold:
@camdenmoors @Amndeep7 what is blocking us here? Can we rebase this and get it off the board?

Amndeep7:
We still haven't done the code review - work keeps on coming up.

Rlin232:
Note to self: Add functionality for assigning owners upon group creation.

Rlin232:
https://user-images.githubusercontent.com/38306648/224396196-cb74b175-e3b8-4b43-80e0-8f0797f50836.mov

Currently, updating the same user's role twice requires a hard refresh for the second update to register. Note that if we try to update a different user's role (or the same user's role, but in another group) this does not require a refresh. This is likely a reactivity issue that we will look into.

Amndeep7:
Resolves https://github.com/mitre/heimdall2/issues/4038

apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -120,35 +125,78 @@ export default class Users extends Vue {
...
+          ...user,
+          groupRole: 'member'
+        });
       }
     });
     this.usersToAdd = [];
   }
 
+  @Emit()
+  onUpdateGroupUserRole(newValue: string) {
```
Amndeep7, line 138: use a more accurate parameter name

apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -120,35 +125,78 @@ export default class Users extends Vue {
...
+    const userToUpdate = this.currentUsers.indexOf(editedUser);
+    const updatedGroupUser: ISlimUser = {
+      ...editedUser,
+      groupRole: newValue
+    };
+    this.currentUsers[userToUpdate] = updatedGroupUser;
+    return saveable;
+  }
+
+  numberOfOwners(): number {
```
Amndeep7, line 156: can have a better implementation using functional programming

Amndeep7 (changes requested):

Amndeep7 (approved):

Amndeep7 (changes requested):
see DMs - the notification for the group being successfully created is fine; however, then we have one about axios failing for some reason.

PR #1519, opened 2021-06-22
Add Logging to Backend

### LINKED ISSUE #965: Add Enhanced Logging Support
I believe it would be beneficial for admins to be able to keep track of users' actions, for example:

```[DateTime - User IP] <User> created <EvaluationName>
[DateTime - User IP] <User> added groups <Group1, Group2> to evaluation <EvaluationName>
[DateTime - User IP] <User (Admin)> deleted user <User>
[DateTime - User IP] <IP> attempted logon as <User>
[DateTime - User IP] <IP> attempted LDAP logon as <User>
```

This should be able to integrate with existing logging providers such as:

[Azure Monitor](https://azure.microsoft.com/en-us/services/monitor/)
[AWS Centralized Logging](https://aws.amazon.com/solutions/implementations/centralized-logging/)
[Google Cloud Logging](https://cloud.google.com/logging/)
[Logflare](https://logflare.app/)

Or just be as simple as logging to `/var/log/heimdall/server.log`.

camdenmoors:
Closes #965, adds user action logging.


apps/backend/src/logging/logging.service.ts
```diff
@@ -0,0 +1,170 @@
...
+    return `Group<ID: ${group.id} | Name: ${group.name} | Users[${group.users
+      .map((user) => this.userToString(user))
+      .join(', ')}]>`;
+  }
+
+  // Actions involving authentication
+  logAuthenticationAction(request: Request, action: string, user?: User): void {
+    this.logger.info({
+      ip: request.ip,
+      message: `${action}: ${this.userToString(user)}`
```
Bialogs, line 131:
```suggestion
      message: `${request.originalUrl} ${request.method} ${action}: ${this.userToString(user)}`
```

Will produce `[Jun-24-2021 14:01:16 ::ffff:127.0.0.1] /authn/login POST Succeded Login As: User<ID: 1>`

apps/backend/src/logging/logging.service.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  userToString(user?: User | UserDto | SlimUserDto): string {
+    if (user) {
+      if ('organization' in user && user.organization && user.title) {
+        return `User<ID: ${user.id} | Name: ${user.lastName}, ${user?.firstName} | Organization: ${user.organization} (${user.title})>`;
+      } else if ('organization' in user && user.organization) {
+        return `User<ID: ${user.id} | Name: ${user.lastName}, ${user?.firstName} | Organization: ${user.organization}>`;
+      }
+      return `User<ID: ${user.id} | Name: ${user.lastName}, ${user?.firstName}>`;
```
Bialogs, line 157:
```suggestion
      return `User<ID: ${user.id}>`;
```

Bialogs (changes requested):
I suggested a change specifically to how authentication logging is handled in `logging.service.ts` but I think that in general we need to be logging the route, method, parameters and remove extra potential PII such as names (`request.originalUrl`, `request.method`, etc)

This needs to be done for all logging types so I'm thinking we have a a base logger that pulls this request and user information. In the event that the parameter contains sensitive inputs (such as a password) there needs to be a filter with a developer-definable list of keys to "remove" (by substituting `[FILTERED]`) when writing to the log.

I think that having a logging function for each different controller is way too specific. We need a generic logging function that takes the request, filters, an action string, and a generic model so that we can get the ID, or an ID and the model's type.

apps/backend/src/interceptors/logging.interceptor.ts
```diff
@@ -0,0 +1,52 @@
...
+import {UserDto} from '../users/dto/user.dto';
+import {User} from '../users/user.model';
+
+@Injectable()
+export class LoggingInterceptor implements NestInterceptor {
+  public logger = winston.createLogger({
+    transports: [new winston.transports.Console()],
+    format: winston.format.combine(
+      winston.format.timestamp({
+        format: 'MMM-DD-YYYY HH:mm:ss'
```
Bialogs, line 21: please include timezone in format.

Bialogs (changes requested):
I don't see the parameter logging or parameter filtering in place.

Bialogs (approved):

PR #1521, opened 2021-06-22
Migrates InSpecJS into Heimdall2

### LINKED ISSUE #1496: Move InSpec JS into a Module
It is currently difficult to modify InSpecJS to make changes and fix bugs encountered when developing Heimdall.

Moving InSpecJS into a module within Heimdall using yarn workspaces (similar to how the test suite runs).

InSpecJS should still be able to publish to NPM as a separate package.

sesheikholeslam:
Closes #1496. 
Closes #1685.
Moves InSpecJS into Heimdall2 local root library for ease of access as a git subtree. InSpecJS is still able to be published to NPM as an independent package. 

Moves https://github.com/mitre/inspecjs into `libs/inspecjs`

The InSpecJS GitHub repository is now no longer needed.
Documentation needs to be updated and references on the saf.mitre.org website need to be updated as well.


libs/inspecJS/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "inspecjs",
+  "version": "2.4.6",
```
camdenmoors, line 3: Is jumping from version 0.12.4 to 2.4.6 going to cause issues? If we are going to match InspecJS with Heimdall can this be updated to the current release? (2.4.8)

sesheikholeslam: I don't believe this will cause any issues, so I also think it should be fine to update it to the current release.

rbclark: Lerna automatically modifies these on release as well.

rbclark (changes requested):

.github/workflows/push-to-npm.yml
```diff
@@ -35,3 +35,25 @@ jobs:
...
+        run: yarn inspecjs install --frozen-lockfile
+
+      - name: Build InSpecJS package files
+        run: yarn inspecjs build
+
+      - name: Build tarball for publishing
+        run: yarn inspecjs pack
+
+      - name: Publish InSpecJS to NPM
+        run: yarn publish --access public apps/frontend/inspecjs*.tgz
```
rbclark, line 56: This path is incorrect, the inspecjs code is located in `libs/inspecjs`.

libs/inspecJS/_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-cayman
```
rbclark, line 1: This file should no longer be necessary.

libs/inspecJS/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/mitre/inspecjs.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/mitre/inspecjs/issues"
```
rbclark, line 35: This is no longer correct.

libs/inspecJS/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "homepage": "https://github.com/mitre/inspecjs#readme"
```
rbclark, line 37: This is no longer correct.

rbclark (changes requested):
Can you change the path from `libs/inspecJS` to `libs/inspecjs`?

libs/inspecjs/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "@typescript-eslint/parser": "^2.34.0",
+    "eslint": "^6.8.0",
+    "eslint-config-prettier": "^6.15.0",
+    "eslint-plugin-prettier": "^3.3.1",
+    "prettier": "^1.19.1",
+    "typedoc": "^0.20.35",
+    "typescript": "^3.9.9"
+  },
+  "scripts": {
+    "build": "rm -r ./dist; tsc",
```
rbclark, line 22: Instead of using `rm` here, this should instead just use a prebuild step just like heimdall-lite does: https://github.com/mitre/heimdall2/blob/master/apps/frontend/package.json#L16

libs/inspecjs/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "docs": "typedoc --out docs ./src",
+    "gen-types": "bash scripts/update_schema.sh",
+    "lint": "eslint --ext .ts src; exit 0"
```
rbclark, line 25: Not sure what the point of eslint is here if we are just going to make it always pass. Can this be updated to look how our other linters look?

.github/workflows/inspecjs-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+     - uses: actions/checkout@v2
+
+     - name: Cache node modules
+       uses: actions/cache@v2
+       env:
+         cache-name: cache-node-modules
+       with:
+         # npm cache files are stored in `~/.npm` on Linux/macOS
+         path: ~/.npm
+         key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
rbclark, line 30: We're using yarn which generates a `yarn.lock` not a `package-lock.json`, this is wrong in all of the other workflows as well unfortunately.

package.json
```diff
@@ -11,16 +11,19 @@
...
     "start": "yarn backend start",
-    "start:dev": "dotenv -e apps/backend/.env -- lerna exec yarn run start:dev --ignore @heimdall/interfaces --ignore @heimdall/cypress-tests",
+    "start:dev": "dotenv -e apps/backend/.env -- lerna exec yarn run start:dev --ignore @heimdall/interfaces --ignore @heimdall/cypress-tests --ignore inspecjs",
     "test:ui": "cypress run",
     "test:ui:open": "cypress open"
   },
   "dependencies": {
     "@typescript-eslint/eslint-plugin": "^4.8.0",
     "@typescript-eslint/parser": "^4.8.0",
+    "cypress": "7.7.0",
```
rbclark, line 26: Cypress is a dependency of `test/package.json`, why is it needed here?

.github/workflows/inspecjs-tests.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    services:
+      postgres:
+        image: postgres:latest
+        env:
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
```
rbclark, line 18: InspecJS doesn't need postgres

rbclark (approved):

PR #1527, opened 2021-06-23
Revert "Bump sass from 1.32.13 to 1.35.1"

rbclark:
Reverts mitre/heimdall2#1502


camdenmoors:
The deploy preview doesn't seem to load correctly with:
image
And on firefox:
image

PR #1548, opened 2021-06-29
Add Sorting Compare view

### LINKED ISSUE #1416: Allow sorting of Comparisons by fields other than time
There are times where time is not the relevant field to sort by for comparison.

A dropdown to select which field to use for sorting (including sorting by passthrough data fields if applicable). This dropdown can be present in the top right corner of comparison view.

Comparison of results over time for situations where comparison is not as straightforward as just time-based comparison.

camdenmoors:
Closes #1416 

Adds sort by:
 - Run Time
 - Execution Length
 - Control Count
 - Completed Control Count
 - Completed Control %
 - Passed Control Count
 - Compliance (Passed Control %)
 - Passthrough fields (String, Number, Boolean)


rbclark:
Blocked until #1382 is merged

rbclark:
@camdenmoors now that advanced search is merged can you point this PR at master and fix any conflicts?

apps/frontend/src/types/search-parser.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'search-parser';
```
rbclark, line 1: This is not relevant anymore correct?

apps/frontend/src/views/Compare.vue
```diff
@@ -260,6 +267,17 @@ export default class Compare extends Vue {
     }
   ];
 
+  compareItems = [
+    'Run Time',
+    'Execution Length',
```
rbclark, line 272: This is the only instance of `Execution Length` in this PR. It seems to be called `Execution Time` below.

Also this should probably be called `Run Time`

apps/frontend/src/views/Compare.vue
```diff
@@ -260,6 +267,17 @@ export default class Compare extends Vue {
     }
   ];
 
+  compareItems = [
+    'Run Time',
```
rbclark, line 271: `Run Time` is not the ideal word to use here, `runtime` normally means how long something takes to run, this is comparing the actual start time. This should probably be called `Scan Start Time` or maybe just `Start Time`

apps/frontend/src/views/Compare.vue
```diff
@@ -260,6 +267,17 @@ export default class Compare extends Vue {
...
+    'Execution Length',
+    'Control Count',
```
rbclark, line 273: This is the `Total Number of Controls` correct?

apps/frontend/src/views/Compare.vue
```diff
@@ -260,6 +267,17 @@ export default class Compare extends Vue {
...
+    'Completed Control Count',
```
rbclark, line 274: What is the difference between `Completed` and `Passed` controls?

apps/frontend/src/utilities/delta_util.ts
```diff
@@ -163,6 +165,68 @@ export function get_eval_start_time(
...
+}
+
+export function getControlCount(
+  resultsSet: SourcedContextualizedEvaluation
+): number {
+  let count = 0;
+  resultsSet.contains.forEach((value) => {
+    count += value.contains.length;
+  });
+  return count;
```
rbclark, line 189: This can probably be replaced with the `Array.reduce` method.

apps/frontend/src/utilities/delta_util.ts
```diff
@@ -163,6 +165,68 @@ export function get_eval_start_time(
...
+  resultsSet: SourcedContextualizedEvaluation
+): number {
+  let time = 0;
+  resultsSet.contains.forEach((value) => {
+    value.contains.forEach((control) => {
+      control.hdf.segments?.forEach((segment) => {
+        time += segment.run_time || 0;
+      });
+    });
+  });
```
rbclark, line 178: There's already a function in the codebase to do the last layer of this check, (see `getControlRunTime` in ControlRowHeader). If you move that function to a more accessible location then this becomes 2 reduce statements wrapping getControlRunTime.

apps/frontend/src/utilities/delta_util.ts
```diff
@@ -163,6 +165,68 @@ export function get_eval_start_time(
...
+  const total =
+    passed +
+    StatusCountModule.countOf(filter, 'Failed') +
+    StatusCountModule.countOf(filter, 'Profile Error') +
+    StatusCountModule.countOf(filter, 'Not Reviewed');
+  if (total === 0) {
+    return 0;
+  } else {
+    return Math.round((100.0 * passed) / total);
+  }
```
rbclark, line 227: This function is identical to the one used to calculate compliance elsewhere. Can you just add a `StatusCountModule.calculateCompliance` method and then reuse it everywhere?

rbclark (dismissed):
Having the ability to reverse the sort direction on this may also be useful.

apps/frontend/src/views/Compare.vue
camdenmoors: This is the same as `Control Count`, will remove.

PR #1554, opened 2021-06-30
Ignore dist and sample in search

Bialogs:


rbclark:
Considering we have a workspace, can't you just make this a workspace setting? Also why do you need it in the backend as well?

rbclark (approved):

PR #1560, opened 2021-07-01
Add Support for Custom Gitlab Base URL

### LINKED ISSUE #1553: Allow for authorization to gitlab instances other than gitlab.org
The `passport-gitlab2` library currently in use supports custom `baseURL's` however that option is not exposed to users of Heimdall2 currently. 

http://www.passportjs.org/packages/passport-gitlab2/

This would be a useful feature since many organizations have an internal gitlab server.

camdenmoors:
Closes #1553, adds `GITLAB_BASEURL` environment variable to specify a custom Gitlab instance.

### DIFF
--- a/apps/backend/src/authn/gitlab.strategy.ts
+++ b/apps/backend/src/authn/gitlab.strategy.ts
@@ -24,6 +24,7 @@ export class GitlabStrategy extends PassportStrategy(Strategy, 'gitlab') {
     super({
       clientID: configService.get('GITLAB_CLIENTID') || 'disabled',
       clientSecret: configService.get('GITLAB_SECRET') || 'disabled',
+      baseURL: configService.get('GITLAB_BASEURL'),
       callbackURL:
         `${configService.get('EXTERNAL_URL')}/authn/gitlab/callback` ||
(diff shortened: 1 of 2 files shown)


apps/backend/src/authn/gitlab.strategy.ts
```diff
@@ -24,6 +24,7 @@ export class GitlabStrategy extends PassportStrategy(Strategy, 'gitlab') {
     super({
       clientID: configService.get('GITLAB_CLIENTID') || 'disabled',
       clientSecret: configService.get('GITLAB_SECRET') || 'disabled',
+      baseURL: configService.get('GITLAB_BASEURL'),
```
rbclark, line 27: Does this still handle the default case properly?

rbclark (changes requested):
Can the documentation be updated as well?

apps/backend/src/authn/gitlab.strategy.ts
camdenmoors: It does as the default is set within passport-gitlab2

rbclark (approved):

PR #1561, opened 2021-07-01
Use First LDAP Email Available if Multiple

camdenmoors:
This fixes the server failing to create an account when multiple email values are present on the LDAP account.

### DIFF
--- a/apps/backend/src/authn/ldap.strategy.ts
+++ b/apps/backend/src/authn/ldap.strategy.ts
@@ -37,12 +37,22 @@ export class LDAPStrategy extends PassportStrategy(Strategy, 'ldap') {
           user,
           configService.get('LDAP_MAILFIELD') || 'mail'
         );
-        req.user = this.authnService.validateOrCreateUser(
-          email,
-          firstName,
-          lastName,
-          'ldap'
-        );
+        // If user has multiple emails
+        if (typeof email === 'object') {
+          req.user = this.authnService.validateOrCreateUser(
+            email[0],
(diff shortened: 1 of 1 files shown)


apps/backend/src/authn/ldap.strategy.ts
```diff
@@ -37,12 +37,22 @@ export class LDAPStrategy extends PassportStrategy(Strategy, 'ldap') {
...
-        req.user = this.authnService.validateOrCreateUser(
-          email,
-          firstName,
-          lastName,
-          'ldap'
-        );
+        // If user has multiple emails
+        if (typeof email === 'object') {
+          req.user = this.authnService.validateOrCreateUser(
+            email[0],
```
rbclark, line 43: Shouldn't this just be `typeof email === object ? email[0] : email` or something similar instead of duplicating the whole block?

rbclark (changes requested):

apps/backend/src/authn/ldap.strategy.ts
```diff
@@ -38,7 +38,7 @@ export class LDAPStrategy extends PassportStrategy(Strategy, 'ldap') {
           configService.get('LDAP_MAILFIELD') || 'mail'
         );
         req.user = this.authnService.validateOrCreateUser(
-          email,
```
rbclark, line 41: I think `Array.isArray()` may be more relevant here

rbclark (changes requested):

PR #1580, opened 2021-07-06
Move Treemap to use Apex Charts

### LINKED ISSUE #210: Unify on a single chart library
This application currently pulls in d3 and ApexCharts. d3 is used for the Treemap and ApexCharts is used for the Status Counts, Severity Counts, and Compliance level. Using 2 chart libraries brings some additional overhead and is somewhat unnecessary since d3 is able to do everything that ApexCharts can. The options are as follows:

* Switch fully over to ApexCharts (this may not be possible since it doesn't support Treemaps)
* Switch fully over to d3 (this may be difficult since d3 does not provide the simplest interface, there is a Vue d3 charts package available, https://github.com/Saigesp/vue-d3-charts)
* Switch everything over to a new package such as https://www.chartjs.org/

camdenmoors:
Closes #210, removes the d3 library and replaces the d3 treemap with an Apex Charts treemap.

### DIFF
--- a/apps/frontend/tests/unit/Compare.spec.ts
+++ b/apps/frontend/tests/unit/Compare.spec.ts
@@ -7,7 +7,6 @@ import {shallowMount, Wrapper} from '@vue/test-utils';
 import 'jest';
 import Vue from 'vue';
 import Vuetify from 'vuetify';
-import {SearchModule} from '../../src/store/search';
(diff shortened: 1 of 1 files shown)


camdenmoors:
The new goal is to convert the other charts to D3 due to the lack of customization for treemaps using Apex Charts.

PR #1656, opened 2021-07-14
HDF Converters

rbclark:
This adds all of the non-API based mappers from `heimdall_tools` to the heimdall2 monorepo and rewrites them in Typescript.


.vscode/launch.json
```diff
@@ -18,6 +18,21 @@
...
+      },
+      "type": "node",
+      "request": "launch",
+      "runtimeArgs": [
+        "-r",
+        "ts-node/register"
+      ],
+      "args": [
+        "${workspaceRoot}/libs/hdf_converters/src/test.ts"
+      ]
```
rbclark, line 35: This should probably be removed since I don't think the `test.ts` file is the long term way we were performing tests.

libs/hdf_converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,345 @@
...
+          rule = _.set(rule, 'impact', 0);
+          rule = _.set(
+            rule,
+            'results',
+            _.get(rule, 'results').concat([
+              {
+                run_time: 0,
+                code_desc: NOT_APPLICABLE_MSG,
+                skip_message: NOT_APPLICABLE_MSG,
+                start_time: Date.now(), //GOTTA FIGURE THIS OUT TOO
```
rbclark, line 108: Is this comment still relevant?

libs/hdf_converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,345 @@
...
+          rule = _.set(rule, 'impact', 0);
+          rule = _.set(
+            rule,
+            'results',
+            _.get(rule, 'results').concat([
+              {
+                run_time: 0,
+                code_desc: INSUFFICIENT_DATA_MSG,
+                skip_message: INSUFFICIENT_DATA_MSG,
+                start_time: Date.now(), //GOTTA FIGURE THIS OUT TOO
```
rbclark, line 125: Is this comment still relevant?

libs/hdf_converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,345 @@
...
+        result,
+        'nist',
+        _.get(result, 'nist').concat(defaultMatch)
+      );
+    }
+
+    //Deal with custom mapping later
+    // custom_mapping_match = @custom_mapping[source_identifier]
+
+    // result['nist'] += custom_mapping_match['NIST-ID'].split('|').map { | name | "#{name} (user provided)"} unless custom_mapping_match.nil ?
```
rbclark, line 243: Is this comment still relevant? If the code is not necessary then it should be removed.

libs/hdf_converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,345 @@
...
+  public async toHdf(): Promise<ExecJSON.Execution> {
+    const hdf: ExecJSON.Execution = {
+      platform: {
+        name: 'Heimdall Tools',
+        release: HeimdallToolsVersion,
+        target_id: ''
+      },
+      version: HeimdallToolsVersion,
+      statistics: {
+        //aws_config_sdk_version: ConfigService., // How do i get the sdk version?
```
rbclark, line 316: Is this comment still relevant?

libs/hdf_converters/src/base-converter.ts
```diff
@@ -0,0 +1,265 @@
+import {createHash} from 'crypto';
+import * as htmlparser from 'htmlparser2';
+import {ExecJSON} from 'inspecjs';
+import _ from 'lodash';
+
+export interface ILookupPath {
+  path?: string;
+  // Parameter will be the return type of handlePath, which can either be a string or a number
```
rbclark, line 8: Is this comment still accurate?

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+import {NiktoMapper} from '../src/nikto-mapper';
+import {SarifMapper} from '../src/sarif-mapper';
+import {ScoutsuiteMapper} from '../src/scoutsuite-mapper';
+import {XCCDFResultsMapper} from '../src/xccdf-results-mapper';
+import {ZapMapper} from '../src/zap-mapper';
+
+test('Returns proper Hdf output from Burpsuite', () => {
+  const mapper = new BurpSuiteMapper(
+    fs.readFileSync(
+      '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/burpsuite_mapper/sample_input_report/zero.webappsecurity.com.min',
```
rbclark, line 14: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+  const result: ExecJSON.Execution = mapper.toHdf();
+  fs.writeFileSync(
+    'libs/hdf_converters/outputs/burpsuite.json',
+    JSON.stringify(mapper.toHdf())
+  );
+  if (result !== undefined) {
+    expect(result).toEqual(
+      JSON.parse(
+        fs.readFileSync(
+          '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/burpsuite_mapper/zero.webappsecurity.json',
```
rbclark, line 27: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+        )
+      )
+    );
+  }
+});
+
+test('Returns proper Hdf output from JFrog', () => {
+  const mapper = new JfrogXrayMapper(
+    fs.readFileSync(
+      '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/jfrog_xray_mapper/sample_input_report/jfrog_xray_sample.json',
```
rbclark, line 38: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+      {encoding: 'utf-8'}
+    )
+  );
+  const result: ExecJSON.Execution = mapper.toHdf();
+  if (result !== undefined) {
+    expect(result).toEqual(
+      JSON.parse(
+        fs.readFileSync(
+          '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/jfrog_xray_mapper/jfrog_xray_hdf.json',
```
rbclark, line 47: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+          {encoding: 'utf-8'}
+        )
+      )
+    );
+  }
+});
+test('Returns proper Hdf output from Nikto', () => {
+  const mapper = new NiktoMapper(
+    fs.readFileSync(
+      '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/nikto_mapper/sample_input_jsons/zero.webappsecurity.json',
```
rbclark, line 57: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+      {encoding: 'utf-8'}
+    )
+  );
+  const result: ExecJSON.Execution = mapper.toHdf();
+  if (result !== undefined) {
+    expect(result).toEqual(
+      JSON.parse(
+        fs.readFileSync(
+          '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/nikto_mapper/zero.webappsecurity.json',
```
rbclark, line 66: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+          {encoding: 'utf-8'}
+        )
+      )
+    );
+  }
+});
+test('Returns proper Hdf output from Sarif', () => {
+  const mapper = new SarifMapper(
+    fs.readFileSync(
+      '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/sarif_mapper/sample_input_jsons/sarif_input.sarif',
```
rbclark, line 76: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+      {encoding: 'utf-8'}
+    )
+  );
+  const result: ExecJSON.Execution = mapper.toHdf();
+  if (result !== undefined) {
+    expect(result).toEqual(
+      JSON.parse(
+        fs.readFileSync(
+          '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/sarif_mapper/sarif_output.json',
```
rbclark, line 85: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+          {encoding: 'utf-8'}
+        )
+      )
+    );
+  }
+});
+test('Returns proper Hdf output from Scoutsuite', () => {
+  const mapper = new ScoutsuiteMapper(
+    fs.readFileSync(
+      '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/scoutsuite_mapper/sample_input_jsons/scoutsuite_sample.js',
```
rbclark, line 95: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/test/test.ts
```diff
@@ -0,0 +1,149 @@
...
+      {encoding: 'utf-8'}
+    )
+  );
+  const result: ExecJSON.Execution = mapper.toHdf();
+  if (result !== undefined) {
+    expect(result).toEqual(
+      JSON.parse(
+        fs.readFileSync(
+          '/Users/rlin/Desktop/Repositories/heimdall_tools/sample_jsons/scoutsuite_mapper/scoutsuite_hdf.json',
```
rbclark, line 104: You need to use relative paths here, absolute paths will only work on your system.

libs/hdf_converters/src/aws-config-mapper.ts
Rlin232: Yes, I have to do a bit more reading into the Date module to find out how to convert it into something readable rather than UTC

libs/hdf_converters/src/aws-config-mapper.ts
Rlin232: Yes, I need to ask how the user would pass in a custom mapping in the first place. In addition, I'm unable to test the mapper to begin with, so I'm just sort of waiting on an access key at this point in time.

libs/hdf_converters/src/aws-config-mapper.ts
Rlin232: Yes, since the 
ExecJSON type does not allow the extra field "aws_config_sdk_version", I need to find a workaround (will likely involve transforming/appending the field at the end after the result is generated)

Rlin232 (approved):

PR #1690, opened 2021-07-21
Automates Quicktype execution as a part of InspecJS's CI

### LINKED ISSUE #1675: Automate quicktype execution as part of CI
The `generated_parsers` package is used to generate code based on the schema which is also stored in the inspecjs repository. The whole process of generating the code (from schema export from inspec to generating the generated_parsers folder) should be automated so that updates to quicktype or inspec are easily integrated into the schema and generated code in this repository.

In order for this to work the quicktype package will need to be re-added as a devDependency of inspecJS.

sesheikholeslam:
Closes #1675.

Enables quicktype execution in the InspecJS github action workflow. This is done to ensure the schemas extracted from a specific version of inspec have no differences. Additionally, the automatic quicktype execution ensures no changes have been made to the generated parsers, which are the results from the quicktype conversion of the inspec schemas.


.github/workflows/inspecjs-tests.yml
```diff
@@ -12,6 +12,9 @@ jobs:
 
     steps:
      - uses: actions/checkout@v2
+     - uses: actions/setup-ruby@v1
+       with:
+         ruby-version: '2.7.4'
```
rbclark, line 17: `ruby/setup-ruby` seems to be the officially supported action, this one is deprecated.

.github/workflows/inspecjs-tests.yml
```diff
@@ -41,6 +44,12 @@ jobs:
 
      - name: Run unit tests
        run: yarn inspecjs test
+    
+     - name: Validate there are no changes between the inspec schemas and generated parsers
+       run: |
+        gem install inspec-bin -v 4.38.3
```
rbclark, line 50: I know I originally said we wanted to just do a `gem install` here, but the more I think about it I can see a benefit to just adding a `Gemfile` and `Gemfile.lock`, just so that we can get Dependabot updates automatically for testing against new inspec versions. Thoughts?

libs/inspecjs/scripts/update_schema.sh
```diff
@@ -34,18 +20,15 @@ for SCHEMA in ${SCHEMAS[@]}
...
-    SCHEMAFILE="./work/schemas/$VERSION/$SCHEMA.json"
+    SCHEMAFILE="./schemas/$SCHEMA.json"
     MODIFIED_SCHEMAFILE="./work/tmp_$SCHEMA.json"
 
     # Loosen the schema
     COMPAT="./scripts/schema_compat_patches"
     cat $SCHEMAFILE | . "$COMPAT/generic.sh" | . "$COMPAT/$SCHEMA.sh" | . "./scripts/null_compat_schema.sh" > $MODIFIED_SCHEMAFILE
-
     # Generate the parser
     npx quicktype -l ts -s schema --src $MODIFIED_SCHEMAFILE -o $OUTFILE --runtime-typecheck
```
rbclark, line 30: I'm pretty sure using `npx` here is completely bypassing the version of quicktype that is installed and downloading a new version directly. This should just be `quicktype` without the npx prefix I believe.

.github/workflows/inspecjs-tests.yml
```diff
@@ -41,6 +44,12 @@ jobs:
...
+        yarn inspecjs gen-types
+        yarn inspecjs validates-types
```
rbclark, line 52: `validates-types` is not working correctly, it is currently returning 0 even though the file are in fact modified on my local system:

```git
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   libs/inspecjs/schemas/exec-json.json
	modified:   libs/inspecjs/schemas/exec-jsonmin.json
	modified:   libs/inspecjs/schemas/profile-json.json
	modified:   libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts
	modified:   libs/inspecjs/src/generated_parsers/v_1_0/exec-jsonmin.ts
	modified:   libs/inspecjs/src/generated_parsers/v_1_0/profile-json.ts
```

```bash
➜  heimdall2 git:(automateQuicktypeInspecJS) ✗ git diff --quiet HEAD master -- libs/inspecjs/schemas libs/inspecjs/src/generated_parsers/
➜  heimdall2 git:(automateQuicktypeInspecJS) ✗ echo $?
0
```

I think the diff command needs to be

```
git diff --quiet HEAD -- libs/inspecjs/schemas libs/inspecjs/src/generated_parsers/ 
```

for debugging purposes if you remove the `--quiet` you can see what it is actually returning.

.github/workflows/inspecjs-tests.yml
rbclark: Also, I think ruby 2.7 without specifying the 4 is sufficient, that way it will always run against the latest security patch of ruby 2.7.

rbclark (changes requested):

.github/workflows/inspecjs-tests.yml
sesheikholeslam: Yeah I completely agree. I was thinking the one of the main points of running the inspecjs tests was to make sure nothing has been altered in the generated parsers of inspec schemas. So it would be nice to also know when something has changed on inspec's end upon a new version instead of just comparing against the schemas of one version of inspec.

libs/inspecjs/scripts/update_schema.sh
sesheikholeslam: Right, I was meaning to ask you would it be the wrong move to use npx to run quicktype instead of downloading the latest version?

rbclark: Yes, the issue is using `npx` will always pull the latest version off of npm meaning that anytime a new version comes out CI could potentially start randomly failing for all PRs due to the new version.

sesheikholeslam: Gotcha, ok good to know, i'll make the change

libs/inspecjs/package.json
```diff
@@ -26,6 +27,7 @@
     "build": "tsc -p ./tsconfig.build.json",
     "docs": "typedoc --out docs ./src",
     "gen-types": "bash scripts/update_schema.sh",
+    "validates-types": "git diff --quiet HEAD -- schemas/ src/generated_parsers/",
```
rbclark, line 30: Can this be renamed to `validate-types` instead of `validates-types`?

rbclark (changes requested):
Looks really good! Tested the functionality and it works great, just a few small comments.

rbclark (approved):

PR #1754, opened 2021-08-05
Hdf converters nessus hotfix

Rlin232:


libs/hdf_converters/src/nessus-mapper.ts
```diff
@@ -143,14 +143,18 @@ function getStig(item: unknown): string {
...
       return ExecJSON.ControlResultStatus.Passed;
-    } else {
+    case 'WARNING':
+      return ExecJSON.ControlResultStatus.Skipped;
+    case 'ERROR':
+      return ExecJSON.ControlResultStatus.Error;
+    case 'FAILED':
+      return ExecJSON.ControlResultStatus.Failed;
+    default:
       return ExecJSON.ControlResultStatus.Failed;
```
rbclark, line 157: If Failed is the default then there is no need to include the Failed case as well here.

Rlin232: Ok, I'll make the change

rbclark (approved):

Issue #1653, opened 2021-07-13
STIG Requirement: Log Heimdall shutdown event


dtingstrom:
V-222469 states "The application must log application shutdown events."

We need to add a log statement stating that Heimdall is shutting down.

PR #1758, opened 2021-08-06
Add Startup/Shutdown Lifecycle Logs

### LINKED ISSUE #1645: STIG Requirement: Log OS correlation information
Using node process module include the web server's process ID and current user.

camdenmoors:
Closes #1645, #1653. Adds Lifecycle hooks for startup and shutdown:

```
[Aug-06-2021 13:39:45 -04:00] Started Heimdall Enterprise Server 2.0 on M-PC (darwin 20.5.0) with PID 5106 and UID 501
[Aug-06-2021 13:39:46 -04:00] Recieved SIGTERM, starting shutdown for PID 5106
[Aug-06-2021 13:39:46 -04:00] Finished shutdown for SIGTERM for PID 5106
```


apps/backend/src/app.service.ts
```diff
@@ -0,0 +1,45 @@
...
+      winston.format.timestamp({
+        format: 'MMM-DD-YYYY HH:mm:ss Z'
+      }),
+      winston.format.printf((info) => `[${[info.timestamp]}] ${info.message}`)
+    )
+  });
+
+  onApplicationBootstrap(): void {
+    this.logger.info({
+      message: `Started Heimdall Enterprise Server 2.0 on ${os.hostname()} (${os.platform()} ${os.release()}) with PID ${
```
Bialogs, line 29: If we could remove 2.0 so we don't need to figure out how to keep this updated that would be good.

Bialogs (changes requested):

PR #1759, opened 2021-08-06
Limit JWT Expire Time to Two Days

### LINKED ISSUE #1720: STIG Requirement: Limit expiration time of JWT
V-222619 of the STIG requires that "individual accounts designated as inactive, suspended, or terminated are promptly removed" and "verify that their accounts were removed or deactivated on all systems in a timely manner (e.g., less than two days)."

We currently allow administrators to set a JWT expiration time much greater than 2 days. So we should set a limit to this value that causes JWTs to be renewed more often (disabling removed accounts) and satisfy this STIG requirement.

camdenmoors:
Closes #1720. Limits JWT expire time to two days maximum.


rbclark (changes requested):
Is there a way to check and throw an error to inform the user that they have provided an invalid value instead of just silently changing the value they provided?

rbclark (approved):

PR #1760, opened 2021-08-06
Use Secure Cookies in Production Mode

### LINKED ISSUE #1665: STIG Requirement: Set secure flag on cookies
V-222576 of the STIG states: "The application must set the secure flag on session cookies."

Inspecting the cookie metadata using Chrome developer tools show that the secure flag is not set. We should set this flag.

camdenmoors:
Closes #1665, uses secure cookies when in production mode.

### DIFF
--- a/apps/backend/src/authn/authn.controller.ts
+++ b/apps/backend/src/authn/authn.controller.ts
@@ -9,6 +9,7 @@ import {
 } from '@nestjs/common';
 import {AuthGuard} from '@nestjs/passport';
 import {Request} from 'express';
+import {ConfigService} from '../config/config.service';
 import {AuthenticationExceptionFilter} from '../filters/authentication-exception.filter';
 import {LocalAuthGuard} from '../guards/local-auth.guard';
 import {LoggingInterceptor} from '../interceptors/logging.interceptor';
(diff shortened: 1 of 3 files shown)


apps/backend/src/authn/authn.controller.ts
```diff
@@ -126,8 +130,12 @@ export class AuthnController {
...
     }
   ): Promise<void> {
-    req.res?.cookie('userID', session.userID);
-    req.res?.cookie('accessToken', session.accessToken);
+    req.res?.cookie('userID', session.userID, {
+      secure: this.configService.get('NODE_ENV')?.toLowerCase() === 'production'
+    });
+    req.res?.cookie('accessToken', session.accessToken, {
+      secure: this.configService.get('NODE_ENV')?.toLowerCase() === 'production'
+    });
```
Bialogs, line 138: It would be cleaner if the check for production was made within the configService itself, similar to 

```
  isRegistrationAllowed(): boolean {
    return this.get('REGISTRATION_DISABLED')?.toLowerCase() !== 'true';
  }
```

Bialogs (changes requested):

Bialogs (changes requested):

Bialogs (approved):

Bialogs:
@dtingstrom can you test & ensure this PR meets the requirements?

rbclark (approved):

PR #1762, opened 2021-08-06
Add Sensitivity Classifications to Evaluations

### LINKED ISSUE #1716: STIG Requirement: Display classification banner
V-222643 states "The application must have the capability to mark sensitive/classified output when required."

We should implement a display banner on the client-side that shows the appropriate classification level for data viewed on Heimdall.

camdenmoors:
Closes #1716. Adds a sensitivity banner to EvaluationInfo when files have been marked with a classification value.

<details>
<summary>Examples</summary>

image
image

</details>

### DIFF
--- a/apps/backend/src/evaluations/dto/update-evaluation.dto.ts
+++ b/apps/backend/src/evaluations/dto/update-evaluation.dto.ts
@@ -13,4 +13,8 @@ export class UpdateEvaluationDto implements IUpdateEvaluation {
   @IsOptional()
   @IsBoolean()
   readonly public: boolean | undefined;
+
+  @IsOptional()
+  @IsString()
+  readonly classification: boolean | undefined;
 }
--- a/apps/backend/migrations/20210806190424-add-classification-to-evaluations.js
+++ b/apps/backend/migrations/20210806190424-add-classification-to-evaluations.js
@@ -1,0 +1,17 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+   await queryInterface.addColumn('Evaluations', 'classification', {
+     type: Sequelize.TEXT,
+     allowNull: true
+   })
+  },
+
+  down: async (queryInterface, Sequelize) => {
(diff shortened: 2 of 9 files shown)


apps/backend/src/evaluations/dto/update-evaluation.dto.ts
```diff
@@ -13,4 +13,8 @@ export class UpdateEvaluationDto implements IUpdateEvaluation {
   @IsOptional()
   @IsBoolean()
   readonly public: boolean | undefined;
+
+  @IsOptional()
+  @IsString()
+  readonly classification: boolean | undefined;
```
rbclark, line 19: I don't think this is intended to be a boolean.

rbclark (changes requested):
I think Aaron's comment was intended for this PR, not for the issue:

> This is intended to be a banner for the entire application, not individual files. One file makes the whole system the same level. 

The original implementation described in #1716 is basically just adding another classification banner and color and have it displayed in the standard way to all users, it normally looks something like this:

https://www.patternfly.org/v3/pattern-library/communication/classification-banner/img/green.png

https://i.stack.imgur.com/qVJkE.png

https://www.drupal.org/files/project-images/screen_shot_1.png

I believe the original intent of the issue was to add the ability for the user to set the color and text of the classification banner just like the warning banner can be set now and that would resolve the issue.

Bialogs (changes requested):
* Rename the variable/config `classificationBannerText`
* Needs to also take a color like `classificationBannerColor` to set the background color of the banner.

rbclark:
@Bialogs Is it a requirement that the banner displays on the top and bottom? This one currently only displays on the bottom.

rbclark (changes requested):
This needs a top banner as well.

rbclark:
https://user-images.githubusercontent.com/2308869/129937670-919dcfb1-614e-4b86-b2b2-09eb28420eb8.mov

The footer z-index seems to be incorrect.

PR #1766, opened 2021-08-09
De-duplicate sensitiveKeys filtered parameters

### LINKED ISSUE #1723: One place for all filtered parameters
Deduplicate sensitive parameters between HTTP logging and DB logging.

camdenmoors:
Closes #1723. Exports sensitiveKeys as part of DatabaseModule

### DIFF
--- a/apps/backend/src/database/database.module.ts
+++ b/apps/backend/src/database/database.module.ts
@@ -5,7 +5,7 @@ import {ConfigModule} from '../config/config.module';
 import {ConfigService} from '../config/config.service';
 import {DatabaseService} from './database.service';
 
-const sensitiveKeys = [
+export const sensitiveKeys = [
   /cookie/i,
   /passw(or)?d/i,
   /^pw$/,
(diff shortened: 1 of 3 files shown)


Bialogs (changes requested):
Organizationally, I think it would make more sense to have the sensitiveKeys declaration be part of the config service instead of the database service. Then Database and Log can consume the config service (which they already import)

apps/backend/src/config/config.service.ts
```diff
@@ -2,6 +2,17 @@ import {SequelizeOptions} from 'sequelize-typescript';
...
+  /cookie/i,
+  /passw(or)?d/i,
+  /^pw$/,
+  /^pass$/i,
+  /secret/i,
+  /token/i,
+  /api[-._]?key/i,
+  /data/i
+];
+
```
Bialogs, line 15: Can this be a function as part of the module instead of exported separately.

Bialogs (changes requested):

rbclark:
Once the existing comment on this is addressed it should be fine to merge.

apps/backend/src/config/config.service.ts
camdenmoors: I have moved it into ConfigService assuming that's what you meant here.

georgedias (approved):

PR #1783, opened 2021-08-11
Allows Github oAuth Users Without a Name to Login

### LINKED ISSUE #1777: [BUG] Github oauth name field is null instead of string
**Describe the bug**
Heimdall-demo fails to log in using the github oauth process due to attempting to do a 'split' on a null.  A null occurs if the Github profile does not have a value set in the name field.  

**To Reproduce**
1. Make sure your github account does not have a name set in the profile.
2. Attempt to log in via oauth on heimdall-demo.mitre.org.
3. See the error where it says the reason is "Cannot read property 'split' of null".

**Expected behavior**
You should still be able to log in.

**Screenshots**
image

**Additional context**
It's doing the split when trying to do a separate first name / last name sorta display in the user profile.  IMO that's not the move since some folks can have just the one name or multiple last names or not want to share their name in Github.  We should just pass through whatever we get from Github directly.  W/r to the little logo thing in the corner, it doesn't need to be the user's initials - it can be some sorta static or even dynamic symbol/graphic.

camdenmoors:
Closes #1777, allows Github oAuth users without a name to login.


apps/backend/src/authn/github.strategy.ts
```diff
@@ -8,6 +8,7 @@ import {AuthnService} from './authn.service';
 
 interface GithubProfile {
   name: string;
```
rbclark, line 10: Name is actually a `string | null`.

apps/backend/src/authn/github.strategy.ts
```diff
@@ -74,9 +75,13 @@ export class GithubStrategy extends PassportStrategy(Strategy, 'github') {
...
       });
-    const {firstName, lastName} = this.authnService.splitName(
-      userInfoResponse.name
-    );
+    let firstName = userInfoResponse.login;
+    let lastName = '';
+    if (typeof userInfoResponse.name === 'string') {
+      firstName = this.authnService.splitName(userInfoResponse.name).firstName;
+      lastName = this.authnService.splitName(userInfoResponse.name).lastName;
+    }
```
rbclark, line 83: Split with the all additional names going into the lastname field instead of first name field.

Try using a ternary operator so we can still use `const {firstName, lastName}` destructuring.

rbclark (changes requested):

Amndeep7 (approved):

PR #1794, opened 2021-08-13
Aws config mapper

Rlin232:


jkufro:
@Rlin232 see https://github.com/mitre/heimdall_tools/pull/115 for an enhancement to HT version

libs/hdf-converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,334 @@
...
+    params: DescribeConfigRulesCommandInput
+  ): Promise<DescribeConfigRulesCommandOutput> {
+    return this.configService.describeConfigRules(params);
+  }
+  private async getResults(
+    configRules: ConfigRule[]
+  ): Promise<ExecJSON.ControlResult[][]> {
+    const complianceResults: ComplianceByConfigRule[] =
+      await this.fetchAllComplianceInfo(configRules);
+    const results = configRules.map(async (rule) => {
```
jkufro, line 72: This map call will simultaneously make requests for every rule, which in any normal use case will almost always result in a throttling exception from AWS. This is a limitation that AWS places on users, so this should be changed to a simple for loop where rules are queried sequentially.

libs/hdf-converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,334 @@
...
+          depends: [],
+          groups: [],
+          status: 'loaded',
+          controls: await this.getControls(),
+          sha256: ''
+        }
+      ]
+    };
+    return hdf;
+  }
```
jkufro, line 333: Should add a delay function to this file so that it can be used to avoid throttling exceptions from AWS.

```js
private delay(ms: number): Promise<void> {
  return new Promise((resolve) => setTimeout(resolve, ms));
}
```

libs/hdf-converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,334 @@
...
+        ConfigRuleName: rule.ConfigRuleName,
+        Limit: 100
+      };
+      let response = await this.configService.getComplianceDetailsByConfigRule(
+        params
+      );
+      let ruleResults = response.EvaluationResults || [];
+      while (response.NextToken !== undefined) {
+        params = _.set(params, 'NextToken', response.NextToken);
+        response = await this.configService.getComplianceDetailsByConfigRule(
```
jkufro, line 84: add `await this.delay(500);` before this line.

libs/hdf-converters/src/aws-config-mapper.ts
```diff
@@ -0,0 +1,334 @@
...
+    } else {
+      return undefined;
+    }
+  }
+  private async fetchAllComplianceInfo(
+    configRules: ConfigRule[]
+  ): Promise<ComplianceByConfigRule[]> {
+    const complianceResults: ComplianceByConfigRule[] = [];
+    for (const rule of configRules) {
+      const response = await this.configService.describeComplianceByConfigRule({
```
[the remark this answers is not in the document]
jkufro: Making this request for every single config rule is inefficient when you can really request 25 rules at once with one call.

Add this helper function to the file:
```js
private chunkArray(sourceArray: Array<any>, chunkSize: number) {
  let result = []
  for (var i = 0; i < sourceArray.length; i += chunkSize) {
    result.push(sourceArray.slice(i, i + chunkSize));
  }
  return result;
}
```

... then change this snippet to something like this:
```js
private async fetchAllComplianceInfo(
  configRules: ConfigRule[]
): Promise<ComplianceByConfigRule[]> {
  const complianceResults: ComplianceByConfigRule[] = [];
  // Should slice config rules into arrays of max size: 25 and make one request for each slice
  const configRuleSlices = this.chunkArray(configRules, 25);
  for (const slice of configRuleSlices) {
    await this.delay(750);
    const response = await this.configService.describeComplianceByConfigRule({
      ConfigRuleNames: slice.map((rule) => rule.ConfigRuleName || '')
    });
    if (response.ComplianceByConfigRules === undefined) {
      throw new Error('No compliance data was returned');
    } else {
      response.ComplianceByConfigRules?.forEach((compliance) =>
        complianceResults.push(compliance)
      );
    }
  }
  return complianceResults;
}
```

jkufro (dismissed):
See all inline comments

camdenmoors (approved):

PR #1819, opened 2021-08-20
No read file

### LINKED ISSUE #1774: Remove ReadFileSync from CweNistMapping
Currently `CweNistMapping` in `hdf-converters` uses `readFileSync` to load the CSV CWE2NIST mappings. This is not supported in the browser, preventing the library from being imported blocking #1731. Being able to manually pass the CSV string after making a web request should work.

georgedias:
Closes #1774 
Adds
 - JSON Converted Mappings for all CSV files
 - Enables `resolveJsonModule` to allow reading the mappings


tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "sourceMap": true,
     "target": "ES2019",
     "esModuleInterop": true,
-    "noImplicitAny": true
+    "noImplicitAny": true,
+    "resolveJsonModule": true
```
camdenmoors, line 13: It looks like this is what's causing the [deploy preview](https://deploy-preview-1819--heimdall-lite.netlify.app/) to show a white page instead of the app. Is this necessary to include for the whole app or just hdf-converters?

.vscode/settings.json
```diff
@@ -0,0 +1,31 @@
+{
```
camdenmoors, line 1: What is this file for?

yarn.lock
```diff
@@ -13870,9 +13875,9 @@ lines-and-columns@^1.1.6:
   integrity sha1-HADHQ7QzzQpOgHWPe2SldEDZ/wA=
 
 listr2@^3.8.3:
-  version "3.10.0"
-  resolved "https://registry.yarnpkg.com/listr2/-/listr2-3.10.0.tgz#58105a53ed7fa1430d1b738c6055ef7bb006160f"
-  integrity sha512-eP40ZHihu70sSmqFNbNy2NL1YwImmlMmPh9WO5sLmPDleurMHt3n+SwEWNu2kzKScexZnkyFtc1VI0z/TGlmpw==
+  version "3.11.0"
+  resolved "https://registry.yarnpkg.com/listr2/-/listr2-3.11.0.tgz#9771b02407875aa78e73d6e0ff6541bbec0aaee9"
+  integrity sha512-XLJVe2JgXCyQTa3FbSv11lkKExYmEyA4jltVo8z4FX10Vt1Yj8IMekBfwim0BSOM9uj1QMTJvDQQpHyuPbB/dQ==
```
rbclark, line 13880: I don't see any new dependencies added, so this change may be out of scope of this PR.

libs/hdf-converters/src/mappings/ScoutsuiteNistMapping.ts
```diff
@@ -1,19 +1,20 @@
...
-    const contents = parse(fs.readFileSync(csvDataPath, {encoding: 'utf-8'}), {
-      skip_empty_lines: true
-    });
-    if (Array.isArray(contents)) {
-      contents.slice(1).forEach((line: string[]) => {
+
+    if (Array.isArray(data)) {
+      data.forEach((line: ISCOUTSUITEJSONID) => {
         this.data.push(new ScoutsuiteNistMappingItem(line));
       });
```
rbclark, line 19: This makes more sense as a map.

libs/hdf-converters/data/owasp-nist-mapping.json
```diff
@@ -0,0 +1,72 @@
+[
```
rbclark, line 1: How were these files generated?

heimdall2.code-workspace
```diff
@@ -25,6 +25,35 @@
...
+      "!GetAtt",
+      "!GetAZs",
+      "!ImportValue",
+      "!ImportValue sequence",
+      "!Select",
+      "!Select sequence",
+      "!Split",
+      "!Split sequence"
+    ]
   }
```
rbclark: Does this belong in the shared workspace?

libs/hdf-converters/src/mappings/CciNistMapping.ts
```diff
@@ -1,40 +1,15 @@
...
-            path = 'references.reference[0].index';
-          }
-        } else {
-          path = 'references.reference.index';
-        }
-        this.data.push(
-          new CciNistMappingItem(_.get(element, 'id'), _.get(element, path))
-        );
+
+    if (data.constructor === {}.constructor) {
```
rbclark, line 10: I'm not sure I understand what this is checking.

rbclark (changes requested):

libs/hdf-converters/data/owasp-nist-mapping.json
camdenmoors: Using a csv2json converter, e.g https://github.com/julien-f/csv2json

libs/hdf-converters/src/mappings/CciNistMapping.ts
camdenmoors: I believe this is checking if data is an Object, moving to `typeof`.

heimdall2.code-workspace
camdenmoors: This looks like something that was automatically added by a plugin on George's computer (yaml extension pack) but isn't necessary for this change, removed now.

libs/hdf-converters/data/owasp-nist-mapping.json
rbclark: If you are going to maintain both of these files in this directory, something should be implemented to ensure these don't get out of sync, similar to how the code is generated for inspecjs.

PR #1846, opened 2021-08-30
Create CKL template and exporter

georgedias:


apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -0,0 +1,240 @@
...
+        if (file) {
+          this.addFiledata(file);
+        }
+      });
+      this.outputData.controlSets.forEach((controlSet) => {
+        const body = Mustache.render(data, controlSet).replace(
+          /[^\x00-\x7F]/g,
+          ''
+        );
+        saveAs(
```
georgedias, line 230: Will need to add these to a zip file!

camdenmoors (approved):

PR #2017, opened 2021-10-11
Automatically Convert Non-HDF Files to HDF on Upload if Supported

### LINKED ISSUE #1731: Integrate HDF Converters with Heimdall Upload
Now that heimdall_tools (and our new HDF Converters package) are written in typescript, it can be integrated in with the Heimdall upload. Allowing users to drop arbitrary burpsuite, nessus, etc scans onto Heimdall would make for a more seamless experience.

The HDF Converters package should be integrated with the Heimdall frontend so that non-HDF uploads can be detected and converted into HDF and then displayed and saved in Heimdall.

camdenmoors:
Closes #1731, automatically convert files based off of multiple fingerprints.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   "dependencies": {
     "@aws-sdk/client-config-service": "^3.23.0",
     "@typescript-eslint/eslint-plugin": "^4.8.0",
-    "@typescript-eslint/parser": "^4.8.0",
+    "@typescript-eslint/parser": "^5.0.0",
     "eslint": "^7.13.0",
     "eslint-config-prettier": "^8.0.0",
     "eslint-plugin-prettier": "^4.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4436,14 +4436,14 @@
     eslint-scope "^5.1.1"
     eslint-utils "^3.0.0"
 
-"@typescript-eslint/parser@^4.8.0":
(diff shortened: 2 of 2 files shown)


apps/frontend/src/store/report_intake.ts
```diff
@@ -93,14 +132,120 @@ export class InspecIntake extends VuexModule {
...
+      'Invalid file uploaded, no fingerprints matched.'
+    );
+  }
+
+  @Action
+  async guessType(parsedJSON: Record<string, unknown>): Promise<string> {
+    // Find the fingerprints that have the most matches
+    const fingerprinted = Object.entries(fileTypeFingerprints).reduce(
+      (a, b) => {
+        return a[1].filter((value) => _.get(parsedJSON, value)).length >
```
georgedias, line 242: This logic is a bit fussy for me to realize how the reduction is occurring. I believe there is a risk here in returning the wrong guess.

camdenmoors: This works by counting the number of matches for every fingerprint in `fileTypeFingerprints`. The type that has the most matches is determined to be the filetype.

I don't believe that this has much of a chance of choosing the wrong value since the only fingerprints that overlap are the 'vulnerabilities' field in Nikto and Snyk, both of which have other fingerprints to avoid such a scenario.

PR #2066, opened 2021-10-15
Enable Secure Cookie on OAuth Authentication Exception

### LINKED ISSUE #1665: STIG Requirement: Set secure flag on cookies

camdenmoors:
Closes #1665, enables Secure flag on OAuth Authentication Failure.


apps/backend/src/filters/authentication-exception.filter.ts
```diff
@@ -1,11 +1,15 @@
...
 @Catch(Error)
 export class AuthenticationExceptionFilter implements ExceptionFilter {
+  configService = new ConfigService();
   catch(exception: Error, host: ArgumentsHost): void {
     const ctx = host.switchToHttp();
     const response = ctx.getResponse();
-    response.cookie('authenticationError', exception.message);
+    response.cookie('authenticationError', exception.message, {
+      secure: this.configService.isInProductionMode()
+    });
```
georgedias, line 12: Here we are returning a Boolean based if we're running a production or other env.
I assume that when the @UseFilters(new AuthenticationExceptionFilter()) is called from the  authn.controller.ts and true is returned the line "this.authnService.login(req.user as User);" forces authentication?

georgedias (approved):

PR #2074, opened 2021-10-18
Remove VueClipboard.config.autoSetContainer

camdenmoors:
The current implement ion of vue-clipboard2 will occasionally fail while setting the options on the Heimdall Netlify preview. This directly calls $copyText instead.

fail read config

### DIFF
--- a/apps/backend/src/filters/authentication-exception.filter.ts
+++ b/apps/backend/src/filters/authentication-exception.filter.ts
@@ -1,11 +1,15 @@
 import {ArgumentsHost, Catch, ExceptionFilter} from '@nestjs/common';
+import {ConfigService} from '../config/config.service';
 
 @Catch(Error)
 export class AuthenticationExceptionFilter implements ExceptionFilter {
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/generic/CopyButton.vue
```diff
@@ -26,12 +17,10 @@ export default class CopyButton extends Vue {
...
-  onCopy() {
-    SnackbarModule.notify('Text copied to your clipboard');
-  }
-
-  onCopyFailure() {
-    SnackbarModule.failure('Failed to copy to your clipboard');
+  copy() {
+    this.$copyText(this.text, document.querySelector('.v-dialog') || undefined)
+      .then(() => SnackbarModule.notify('Text copied to your clipboard'))
+      .catch(() => SnackbarModule.failure('Failed to copy to your clipboard'));
```
georgedias, line 23: Was this implementation of Vue added on a later version - not sure why it wasn't implemented this way from the start!

PR #2150, opened 2021-11-03
Add Reverse ASFF->HDF Mapper

### LINKED ISSUE #2149: Complete Reverse ASFF Mapper

camdenmoors:
Closes #2149, adds the reverse HDF base converter and reverse ASFF mappings.


apps/frontend/src/components/global/ExportASFFModal.vue
```diff
@@ -0,0 +1,112 @@
...
+        @click="showModal"
+        v-on="on"
+      />
+    </template>
+    <v-card>
+      <v-card-title class="headline"> Export as ASFF </v-card-title>
+      <v-card-text>
+        <v-text-field v-model="awsAccountId" label="AWS Account ID" />
+        <v-text-field v-model="target" label="Target Name" />
+        <v-text-field v-model="region" label="Region" />
```
georgedias, line 17: Can we provide an information icon letting the user known what he/she needs to provide for each field. For example:
AWS Account ID -> To find out your ASW account ID, log into AWS, in the navigation bar, choose Support, and then Support Center. Your currently signed-in 12-digit account number (ID) appears in the Support Center navigation pane.
Target Name -> II believe it is the name they want to see in AWS, but list here the correct message
Region -> To find out your ASW Region, log into AWS, and look at the top bar. See [AWS Regions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html)

libs/hdf-converters/src/converters-from-hdf/asff/reverse-asff-mapper.ts
```diff
@@ -0,0 +1,198 @@
...
+  defaultOptions(): IOptions {
+    return {
+      input: '',
+      awsAccountId: '',
+      target: 'default',
+      region: ''
+    };
+  }
+
+  sleep(milliseconds: number) {
```
georgedias, line 147: Add time to the date object to simplify this time delay.

georgedias (changes requested):

apps/frontend/src/components/global/ExportASFFModal.vue
```diff
@@ -0,0 +1,159 @@
...
+          <v-tooltip slot="append" bottom>
+            <template #activator="{on}">
+              <v-icon dark v-on="on" @click="openRegionDocumentation">
+                mdi-information-outline
+              </v-icon>
+            </template>
+            <span
+              >Choose the region that contains the Security Hub instance you
+              would like to upload findings to<br />
+              Click the <v-icon dense>mdi-information-outline</v-icon> icon to
```
georgedias, line 54: Replace Click with (In AWS management console click).

apps/frontend/src/components/global/ExportASFFModal.vue
```diff
@@ -0,0 +1,159 @@
...
+          </v-tooltip>
+        </v-text-field>
+
+        <v-text-field v-model="target" label="Target Name">
+          <v-tooltip slot="append" bottom>
+            <template #activator="{on}">
+              <v-icon dark v-on="on"> mdi-information-outline </v-icon>
+            </template>
+            <span
+              >The target name is a user-defined string that allows tracking
```
georgedias, line 34: Does the Target Name becomes the result set Name in AWS? if so may want to stat it here.

georgedias (changes requested):

georgedias (approved):

Issue #2581, opened 2022-02-14
Issues with importing ASFF generated by `convert:hdf2asff`


ben-harvey:
I'm experiencing some validation issues when trying to import ASFF generated from HDF using `saf convert:hdf2asff` to AWS Security Hub

1) `Region` is not supported as a top-level attribute by the `BatchImportFindings` API
https://github.com/mitre/heimdall2/blob/4b77657a03728b254b97a7b9bff1e96846e7cddd/libs/hdf-converters/src/converters-from-hdf/asff/reverse-asff-mapper.ts#L62
See [ASFF docs](https://docs.aws.amazon.com/securityhub/latest/userguide/asff-top-level-attributes.html).  

2). Validation fails for `FindingProviderFields.Types` if the HDF `profiles.controls.tags.{tag-name}` is an empty Array.
https://github.com/mitre/heimdall2/blob/4b77657a03728b254b97a7b9bff1e96846e7cddd/libs/hdf-converters/src/converters-from-hdf/asff/transformers.ts#L456
This results in a `Types` value that fails input validation by the API.  E.g.
```json
"controls": [
        {
          "tags": {
            "nist": ["SI-10"],
            "cweid": ["400"],
            "cveid": [],  << empty 
            "ghsaid": []
          },
        ...
```
will generate 
```json 
 "Types": [
        ...
        "Tags/nist/SI-10",
        "Tags/cweid/400",
        "Tags/cveid/",
        "Tags/ghsaid/"
      ]
```
which fails with the API error `"ErrorMessage": "Finding does not adhere to Amazon Finding Format. data.FindingProviderFields.Types[8] should match pattern \"^([^/]+)(/[^/]+){0,2}$\"."`

3) This doesn't cause a validation failure, but I don't think it's the desired result:  `us-east-2` is hardcoded into the `GeneratorId` field.  I think perhaps this should be set dynamically based on the provided `region` flag?
https://github.com/mitre/heimdall2/blob/4b77657a03728b254b97a7b9bff1e96846e7cddd/libs/hdf-converters/src/converters-from-hdf/asff/transformers.ts#L296

PR #2601, opened 2022-02-15
Asff import issues

Amndeep7:
Resolves https://github.com/mitre/heimdall2/issues/2581


Amndeep7:
@ejaronne please review the changes in that middle commit.  specifically, for tags that consist of empty arrays, the change I made does not output them in the resultant asff.  ex:

input:

```
"controls": [
        {
          "tags": {
            "nist": ["SI-10"],
            "cweid": ["400"],
            "cveid": [],  << empty 
            "ghsaid": []
          },
        ...
```

old output:
```

 "Types": [
        ...
        "Tags/nist/SI-10",
        "Tags/cweid/400",
        "Tags/cveid/",
        "Tags/ghsaid/"
      ]
```

new output after this change:

```
 "Types": [
        ...
        "Tags/nist/SI-10",
        "Tags/cweid/400"
      ]
```

Nominally this is data loss in that we are no longer tracking a tag that existed, but I'm not sure if that's a problem since the tag was empty.  If you do need that label to go through, what do you want to have in the third block?  An empty array symbol: `[]`?

ejaronne:
@Amndeep7 for context, what source sample created this HDF with blank tags?

ejaronne:
> @ejaronne I'm not sure, I'm deriving it from the code snippet provided in the associated issue

Here is what a snyk original (non-hdf) json might say:

For one result it might be:

      "identifiers": {
        "CVE": [],
        "CWE": [
          "CWE-400"
        ],
        "GHSA": [
          "GHSA-8w57-jfpm-945m"
        ],
        "NSP": [
          1486
        ],
        "WS": [
          "WS-2018-0085"
        ]
      },

in the HDF the CVE has [] (as you know already)

**_For consistency, the Type should have whatever snyk said._** (Not simply [] in all cases. It might be "" in some cases - see final note below) Thankfully our snyk2hdf preserved that. so in this case if the hdf says [], we should use [] (if that is legal for the asff schema.

Note: open a separate issue to double-check that all *2hdf converters do not leave tags (or other hdf schema) blank.

Final note:
As seen here: https://github.com/mitre/apache-site-2.2-stig-baseline/blob/799ba145d5eb490aee089d08ad7e2599ee36f4a5/controls/V-2272.rb#L41-L49
a tag could be "null" in the following way:
`  tag "mitigations": ""`
In this case the hdf should/will have "", and so should the downstream asff Type/mitigation/"" <- again if that's asff-legal

PR #2632, opened 2022-02-21
Allow for string NIST tags in InspecJS and add tests

camdenmoors:
Closes https://github.com/mitre/saf/issues/154


libs/inspecjs/src/compat_impl/compat_inspec_1_0.ts
```diff
@@ -47,7 +47,11 @@ abstract class HDFControl10 implements HDFControl {
     this.waived = waived;
     this.isProfile = isProfile;
 
-    this.rawNistTags = HDFControl10.compute_raw_nist_tags(this.wraps);
+    const tempNistTags = HDFControl10.compute_raw_nist_tags(this.wraps);
+
+    this.rawNistTags = Array.isArray(tempNistTags)
+      ? tempNistTags
+      : [tempNistTags];
```
georgedias, line 54: May want to comment here and stated that rawNisTags must be an array (and possible the reason why)

georgedias (dismissed):
Minor comments clarification

PR #2635, opened 2022-02-21
Replace Slashes in Profile Name for ASFF Export

### LINKED ISSUE #2633: `hdf2asff`: replace slashes in `profile.name` when creating the profile info finding
As noted in the comments for this constant, the AWS API will fail an import if the `Types` field of ASFF contains more than 2 forward slashes:
https://github.com/mitre/heimdall2/blob/5800edfd37f1903dd4501191731a3f22be989bed/libs/hdf-converters/src/converters-from-hdf/asff/reverse-asff-mapper.ts#L29

In `createProfileInfoFindingFields`, the `profile.name` can contain forward slashes (e.g. for JSON generated by the Snyk CLI, the value is `Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities`, and these should be replaced in the same manner as elsewhere.
https://github.com/mitre/heimdall2/blob/5800edfd37f1903dd4501191731a3f22be989bed/libs/hdf-converters/src/converters-from-hdf/asff/transformers.ts#L391

camdenmoors:
Closes #2633


libs/hdf-converters/src/converters-from-hdf/asff/transformers.ts
```diff
@@ -388,7 +388,12 @@ function createProfileInfoFindingFields(hdf: ExecJSON.Execution): string[] {
     targets.forEach((target) => {
       const value = _.get(profile, target);
       if (typeof value === 'string') {
-        typesArr.push(`${profile.name}/${target}/${value}`);
+        typesArr.push(
+          `${profile.name}/${target}/${value.replace(
```
Amndeep7, line 392: Do you think we should optimize for the general case and preemptively do this replacement on the profile name and target as well?

Amndeep7 (approved):

PR #2675, opened 2022-02-28
Splunk2HDF Updates

camdenmoors:
- [x] Supporting older InSpecJS formats with descriptions missing e.g [Good NGINX Results](https://github.com/mitre/heimdall2/blob/master/apps/frontend/public/static/samples/good_nginxresults.json)
- [x] Adding insecure mode 
- [x] Adding logging
- [x] Switch to using https://github.com/splunk/splunk-sdk.js for splunk2hdf
- [x] document help on heimdall2 splunk config expectations - gui / md
- [x] throw better error and display than login when its cors, uri, or something not really user/pass
- [x] allow switching search index
- [x] allow general searches
- [x] allow save to splunk fron heimdall
- [x] Write support documentation

Fixes https://github.com/mitre/saf/issues/175 https://github.com/mitre/saf/issues/174 https://github.com/mitre/saf/issues/172 https://github.com/mitre/saf/issues/171 

Related https://github.com/mitre/saf/issues/154


libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -231,9 +244,11 @@ export function createControlMapping(
       path: 'data.descriptions',
       transformer: (data: {label: string; data: string}[]) => {
         const descObjects: Record<string, string> = {};
-        data.forEach((item) => {
-          descObjects[item['label']] = item['data'];
-        });
+        if (Array.isArray(data)) {
```
aaronlippold, line 247: Will this still work for profiles that use tags for check, fix etc. They would just be the key value pair?

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -335,10 +365,19 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
+      'Data converted, uploading to Splunk with config: ' +
+        JSON.stringify(config)
+    );
     uploads.push(...postDataToSplunkHEC(splunkData.reports, config));
     uploads.push(...postDataToSplunkHEC(splunkData.profiles, config));
     uploads.push(...postDataToSplunkHEC(splunkData.controls, config));
 
-    return Promise.all(uploads).then(() => guid);
+    return Promise.all(uploads).then(() => {
+      logger?.info(
```
aaronlippold, line 377: Is logger.info used during normal run-time? aka I don't have to specifiy `loglevel=info` ? 



@em-c-rod @Amndeep7 @camdenmoors 
I guess we should decide what is logged in:
Normal Mode
Debug Mode
Verbose mode - if we need it
`--quite` or `--silent` is the gnu standard to quite all stdout other than `exit 1/0`

Amndeep7:
@aaronlippold 

The types as defined should be correct.

`ExecJSONProfile.parent_profile` is a conditional string:
https://github.com/mitre/heimdall2/blob/5f9480d8698b8ffb2796040096293816599e2de2/libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts#L41
https://github.com/mitre/inspec/blob/0383e819eb35246f1a31a90d736e94badbfcd488/lib/inspec/schema/exec_json.rb#L95

`ExecJSONProfile.depends` (which I think is what you were referring to with `dependencies`) is an array of type `Dependency` which itself is an object with no required attributes but some expected ones include `name` or `url`:
https://github.com/mitre/heimdall2/blob/5f9480d8698b8ffb2796040096293816599e2de2/libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts#L34
https://github.com/mitre/inspec/blob/0383e819eb35246f1a31a90d736e94badbfcd488/lib/inspec/schema/exec_json.rb#L94
https://github.com/mitre/inspec/blob/0383e819eb35246f1a31a90d736e94badbfcd488/lib/inspec/schema/primitives.rb#L167

if we need to refer to both types as one object, then it should probably be `null | string | Record<string, string>[]`

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
Amndeep7: this comment will be talking more from the perspective of the saf cli than the library itself

presumably normal will just be states that we are passing through.  at minimum (and in most cases maximum), it will be 1) received input with x flags set, 2) processed input into hdf (or asff/xccdf/whatever), 3) (successfully or not) saved to file/uploaded to service the output.

debug mode would be making processes more visible.  i think io would fall in here - are we failing/succeeding (and if necessary how/why) with creating a file?  are we failing/succeeding (and if necessary how/why) with creating a connection to an external service?

verbose mode would be any data that is even more granular.  ex. if we decide to add any additional logging for where we are in the processing of an input file within hdf converters, it would go here (processed x/y findings from input into hdf/whatever) - personally I think this type of logging is useless except when trying to give the user a progress bar of some sort, which i think at the current speed of our mappers is not that useful since they usually complete within 30s at least with the possible inputs that i've seen thus far.

i don't think this pr can nor should resolve https://github.com/mitre/saf/issues/172

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
camdenmoors: Yes these are still preserved, you'll retain both tags and descriptions

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
camdenmoors: The way this is setup is when this is run without a logging service or log level passed to it (e.g within Heimdall Lite), it will log debug messages by default, and would show inside of the browser console if opened. When done in the SAF CLI, we will allow the user to provide the option of what verbosity level they want to use, with a **default** of **Info**. All logging related to IO should be done within the application calling the mapper, not the mapper itself.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -329,104 +330,119 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
     return splunkData;
   }
 
-  uploadSplunkData(targetIndex: any, splunkData: SplunkData) {
-    // Upload execution event
-    splunkData.reports.forEach((report) => {
+  async uploadSplunkData(
+    targetIndex: any,
```
georgedias, line 334: Change index to be of type splunk type Index

georgedias (changes requested):

PR #2827, opened 2022-03-25
Convert Mappings into Typescript Objects

camdenmoors:
Fixes some issues with Heimdall crashing on load. Closes https://github.com/mitre/saf/issues/236


libs/hdf-converters/src/mappings/NiktoNistMapping.ts
```diff
@@ -1,34 +1,20 @@
...
+import {data} from './NiktoNistMappingData';
 
 export interface INIKJSONID {
   'NIKTO-ID': number;
   'PLUGIN-CATEGORY': string;
   'NIST-ID': string;
   OSVDB: number;
 }
-const DEFAULT_NIST_TAG = ['SA-11', 'RA-5'];
+const DEFAULT_NIST_TAG = ['AC-3', 'SA-11', 'RA-5'];
```
ejaronne, line 9: It is unclear why this DEFAULT was changed to add AC-3. I can't find an issue on this.

PR #2900, opened 2022-04-11
Ensure valid types for Nessus Mapper segments[].message

camdenmoors:


camdenmoors:
Looks like we don't have any tests for this. I'm keeping this open until tests are finished. Non-sensitive sample data:
[Home_Network_4v072v.txt](https://github.com/mitre/heimdall2/files/8467605/Home_Network_4v072v.txt)

Amndeep7 (changes requested):

libs/hdf-converters/src/nessus-mapper.ts
```diff
@@ -320,7 +319,8 @@ export class NessusMapper extends BaseConverter {
                 status: {transformer: getStatus},
                 code_desc: {transformer: formatCodeDesc},
                 message: {
-                  path: ['plugin_output', 'cm:compliance-actual-value']
+                  path: ['plugin_output', 'cm:compliance-actual-value'],
```
Amndeep7, line 322: Also please add tests like you mentioned in the comment above.

Amndeep7 (approved):

PR #2957, opened 2022-04-25
Add IonChannel SBOM Converter (Dependencies Only)

camdenmoors:


libs/hdf-converters/src/ionchannel-mapper.ts
```diff
@@ -0,0 +1,346 @@
...
+      );
+    }
+  }
+
+  async getTeams(): Promise<Team[]> {
+    if (this.apiKey) {
+      return this.apiClient
+        .get('https://api.ionchannel.io/v1/teams/getTeams')
+        .then(({data}) => data.data);
+    } else {
```
Amndeep7, line 160: this happens a few times in this file so i'll only call it out here, but i personally prefer to reduce the indent levels of stuff as much as possible by putting guard clauses at the top and then letting the rest of the code do its thing

in this case i'd do a check if apikey is not defined (in which case throw the error), and then let the rest of the code be at that top level instead of nested in a if statement

Amndeep7 (changes requested):

Amndeep7:
@camdenmoors thanks for addressing the review.  i spotted a missing guard clause and some errors that weren't in guard clause format so fixed that.  can you retest against the ionchannel system you have access to in order to confirm if everything still works?  we can merge after that.

libs/hdf-converters/src/ionchannel-mapper.ts
```diff
@@ -0,0 +1,337 @@
...
+
+    this.apiClient = axios.create();
+    this.apiClient.defaults.headers.common[
+      'Authorization'
+    ] = `Bearer ${this.apiKey}`;
+    this.apiClient.defaults.headers.common['Accept'] =
+      'application/json, text/plain, */*';
+  }
+
+  async toHdf(): Promise<ExecJSON.Execution> {
```
Amndeep7, line 133: test the code after my additional commits then should be good to go

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #3059, opened 2022-05-20
Veracode mapper2

jweissm:
code for veracode hdf mapper


libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+  } else {
+    return 0.1;
+  }
+}
+
+function nistTag(input: string): string[] {
+  const len = Number(`${_.get(input, CWE_LENGTH)}`)
+  const cwes = []
+  for (let index = 0; index < len; index++) {
+    cwes.push(`${_.get(input, 'cwe.cweid')}`)
```
Amndeep7, line 41: i'd appreciate it if you could write out to a file the json file that was derived from converting it from xml (not the hdf file) and sending it to me over slack.

anyways does not seem to be incorporating the index into which cweid it's getting which i think also probably explains why it's getting the default nist tags every time (iirc) cause cwe.cweid is not a thing - think you'd need cwe[index].cweid

anyways, instead of using for loops, we could probably use a map here - something like `const cwes = _.get(input, 'cwe').map(({cweid}) => cweid);`

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+  }
+  return CWE_NIST_MAPPING.nistFilter(cwes, DEFAULT_NIST_TAG).concat(['Rev_4']);
+}
+
+function formatRecommendations(input: unknown): string {
+  const text = [];
+  if (_.has(input, 'recommendations.para')) {
```
Amndeep7, line 48: my understanding from reading this code is that recommendations.para could be a single object or an array of those objects from which we need to extract this text attribute

i would replace the inside of this if block with something like the following:

```
if(Array.isArray(_.get(input, 'recomenndations.para'))) {
that same map trick from above but maybe not using destructuring unless you can guarantee that each one of these para objects contains 'text' which i think we can't cause the non-array variant needs to check if it's there
} else {
what you got there but with lodash has against recommendations.para.text instead of doing a get and comparing it to undefined
}
```

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+    }
+    else {
+      const len = Number(`${_.get(input, 'recommendations.para.length')}`)
+      for (let index = 0; index < len; index++) {
+        const topush = `recommendations.para[${index}].text`
+        text.push(`${_.get(input, topush)}`)
+      }
+    }
+  }
+  if (_.has(input, 'recommendations.para.bulletitem')) {
```
Amndeep7, line 62: see above.  basically for loops can be annoying and a possible source of errors so i try to avoid them if possible

also is text guaranteed to be within bulletitem?  otherwise might wanna do a has check here too for that

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+    const len = Number(`${_.get(input, 'recommendations.para.bulletitem.length')}`)
+    for (let index = 0; index < len; index++) {
+      const bulletitem = `recommendations.para.bulletitem[${index}].text`
+      text.push(`${_.get(input, bulletitem)}`)
+    }
+  }
+  return text.join('\n');
+}
+
+function formatDesc(input: unknown): string {
```
Amndeep7, line 72: see above

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+      for (let index = 0; index < len; index++) {
+        const paratext = `desc.para[${index}].text`
+        text.push(`${_.get(input, paratext)}`)
+      }
+    }
+  }
+  return text.join('\n');
+}
+
+function cweloop(input: unknown, index:integer ): string{
```
Amndeep7, line 91: linter probably gonna pitch a fit here haha

Amndeep7, line 91: also generally i like to at least push down the unknown as far as i can.  like we're reasonably confident that input is an object so why don't we type it so: `input: Record<string, unknown>` - it's not like a LOT better, but at least it's something

Amndeep7: also why is this even a separate function instead of being embedded as the loop body within formatcwedata?

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+
+  let inputstr = `cwe[${index}].cweid`
+      let cwe = 'CWE-'.concat(`${_.get(input, inputstr)}`) + ': ';
+      const cwename = `cwe[${index}].cwename`
+      cwe += `${_.get(input, cwename)}`
```
Amndeep7, line 96: just embed the template string in here instead of creating a variable cause it doesn't appear to be reused in the rest of the func

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+      }
+    }
+  }
+  return text.join('\n');
+}
+
+function cweloop(input: unknown, index:integer ): string{
+
+  let inputstr = `cwe[${index}].cweid`
+      let cwe = 'CWE-'.concat(`${_.get(input, inputstr)}`) + ': ';
```
Amndeep7, line 94: linter'll probably yell at you too, but use a template string here instead of a concat and also a +

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+      const cwename = `cwe[${index}].cwename`
+      cwe += `${_.get(input, cwename)}`
+      inputstr = `cwe[${index}].pcirelated`
+      if (`${_.get(input, inputstr)}` !== 'undefined') {
```
Amndeep7, line 98: use lodash has instead of checking against undefined

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+  return text.join('\n');
+}
+
+function cweloop(input: unknown, index:integer ): string{
+
+  let inputstr = `cwe[${index}].cweid`
+      let cwe = 'CWE-'.concat(`${_.get(input, inputstr)}`) + ': ';
+      const cwename = `cwe[${index}].cwename`
+      cwe += `${_.get(input, cwename)}`
+      inputstr = `cwe[${index}].pcirelated`
```
Amndeep7, line 97: you can probably consolidate this by making pcirelated, owasp, etc into an array of strings that you can then iterate over (without using a for loop!) and substitute in as appropriate so you only need to have written the one if block

Amndeep7 (changes requested):
I'll leave more comments later, but what i've done thus far will give you plenty to do and also the patterns I'm recommending you change from will probably occur later on in this file as well so you can just get ahead of me and fix those up as well

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+  const len = Number(`${_.get(input, 'length')}`);
+  for (let index = 0; index < len; index++) {
+    const source = `[${index}].sourcefile`;
+    text.push(`${_.get(input, source)}`);
+  }
+  return text.join('\n');
+}
+
+function stringifyCodeDesc(input: Record<string, unknown>): string {
+  const categories = [
```
Amndeep7, line 145: would probably make this data structure a map or even a string based enum

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    ['Exploit level', 'exploitLevel'],
+    ['Issue ID', 'issueid'],
+    ['Module', 'module'],
+    ['Type', 'type'],
+    ['CWE ID', 'cweid'],
+    ['Date First Occurence', 'date_first_occurence'],
+    ['CIA Impact', 'cia_impact'],
+    ['Description', 'description']
+  ];
+  let flawDesc =
```
Amndeep7, line 158: make the entire line a template string

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    'Sourcefile Path: ' + `${_.get(input, 'sourcefilepath')}` + ';';
+  categories.forEach(function (value) {
```
Amndeep7, line 160: i recommend `for of` loops over foreach when es6 is available (which it is)

https://2ality.com/2021/01/looping-over-arrays.html - really solid article explaining the different for loop structures in js

anyways, i'd still recommend map/reduce (with the reduce being the built-in join func) here

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    if (_.has(input, value[1])) {
+      flawDesc += `${value[0]}: ` + `${_.get(input, value[1])}` + ';';
+    }
+  });
+  return flawDesc;
+}
+
+function formatCodeDesc(input: Record<string, unknown>): string {
```
Amndeep7, line 168: i'd inline the stringify func in here.  you can also probably get rid of the text array stuff since there doesn't seem to be any iteration going on here - unless there ought to be iteration going on in here haha in which case follow the map/join paradigm outlined above

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    'added_date',
+    'component_affects_policy_compliance'
+  ];
+  let flawDesc = 'component_id: ' + `${_.get(input, 'component_id')}` + ';';
+  categories.forEach(function (value) {
+    if (_.has(input, value)) {
+      flawDesc += `${value}: ` + `${_.get(input, value)}` + ';';
+    }
+  });
+  if (`${_.has(input, 'file_paths.file_path.value')}`) {
```
Amndeep7, line 194: if check against template string - please make sure to go through and exhaustively remove all of these

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    flawDesc +=
+      'file_path: ' + `${_.get(input, 'file_paths.file_path.value')}` + ';';
+  }
+  return flawDesc;
+}
+
+function formatSCACodeDesc(input: Record<string, unknown>): string {
+  const text = [];
+  if (`${_.get(input, 'component_id')}` !== 'undefined') {
+    const flawDesc = SCAstringify(input);
```
Amndeep7, line 204: in most cases, separating out these stringify functions is unnecessary unless it'd be drastically clearer being split out or if you're running into function length caps as set by the linter.  so i would again recommend inlining that function into here.

also like above there's no iteration going on here but you're still doing the array thing.  either the iteration is missing or the array stuff is unnecessary.

Amndeep7 (changes requested):
sorry that the review still isn't complete - i was unexpectedly busy today, but i'll wrap it up tomorrow.  i'll be resuming from the staticflawmanip func.

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+        omitted = _.omit(omitted, `${SCA_VULNERABILITIES}${l}].cves`);
+        components.push(_.get(omitted, `${SCA_VULNERABILITIES}${l}]`));
+      }
+    }
+    _.set(parsedXML, `${SCA_CVES}${m}].components`, components);
+  }
+  return parsedXML;
+}
+
+function cvesformatter(parsedXML: any) {
```
Amndeep7, line 312: delete this function

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+      inputstr1 = `${REPORT_SEVERITY}${i}].category.length`;
+      for (let k = 0; k < Number(_.get(parsedXML, inputstr1)); k++) {
+        _.set(parsedXML, '', staticflawmanip(parsedXML, i, k));
+      }
+    }
+  }
+  return parsedXML;
+}
+
+function setcomponents(parsedXML: any) {
```
Amndeep7, line 245: delete this function

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+  return parsedXML;
+}
+function parseXml(xml: string) {
+  const options = {
+    attributeNamePrefix: '',
+    textNodeName: 'text',
+    ignoreAttributes: false
+  };
+
+  const parsedXML = parser.parse(xml, options);
```
Amndeep7, line 384: ok on further review, i think doing any of this pre-processing like was done here is more or less wrong / lends itself to bad practices and unreadable code.

the functional issue is that some attributes are a bit harder to get to like staticflaws.flaws and some sections require a flip (components -> vulnerabilities/cves => cves -> components).  i've already detailed how to do the flaws one.  the other one is not too bad either because of the `arrayTransform` - see the `consolidate` function in the asffmapper but basically you parse the file as is and generate an array that'll look like in a very abstract sense `[{cve: info for the first cve, componentsAffected: [only the first component]}, ..., {cve: info for the mth cve, componentsAffected: [only the first component]}, ..., {cve: info for the first cve, componentsAffected: [only the nth component]}, ..., {cve: info for the mth cve, componentsAffected: [only the nth component]}]` and then you reduce/consolidate this array based off of the cve's id to create an array that looks like `[{cve: info for the first cve, componentsAffected: [first component, ..., nth component]}, ..., {cve: mth cve, componentsAffected: [first component, ..., nth component]}]`.  it sounds more complicated than it actually is.

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+      }
+    ],
+    impact: impactMapping(severity),
+    refs: [],
+    tags: {
+      cweid: {transformer: formatCweData},
+      cweDescription: {transformer: formatCweDesc},
+      nist: {transformer: nistTag}
+    },
+    code: {path: ''},
```
Amndeep7, line 456: remove

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    source_location: {
+      ref: {
+        path: 'staticflaws.flaw',
+        transformer: formatSourceLocation
+      },
+      line: 0
```
Amndeep7, line 462: remove

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    },
+    results: [
+      {
+        path: 'staticflaws.flaw',
+        status: ExecJSON.ControlResultStatus.Failed,
+        code_desc: {transformer: formatCodeDesc},
+        run_time: 0,
```
Amndeep7, line 469: remove

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+        start_time: {path: '$.detailedreport.first_build_submitted_date'},
+        message: 'exploitability_adjustments.exploitability_adjustment.note',
+        resource: ''
```
Amndeep7, line 472: remove

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+      line: 0
+    },
+    results: [
+      {
+        path: 'staticflaws.flaw',
+        status: ExecJSON.ControlResultStatus.Failed,
+        code_desc: {transformer: formatCodeDesc},
+        run_time: 0,
+        start_time: {path: '$.detailedreport.first_build_submitted_date'},
+        message: 'exploitability_adjustments.exploitability_adjustment.note',
```
Amndeep7, line 471: this is a hardcoded string and not a path like i think it was intended to be

Amndeep7 (changes requested):

Amndeep7:
oh yeah, forgot to add that actually it seems like a lot of data isn't being put into the hdf format.  if you can't find a good place to put it in there, shove all the rest of it into the passthrough attribute (see the snyk mapper as an example).  a key part of the normalization process is to not leave any data behind.

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,555 @@
...
+function controlMappingCve() {
+  return {
+    id: {path: 'cve_id'},
+    title: {path: 'cve_id'},
+    desc: {path: 'cve_summary'},
+    impact: {path: 'impact'},
+    refs: [],
+    tags: {
+      cwe: {path: 'cwe_id'},
+      nist: {path: 'nist'}
```
Amndeep7, line 476: currently all of them are ra-5, si-2 - if this is intentional, just hardcode that array in here.  if you can find nist tags for them, then your code is buggy

Amndeep7 (changes requested):

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+  return IMPACT_MAPPING.get(severity.toString()) || 0.1;
+}
+
+function nistTag(input: string): string[] {
+  const cwes = _.get(input, 'cwe').map((value: any) => _.get(value, 'cweid'));
+  return CWE_NIST_MAPPING.nistFilter(cwes, DEFAULT_NIST_TAG);
+}
+
+function formatRecommendations(input: Record<string, unknown>): string {
+  let text: any[] = [];
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+  ['1', 0.1],
+  ['0', 0.0]
+]);
+
+function impactMapping(severity: number): number {
+  return IMPACT_MAPPING.get(severity.toString()) || 0.1;
+}
+
+function nistTag(input: string): string[] {
+  const cwes = _.get(input, 'cwe').map((value: any) => _.get(value, 'cweid'));
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+      target_id: ''
+    },
+    version: HeimdallToolsVersion,
+    statistics: {
+      duration: 0
+    },
+    profiles: [
+      {
+        name: {path: 'detailedreport.policy_name'},
+        version: {path: 'detailedreport.static-analysis.version'},
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+export class VeracodeMapper extends BaseConverter {
+  mappings: MappedTransform<ExecJSON.Execution, ILookupPath> = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion,
+      target_id: ''
+    },
+    version: HeimdallToolsVersion,
+    statistics: {
+      duration: 0
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+function formatCodeDesc(input: Record<string, unknown>): string {
+  const text = [];
+  if (`${_.get(input, 'sourcefilepath')}` !== 'undefined') {
+    const flawDesc = stringifyCodeDesc(input);
+    text.push(flawDesc);
+  }
+  return text.join('\n');
+}
+
+function SCAstringify(input: Record<string, unknown>): string {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    'pcrirelated',
+    'owasp',
+    'owasp2013',
+    'sans',
+    'certc',
+    'certccp',
+    'certjava',
+    'owaspmobile'
+  ];
+  let cwe = 'CWE-' + `${_.get(input, `cwe[${index}].cweid`)}` + ': ';
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+    const len = Number(`${_.get(input, CWE_LENGTH)}`);
+    for (let index = 0; index < len; index++) {
+      const cwe = cweloop(input, index);
+      text.push(cwe);
+    }
+  }
+  return text.join('\n');
+}
+
+function formatCweDesc(input: Record<string, unknown>): string {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,587 @@
...
+      const desc = `cwe[${index}].description.text.text`;
+      cwe += `${_.get(input, cwename)}` + ': ';
+      cwe += `${_.get(input, desc)}`;
+      text.push(cwe);
+    }
+  }
+  return text.join('\n');
+}
+
+function formatSourceLocation(input: Record<string, unknown>): string {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: stayed as array, uses string destructuring

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,552 @@
...
+          },
+        ],
+        sha256: ''
+      }
+    ]
+  };
+  constructor(xml: string) {
+    super(parseXml(xml));
+  }
+  setMappings(
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    ]
+  };
+}
+
+export class VeracodeMapper extends BaseConverter {
+  mappings: MappedTransform<ExecJSON.Execution, ILookupPath> = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion,
+      target_id: ''
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+}
+
+function formatRecommendations(input: Record<string, unknown>): string {
+  let text: any[] = [];
+  if (_.has(input, 'recommendations.para')) {
+    if (_.has(input, 'recommendations.para.text')) {
+      text.push(`${_.get(input, 'recommendations.para.text')}`);
+    }
+    else {
+      text = (_.get(input, `recommendations.para`) as Record<string, unknown>[]).map( function(value: any) {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+        return _.get(value, 'text')
+      });
+    }
+  }
+  if (_.has(input, 'recommendations.para.bulletitem')) {
+    if (_.has(input, 'recommendations.para.bulletitem.text')) {
```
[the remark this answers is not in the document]
jweissm: should be handled correctly

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+      text.push(`${_.get(input, 'recommendations.para.bulletitem.text')}`);
+
+    }
+    else {
+      const text2 = (_.get(input, `recommendations.para.bulletitem`) as Record<string, unknown>[]).map( function(value: any) {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+  return text.join('\n');
+}
+
+function formatDesc(input: Record<string, unknown>): string {
+  const text = [];
+  if (_.has(input, 'desc.para')) {
+    if (_.has(input, 'desc.para.text')) {
+      text.push(`${_.get(input, 'desc.para.text')}`);
+    } else {
+      const len = Number(`${_.get(input, 'desc.para.length')}`);
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    'owasp',
+    'owasp2013',
+    'sans',
+    'certc',
+    'certccp',
+    'certjava',
+    'owaspmobile'
+  ];
+  if (_.has(input, 'cwe')) {
+    const len = Number(`${_.get(input, CWE_LENGTH)}`);
```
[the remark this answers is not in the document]
jweissm: should be fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    ['Type', 'type'],
+    ['CWE ID', 'cweid'],
+    ['Date First Occurence', 'date_first_occurence'],
+    ['CIA Impact', 'cia_impact'],
+    ['Description', 'description']
+  ];
+  if (_.has(input, 'sourcefilepath')) {
+    flawDesc = `Sourcefile Path: ${_.get(input, 'sourcefilepath')};`;
+
+    flawDesc += categories.map(function(value: string[]) {
```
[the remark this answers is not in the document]
jweissm: should use arrow functions now

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+      if (_.has(input, value[1])) {
+          return `${value[0]}: ${_.get(input, value[1])};`;
+      }
+      else{
+          return ''
+      }
+    }).join('');
```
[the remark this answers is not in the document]
jweissm: changed back to '', because of how mapping works joining with \n can add unnecessary newlines when a property is missing

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    'version',
+    'library',
+    'vendor',
+    'descritpion',
+    'added_date',
+    'component_affects_policy_compliance'
+  ];
+  if (_.has(input, 'component_id')) {
+    flawDesc = `component_id: ${_.get(input, 'component_id')};`;
+    flawDesc += categories.map(function(value: string) {
```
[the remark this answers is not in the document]
jweissm: does .push now

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+      line: 0
+    },
+    results: [
+      {
+        path: 'components',
+        status: ExecJSON.ControlResultStatus.Failed,
+        code_desc: {transformer: formatSCACodeDesc},
+        run_time: 0,
+        start_time: {path: '$.detailedreport.first_build_submitted_date'},
+        message: '',
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    source_location: {
+      ref: {path: 'paths'},
+      line: 0
+    },
+    results: [
+      {
+        path: 'components',
+        status: ExecJSON.ControlResultStatus.Failed,
+        code_desc: {transformer: formatSCACodeDesc},
+        run_time: 0,
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    version: HeimdallToolsVersion,
+    statistics: {
+      duration: 0
+    },
+    profiles: [
+      {
+        name: {path: 'detailedreport.policy_name'},
+        version: {path: 'detailedreport.static-analysis.version'},
+        title: {path: 'detailedreport.static-analysis.modules.module.name'},
+        maintainer: null,
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+        summary: null,
+        license: null,
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+        copyright: 'Copyright Veracode, Inc., 2014.',
+        copyright_email: null,
+        supports: [],
+        attributes: [],
+        depends: [],
```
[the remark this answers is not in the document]
jweissm: removed depends: []

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+function parseXml(xml: string) {
+  const options = {
+    attributeNamePrefix: '',
+    textNodeName: 'text',
+    ignoreAttributes: false
+  };
+
+  const parsedXML = parser.parse(xml, options);
+  if (_.has(parsedXML, 'summaryreport')) {
+    throw new Error('Current mapper does not accept sumarry reports');
```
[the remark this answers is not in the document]
jweissm: different format, fixed spelling

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,576 @@
...
+    inputstr = `${REPORT_SEVERITY}${i}${CATEGORY}${k}].cwe[0].staticflaws`;
+    _.set(parsedXML, inputstr, null);
+  }
+  const flaws = {flaw: flawArr};
+  inputstr = `${REPORT_SEVERITY}${i}${CATEGORY}${k}].staticflaws`;
+  _.set(parsedXML, inputstr, flaws);
+  return parsedXML;
+}
+
+function staticflawmover(parsedXML: any) {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,486 @@
...
+}
+
+export class VeracodeMapper extends BaseConverter {
+  mappings: MappedTransform<ExecJSON.Execution & {passthrough: unknown}, ILookupPath> = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion,
+      target_id: ''
+    },
+    passthrough: {
```
Amndeep7, line 378: this didn't need to be specified as such.  just call it `data_remainder` and then make the contents the full data but with an omit around it to get rid of the specific stuff that you used.

apps/frontend/src/store/report_intake.ts
```diff
@@ -21,6 +21,7 @@ import {
   SnykResults,
   TwistlockMapper,
   XCCDFResultsMapper,
+  VeracodeMapper,
```
Amndeep7, line 24: there was a change in behavior that I believe got merged where the fingerprinting is now in hdfconverters instead of here in heimdall proper.  reach out to @Rlin232 for information on this.

libs/hdf-converters/src/base-converter.ts
```diff
@@ -10,6 +10,7 @@ export interface ILookupPath {
   path?: string | string[];
   transformer?: (value: any) => unknown;
   arrayTransformer?: (value: unknown[], file: any) => unknown[];
+  pathTransform?: (value: unknown, file: any) => unknown;
```
Amndeep7, line 13: these changes will probably need to be rebased on top of baseconverter depending on which pr gets merged first

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,486 @@
...
+          },
+          {
+            path: 'detailedreport.severity[4].category',
+            ...controlMappingCwe(1)
+          },
+          {
+            path: 'detailedreport.severity[5].category',
+            ...controlMappingCwe(0)
+          },
+          {
```
Amndeep7, line 454: this is actually better as a separate function - makes it clearer that it's to handle mapping cves

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,486 @@
...
+            path: 'detailedreport.software_composition_analysis.vulnerable_components',
+            pathTransform: componenttransform,
+            id: {path: "cve_id"},
+            title: {path: 'cve_id'},
+            desc: {path: 'cve_summary'},
+            impact: {path: 'impact'},
+            refs: [],
+            tags: {
+              cwe: {path: 'cwe_id'},
+              nist: {path: 'nist'}
```
Amndeep7, line 464: there's gonna be a pr that'll be merged soon that'll add another tag for ccis, so might need to add it if that one gets merged first

https://github.com/mitre/heimdall2/pull/3170

it'll need to be supported on both the cve and the cwe mappers

Amndeep7 (changes requested):

libs/hdf-converters/src/base-converter.ts
jweissm: rebased code

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: still exists in once case where can exist, with value of 'undefined'

libs/hdf-converters/src/veracode-mapper.ts
jweissm: remade controlMappingCve

libs/hdf-converters/src/veracode-mapper.ts
jweissm: changed passthrough to dataremainder, added formatPassthorughData

libs/hdf-converters/src/veracode-mapper.ts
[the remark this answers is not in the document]
jweissm: wrapped categories.map in lodash.compact, removed else {return ''}

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+        }
+      }
+      if (hascve === true) {
+        const proxy = _.cloneDeep(component);
+        components.push(_.omit(proxy, `vulnerabilities`));
+      }
+    }
+    _.set(vuln, `components`, components);
+  })
+  return vulnarr as unknown;
```
Amndeep7, line 368: remove cast

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+        componentlist.push(value);
+      }
+    }
+  } else {
+    componentlist.push(_.get(input, 'component') as Record<string, unknown>);
+  }
+  return componentlist
+}
+
+function componentTransform(input: unknown) {
```
Amndeep7, line 326: I think the implementation of the three relevant functions here can be simplified.  We can discuss again what this'd look like vs how you're currently implementing it.  Basically, I believe there is probably a more functional and consolidated approach than the one you're doing right now.  Imma repaste my comment talking about it from last time:

----



ok on further review, i think doing any of this pre-processing like was done here is more or less wrong / lends itself to bad practices and unreadable code.

the functional issue is that some attributes are a bit harder to get to like staticflaws.flaws and some sections require a flip (components -> vulnerabilities/cves => cves -> components). i've already detailed how to do the flaws one. the other one is not too bad either because of the arrayTransform - see the consolidate function in the asffmapper but basically you parse the file as is and generate an array that'll look like in a very abstract sense [{cve: info for the first cve, componentsAffected: [only the first component]}, ..., {cve: info for the mth cve, componentsAffected: [only the first component]}, ..., {cve: info for the first cve, componentsAffected: [only the nth component]}, ..., {cve: info for the mth cve, componentsAffected: [only the nth component]}] and then you reduce/consolidate this array based off of the cve's id to create an array that looks like [{cve: info for the first cve, componentsAffected: [first component, ..., nth component]}, ..., {cve: mth cve, componentsAffected: [first component, ..., nth component]}]. it sounds more complicated than it actually is.

----

We're not gonna do an array transform, but the idea remains more or less consistent.

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+    title: {path: 'cve_id'},
+    desc: {path: 'cve_summary'},
+    impact: {path: 'impact'},
+    refs: [],
+    tags: {
+      cwe: {path: 'cwe_id'},
+      nist: {
+        path: 'cwe_id',
+        transformer: (value: string) => {
+          if (isArray(value)) {
```
Amndeep7, line 386: dont use lodash isarray

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+          } else {
+            return CWE_NIST_MAPPING.nistFilter([value], DEFAULT_NIST_TAG);
+          }
+        }
+      }
+    },
+    source_location: {
+      ref: {path: 'paths'}
+    },
+
```
Amndeep7, line 400: get rid of newline

Amndeep7, line 400: also make sure to run the linter (yarn run lint) which should autofix most of the issues and complain about the places where it can't fix them.  i would run it as a last step before pushing up to the repo so please don't use it as an easy first win type of commit cause you're probably gonna have to do it again at the end anyways lol

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+    results: [
+      {
+        path: 'components',
+        status: ExecJSON.ControlResultStatus.Failed,
+        ///code vs message, make sure code_desc is a description
```
Amndeep7, line 405: remove unnecessary comment

libs/hdf-converters/src/veracode-mapper.ts
Amndeep7: Should also provide an extended comment here being explicit about what the structure looks like before, the problem it's causing, and what we're turning the structure into.  Reason for this is that `pathTransform` should be an exceptional situation where you really don't have any other way around it.  Due to it being exceptional, it's important to have documentation explaining what's going on and why.

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+export class VeracodeMapper extends BaseConverter {
+  originalData: unknown;
+  defaultMapping(
+    withRaw = false
+  ): MappedTransform<ExecJSON.Execution & {passthrough: unknown}, ILookupPath> {
+    return {
+      platform: {
+        name: 'Heimdall Tools',
+        release: HeimdallToolsVersion,
+        target_id: ''
```
Amndeep7, line 463: please remove unless we can find a good mapping in the original data, which i feel like we probably could if we conjoined a few fields like detailedreport.modules[0...x].name

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+      },
+      passthrough: {
+        auxiliarry_data: [
```
Amndeep7, line 466: typo

Amndeep7 (changes requested):

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
+import {ExecJSON} from 'inspecjs';
+import _, {isArray} from 'lodash';
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+          string,
+          unknown
+        >
+      );
+    }
+  }
+  return vulnarr;
+}
+
+function componentlistcreate(input:unknown){
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+    if (!Array.isArray(_.get(input[0], STATIC_FLAWS))) {
+      flawArr.push(_.get(input[0], STATIC_FLAWS) as string);
+    } else {
+      flawArr.push(...(_.get(input[0], STATIC_FLAWS) as string[]));
+    }
+  }
+  return flawArr.map((value) => _.get(value, 'sourcefile')).join('\n');
+}
+
+function vulnarrcreate(componentlist: Record<string, unknown>[]): Record<string, unknown>[] {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+        >
+      );
+    }
+  }
+  return vulnarr;
+}
+
+function componentlistcreate(input:unknown){
+  const componentlist: Record<string, unknown>[] =[]
+  if (isArray(_.get(input, `component`))) {
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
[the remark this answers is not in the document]
jweissm: fixed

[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+}
+
+function componentTransform(input: unknown) {
+  const componentlist: Record<string, unknown>[] = componentlistcreate(input);
+
+  const vulnarr: Record<string, unknown>[] = vulnarrcreate(componentlist);
+
+  vulnarr.forEach((vuln) => {
+    const components = [];
+    let location = '';
```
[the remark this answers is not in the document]
jweissm: dont believe so, since components set to [] at the beginning of each iteration no?

libs/hdf-converters/src/veracode-mapper.ts
jweissm: changed cast to Record<string, unknown>

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: removed platform

libs/hdf-converters/src/veracode-mapper.ts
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+    flawDesc = `component_id: ${_.get(input, 'component_id')};`;
+    flawDesc += categories
+      .map((value: string) => {
+        if (_.has(input, value)) {
+          return `${value}: ${_.get(input, value)}`;
+        } else {
+          return '';
+        }
+      })
+      .join('\n');
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+
+function formatSCACodeDesc(input: Record<string, unknown>): string {
+  let flawDesc = '';
+  const categories = [
+    'sha1',
+    'max_cvss_score',
+    'version',
+    'library',
+    'vendor',
+    'descritpion',
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/veracode-mapper.ts
```diff
@@ -0,0 +1,519 @@
...
+  if (Array.isArray(input)) {
+    for (const value of input) {
+      if (!Array.isArray(_.get(value, STATIC_FLAWS))) {
+        flawArr.push(_.get(value, STATIC_FLAWS) as string);
+      } else {
+        flawArr.push(...(_.get(value, STATIC_FLAWS) as string[]));
+      }
+    }
+  } else {
+    input = [input];
```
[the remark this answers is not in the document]
jweissm: fixed

Amndeep7 (approved):

PR #3095, opened 2022-05-31
added branch to Sonarqube converter

blake-burton:
Sonarqube API assume default branch when no branch is passed in. This means we are not able to pull the issues related to a Release branch, which can come in handy while creating a ATO package for software in a release branch. also added PR as an option as well, which will come in handy when applying compliance thresholds to PR.

Note this tag is only applicable if SonarQube server is Developer addition or greater 
https://next.sonarqube.com/sonarqube/web_api/api/issues#:~:text=Branch%20key.%20Not%20available%20in%20the%20community%20edition.


Amndeep7:
@blake-burton I think one thing that might help is not duplicating the full describe block.  You can just put more `it` blocks inside the first describe block.  If this is still having errors, then we can give it another look.

Also, it looks like I have to manually approve the workflows every time, which is pretty frustrating, but dunno if there's anything we can do about that until you get this PR merged in and you're no longer a 'first-time contributor'.  Anyways, seems like the linter caught some issues as well, so would appreciate you fixing those!

Amndeep7:
@blake-burton still seems like some linting stuff left.  Try running `yarn lint` and it should fix most near all of it.

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -115,14 +125,18 @@ export class SonarQubeResults {
...
       await axios
         .get<IssueData>(`${this.sonarQubeHost}/api/issues/search`, {
-          auth: {username: this.userToken, password: ''},
+          auth: { username: this.userToken, password: '' },
           params: {
             componentKeys: this.projectId,
             types: 'VULNERABILITY',
-            p: page
+            p: page,
+            ...(this.branchName ? { branch: this.branchName } : {}),
```
Amndeep7, line 133: instead of doing the ternary operator you can use the AND operator (&&)

a small article on it: https://medium.com/@slamflipstrom/conditional-object-properties-using-spread-in-javascript-714e0a12f496

i'd prefer to use this form instead of the ternary operator so as to match the other mappers

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -192,8 +210,13 @@ export class SonarQubeResults {
 }
 
 function createSonarqubeMappings(
-  projectName: string
+  projectName: string,
+  branchName?: string,
+  pullRequestID?: string
 ): MappedTransform<ExecJSON.Execution, ILookupPath> {
+  const scanDescriptionModifier = (
```
Amndeep7, line 217: getting ahead of the linter, it usually prefers single quotes and paired indentation levels for things like braces and parentheses

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -255,9 +278,18 @@ function createSonarqubeMappings(
...
+    pullRequestID?: string
+  ) {
     super(issuesJSON as Record<string, any>);
-    this.setMappings(createSonarqubeMappings(projectName));
+    this.setMappings(createSonarqubeMappings(
+      projectName, branchName, pullRequestID)
+    );
   }
 
   setMappings(
```
Amndeep7, line 295: Can you remove this function?  We've been removing it when it's unnecessary.  All it's doing here is calling the one in the BaseConverter, at which point we might as well use the BaseConverter's implementation directly.

https://github.com/mitre/heimdall2/blob/b5bac50274b935ee349a3018a8d328bcbb24e2a8/libs/hdf-converters/src/base-converter.ts#L147=

Amndeep7 (changes requested):

Amndeep7 (approved):

Amndeep7:
@blake-burton this'll merge if everything still passes all the ci/cd checks after merging in master one last time, which it most probably should.  once we push out a new release of heimdall/hdf-converters, we'll be able to merge in the matching pr on the saf cli and your new functionality will be completely available to the world.  congrats on your first contribution!

PR #3105, opened 2022-06-01
Allow Multiple File Uploads to /evaluations

camdenmoors:


apps/backend/src/evaluations/evaluations.controller.spec.ts
```diff
@@ -240,10 +240,15 @@ describe('EvaluationsController', () => {
     it('should allow a user to create an evaluation', async () => {
```
Amndeep7, line 240: we should add another it block to validate multiple file upload

Amndeep7 (changes requested):

PR #3141, opened 2022-06-08
Updated fast-xml-parser to v4

charleshu-8:


Amndeep7:
@charleshu-8 what was the blocker for this pr again?

charleshu-8:
> @charleshu-8 what was the blocker for this pr again?

I believe the `parse` function was changed, so the output no longer aligns with what we expected for originally.

PR #3151, opened 2022-06-13
Added CCI tags for Twistlock vulnerability results

charleshu-8:


camdenmoors (approved):

camdenmoors:
> @camdenmoors We good to merge?

As long as these are the CCIs that @ejaronne wants then yes we should be good. You should be able to add the `ready-to-merge` tag if so

PR #3170, opened 2022-06-15
Add Default CCIs for Relevant NIST Controls

camdenmoors:


em-c-rod (commented):
From a high level, it looks like these changes are generally to support a new tag field for CCIs. Is that option new? Or just wasn't in the testing before? Anything else to pay close attention to?

camdenmoors:
> From a high level, it looks like these changes are generally to support a new tag field for CCIs. Is that option new? Or just wasn't in the testing before? 
This is a new feature to support conversion to POA&M

> Anything else to pay close attention to?
That the data should be included on import and matches that of the included lookup table

libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -67,7 +68,7 @@ export class TwistlockMapper extends BaseConverter {
             key: 'id',
             tags: {
               nist: ['SI-2', 'RA-5'],
-              cci: ['CCI-002605', 'CCI-001643'],
+              cci: getCCIsForNISTTags(['SI-2', 'RA-5']),
```
Amndeep7, line 71: Don't we have a constant somewhere that defines our default nist tags?  Use that.

libs/hdf-converters/src/utils/global.ts
```diff
@@ -50,3 +54,16 @@ export function getDescription(
 
   return found;
 }
+
+export function getCCIsForNISTTags(nistTags: string[]): string[] {
+  const cciTags: string[] = [];
+  nistTags.forEach((tag) => {
+    const baseTag = /\w\w-\d/g.exec(tag);
+    if (baseTag) {
+      if (baseTag[0] in NistCciMappingData) {
```
Amndeep7, line 63: combine the two conditionals with && in order to get rid of pointless nesting

Amndeep7 (changes requested):

Amndeep7:
@camdenmoors bump

Amndeep7:
@camdenmoors can you fix the merge conflict?  Also the veracode mapper was merged, so can you add the cci stuff there as well?

PR #3182, opened 2022-06-16
Snyk api

Amndeep7:


Amndeep7:
There's some conflicts that i gotta deal with and it'd probably be a good idea to rebase the changes on top of master since the branch is kinda old, but this should be good to at least get the review started.  I'll make the cli integration pr later on today.

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,142 +39,467 @@ function nistTag(identifiers: unknown[]): string[] {
...
+                        paths = Object.values(
+                          _.pick(
+                            _.get(this.unconvertedData, 'dependencyChains'),
+                            (_.get(input, 'pkgVersions') as string[]).map(
+                              (version) =>
+                                `${_.get(input, 'pkgName')}@${version}`
+                            )
+                          )
+                        );
+                      }
```
charleshu-8, line 458: Results `code_desc` feels lacklustre.

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,142 +39,467 @@ function nistTag(identifiers: unknown[]): string[] {
...
+            ? {
+                raw: _.omit(data, [
+                  'organization',
+                  'project',
+                  'dependencyChains'
+                ])
+              }
+            : {
+                snyk_metadata: _.omit(data, ['vulnerabilities'])
+              }
```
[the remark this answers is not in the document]
Amndeep7: Raw is only gonna contain the original data (as converted into json) or the api calls.  The things that I am omitting there are additional attributes that I added to the data object based off of processing that I did.  

I'm renaming "metadata" to "unmapped_data" for lack of a better name atm, and that will contain the stuff that isn't mapped into hdf.  in the case of a file input, that means that it's like a handful of attributes or less since everything else is in the vulnerabilities array, which itself is completely mapped due to dumping each issue into the code attribute.

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,142 +39,467 @@ function nistTag(identifiers: unknown[]): string[] {
   );
```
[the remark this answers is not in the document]
Amndeep7: first question, probably not for this specific mapper

second question, sounds reasonable enough to me

however, if we put the implementation of 'parseidentifier' into 'nisttag' right now, does the premise still hold true for the other mappers using the same implementation?  i think we should make this critique its own pr to address this deduplication.

libs/hdf-converters/src/snyk-mapper.ts
Amndeep7: I dunno what you mean by this lmao.  please give actionable advice in the future.

libs/hdf-converters/src/snyk-mapper.ts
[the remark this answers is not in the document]
charleshu-8: Alright. Also, got a confirmation that the new naming scheme for unused data will be `other_source_tool_data` instead of `unmapped_data`.

libs/hdf-converters/src/snyk-mapper.ts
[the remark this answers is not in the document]
charleshu-8: Realistically then, what we could do is keep the parseIdentifier and nistTag functions separate and move them both to BaseConverter. Anyways, [I wrote it up as an issue](https://github.com/mitre/heimdall2/issues/3237) so someone can address it later down the line.

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,132 +40,580 @@ function nistTag(identifiers: unknown[]): string[] {
...
+            }
+          ],
+          sha256: ''
         }
+      ],
+      passthrough: {
+        transformer: (data: Record<string, unknown>): Record<string, unknown> =>
+          isApi?.apiVersion === 1
+            ? {
+                snyk_unmapped_data: _.omit(data, [
```
charleshu-8, line 590: Rename `snyk_unmapped_data` to `other_source_tool_data`.

libs/hdf-converters/src/base-converter.ts
```diff
@@ -8,8 +8,8 @@ import Papa from 'papaparse';
 export interface ILookupPath {
   shortcircuit?: boolean;
   path?: string | string[];
-  transformer?: (value: any) => unknown;
-  arrayTransformer?: (value: unknown[], file: any) => unknown[];
+  transformer?: (internalData: any) => unknown;
+  arrayTransformer?: (data: unknown[], unconvertedData: any) => unknown[];
```
camdenmoors, line 12: Can we be consistent with these names

libs/hdf-converters/src/base-converter.ts
```diff
@@ -137,12 +142,15 @@ function collapseDuplicates<T extends object>(
 }
 
 export class BaseConverter {
-  data: Record<string, unknown>;
+  unconvertedData: Record<string, unknown>;
```
camdenmoors, line 145: Might want to call this source data, seems more professional

libs/hdf-converters/src/base-converter.ts
```diff
@@ -171,89 +179,97 @@ export class BaseConverter {
     ) as Record<keyof T, V>;
   }
   convertInternal<T>(
-    file: Record<string, unknown>,
-    fields: T
+    data: Record<string, unknown>,
```
camdenmoors, line 182: Whatever we choose to name this please keep it consistent

libs/hdf-converters/src/base-converter.ts
```diff
@@ -171,89 +179,97 @@ export class BaseConverter {
...
+      _.has(mappings, 'shortcircuit') &&
+      _.isBoolean(_.get(mappings, 'shortcircuit')) &&
+      _.get(mappings, 'shortcircuit');
     if (isShortcircuiting) {
-      return _.omit(fields as object, 'shortcircuit') as MappedReform<
+      return _.omit(mappings as object, 'shortcircuit') as MappedReform<
         T,
         ILookupPath
       >;
     }
```
camdenmoors, line 195: This whole block can be

```
if (_.get(mappings, 'shortcircuit' === true) {
   ...
}
```

libs/hdf-converters/src/base-converter.ts
```diff
@@ -171,89 +179,97 @@ export class BaseConverter {
...
+    const result = this.objectMap(mappings, (mapping: ObjectEntryValue<T>) =>
+      this.evaluate(data, mapping)
     );
     return result as MappedReform<T, ILookupPath>;
   }
 
   evaluate<T>(
-    file: Record<string, unknown>,
-    v: T | Array<T>
+    data: unknown,
```
camdenmoors, line 204: Please use a consistent name

libs/hdf-converters/src/base-converter.ts
```diff
@@ -171,89 +179,97 @@ export class BaseConverter {
...
+        | (T & ILookupPath)
+        | Array<T & ILookupPath>;
     }
 
-    if (Array.isArray(pathV)) {
-      return hasTransformer
-        ? (transformer(pathV) as T[])
-        : this.handleArray(file, pathV);
+    const hasPath = _.isObject(mapping) && _.has(mapping, 'path');
+    let internalData = data;
```
camdenmoors, line 235: Another name, also what's the purpose of duplicating this object?

libs/hdf-converters/src/base-converter.ts
```diff
@@ -286,8 +305,11 @@ export class BaseConverter {
         const key = lookupPath.key;
         const arrayTransformer = lookupPath.arrayTransformer?.bind(this);
         const transformer = lookupPath.transformer?.bind(this);
-        if (this.hasPath(file, path)) {
-          const pathVal = this.handlePath(file, path);
+        if (this.hasPath(file as Record<string, unknown>, path)) {
```
camdenmoors, line 308: Add a guard expecting file to be of object instead of asserting the type

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,132 +40,580 @@ function nistTag(identifiers: unknown[]): string[] {
...
-    const results: ExecJSON.Execution[] = [];
-    if (Array.isArray(this.data)) {
-      this.data.forEach((element) => {
-        const entry = new SnykMapper(element);
-        if (this.customMapping !== undefined) {
-          entry.setMappings(this.customMapping);
+  // api constructor
+  static SnykAPI(
+    apiToken: string,
+    apiVersion: 1 = 1,
```
camdenmoors, line 80:
```suggestion
    apiVersion: number = 1,
```

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,132 +40,580 @@ function nistTag(identifiers: unknown[]): string[] {
...
+          groups: [],
+          status: 'loaded',
+          controls: [
+            {
+              path: isApi?.apiVersion === 1 ? 'issues' : 'vulnerabilities',
+              key: 'id',
+              tags: {
+                nist: {
+                  path: `${
+                    isApi?.apiVersion === 1 ? SNYK_API_ISSUE_DATA_PATH : ''
```
camdenmoors, line 418: Make this a function or a variable instead of calculating every time

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,132 +40,580 @@ function nistTag(identifiers: unknown[]): string[] {
...
+                }
+              },
+              descriptions: [
+                {
+                  label: 'fix',
+                  data: {
+                    path: `${
+                      isApi?.apiVersion === 1 ? SNYK_API_ISSUE_DATA_PATH : ''
+                    }description`,
+                    transformer: (description: string): string => {
```
camdenmoors, line 485: Add a guard saying if descriptions is undefined, not sure if this will ever happen but better safe than sorry

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,132 +40,580 @@ function nistTag(identifiers: unknown[]): string[] {
...
-          data: Record<string, unknown>
-        ): Record<string, unknown> => {
-          return _.omit(data, ['vulnerabilities']);
+                {
+                  label: 'check',
+                  data: {
+                    path: `${
+                      isApi?.apiVersion === 1 ? SNYK_API_ISSUE_DATA_PATH : ''
+                    }description`,
+                    transformer: (description: string): string => {
```
camdenmoors, line 502: Same as fix

libs/hdf-converters/test/mappers/forward/snyk_mapper.spec.ts
```diff
@@ -25,16 +33,23 @@ describe('snyk_mapper', () => {
...
     );
   });
-  it('Successfully converts Snyk cli targeted at a remote/online repository data', () => {
-    const mapper = new SnykMapper(
-      JSON.parse(
-        fs.readFileSync(
-          'sample_jsons/snyk_mapper/sample_input_report/nodejs-goof-remote.json',
-          {encoding: 'utf-8'}
-        )
+  it('Successfully converts Snyk cli targeted at a remote/online repository data', async () => {
```
camdenmoors, line 36: Can we make this a mock API? See `heimdall2/test/support/server`

camdenmoors (changes requested):

libs/hdf-converters/src/snyk-mapper.ts
```diff
@@ -37,132 +40,580 @@ function nistTag(identifiers: unknown[]): string[] {
...
+          isApi?.apiVersion === 1
+            ? {
+                snyk_unmapped_data: _.omit(data, [
+                  'issues',
+                  'organization',
+                  'project',
+                  'dependencyChains'
+                ]),
+                ...(withRaw && {
+                  raw: _.omit(data, [
```
charleshu-8, line 597: From my understanding, `raw` will not omit any data from it.

charleshu-8 (changes requested):

charleshu-8 (changes requested):

libs/hdf-converters/test/mappers/forward/snyk_mapper.spec.ts
```diff
@@ -1,18 +1,26 @@
...
     );
-    expect(omitVersions(mapper.toHdf())).toEqual(
+
+    /*
+    fs.writeFileSync(
+      'sample_jsons/snyk_mapper/nodejs-goof-local-hdf.json',
+      JSON.stringify(await mapper.toHdf(), null, 2),
+      {encoding: 'utf-8'}
+    );
+    */
```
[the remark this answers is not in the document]
Amndeep7: I think we're moving forward with keeping the writefilesyncs but commented out in the test code to make it easier to go back and generate new sample hdfs on the fly.  @camdenmoors can you confirm?

libs/hdf-converters/src/snyk-mapper.ts
Amndeep7: based off discussion from earlier today, i'm renaming it to that auxiliary data thing unless aaron prefers something else

libs/hdf-converters/src/base-converter.ts
Amndeep7: i mean nominally this is all source data cause we're not generating data ourselves (except in those cases where we are where we supply default nist tags, etc).  i find 'unconverted' to be a more accurate descriptor.  we can be less concerned about things seeming "professional" since end users are not gonna see that term in any hdf files.

libs/hdf-converters/src/snyk-mapper.ts
Amndeep7: This is intentional here.  When we add support for api version 3 it'll be `apiVersion: 1|3 = 1`.  I'm more less using it as an enum.

libs/hdf-converters/test/mappers/forward/snyk_mapper.spec.ts
Amndeep7: the two tests currently in there are both for the file based inputs (the results are different if the cli was pointed at a local copy vs a remote).  i can add a test for the api mapper using a mock

PR #3200, opened 2022-06-21
Updating Sarif mapper formatting

charleshu-8:


Rlin232 (approved):

libs/hdf-converters/src/sarif-mapper.ts
```diff
@@ -122,25 +118,49 @@ export class SarifMapper extends BaseConverter {
...
+        transformer: (
+          data: Record<string, unknown>
+        ): Record<string, unknown> | string => {
+          return this.passRaw ? data : '';
+        }
+      }
+    }
   };
-  constructor(sarifJson: string) {
+  constructor(sarifJson: string, passRaw = false) {
```
Amndeep7, line 162: i believe we were wanting to call it "withRaw" not "passRaw" to emphasize that the raw stuff would be coming with the rest of the hdf files.  consequently the saf cli would have an option for all mappers for "--with-raw" and then for the api based mappers specifically would have an option for "--raw" to just give the api data only.

libs/hdf-converters/sample_jsons/sarif_mapper/sarif-hdf.json
```diff
@@ -1,981 +1,1499 @@
...
-                            "run_time": 0,
-                            "start_time": ""
-                        }
-                    ]
-                }
-            ],
-            "sha256": "407f6e03f5e117392a5629be0d3fa1af036b4bd09f33a5e3159d9faf1f7381e8"
-        }
-    ]
+	"platform": {
```
camdenmoors, line 2: Can we be consistent with our indentation level? I believe `JSON.stringify(object, null, 2)` is what we normally use

libs/hdf-converters/src/sarif-mapper.ts
```diff
@@ -122,25 +118,40 @@ export class SarifMapper extends BaseConverter {
...
           }
         ],
         sha256: ''
       }
-    ]
+    ],
+    passthrough: {
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
+        const auxData = _.omit(data, ['version']);
+        auxData.runs = auxData.runs.map((run: any) => _.omit(run, ['results']));
```
Amndeep7, line 145: don't use any, esp since we're assuming that `run` is an object/Record<string, unknown> anyways

Amndeep7 (changes requested):

libs/hdf-converters/src/sarif-mapper.ts
charleshu-8: From what I've experienced, typing instances as `unknown` prevents accessing object attributes (which also seem to be the case with this function as well). Is there a workaround or am I just not looking at it right?

Rlin232:
Assuming tests are failing because they haven't been updated with the correct formatting - but the code itself looks good to me!

libs/hdf-converters/src/sarif-mapper.ts
Amndeep7: The primary issue is that `any` is propagated and pollutes the rest of the type system so it's hard to do things safely - any can be anything and so is treated as validly being everything.  If you use unknown instead, you're forced to manually make casts and explicitly specify the "contract" that a given thing is of a specific type.  It is definitely not perfect (everything being precisely typed), but it is significantly better than the alternative of there essentially not being a type at all.

libs/hdf-converters/src/sarif-mapper.ts
```diff
@@ -140,11 +140,21 @@ export class SarifMapper extends BaseConverter {
       }
     ],
     passthrough: {
-      transformer: (data: Record<string, any>): Record<string, unknown> => {
-        const auxData = _.omit(data, ['version']);
-        auxData.runs = auxData.runs.map((run: any) => _.omit(run, ['results']));
+      transformer: (data: Record<string, unknown>): Record<string, unknown> => {
+        let runsData = _.get(data, 'runs');
+        if (Array.isArray(runsData)) {
+          runsData = runsData.map((run: any) => _.omit(run, ['results']));
```
Amndeep7, line 146: still using any here

Amndeep7 (changes requested):

libs/hdf-converters/src/sarif-mapper.ts
Amndeep7:
```suggestion
          runsData = runsData.map((run: Record<string, unknown>) => _.omit(run, ['results']));
```

libs/hdf-converters/src/sarif-mapper.ts
```diff
@@ -122,25 +118,52 @@ export class SarifMapper extends BaseConverter {
               }
             },
             impact: {path: 'level', transformer: impactMapping},
-            code: '',
+            code: {
+              transformer: (vulnerability: Record<string, unknown>): string => {
+                return JSON.stringify(vulnerability, null, 2);
+              }
```
camdenmoors, line 124:
```suggestion
              transformer: (vulnerability: Record<string, unknown>): string => JSON.stringify(vulnerability, null, 2);

```

libs/hdf-converters/src/sarif-mapper.ts
```diff
@@ -122,25 +118,52 @@ export class SarifMapper extends BaseConverter {
...
                 start_time: ''
               }
             ]
           }
         ],
         sha256: ''
       }
-    ]
+    ],
+    passthrough: {
```
camdenmoors, line 142: Passthrough should likely go at the top with the rest of the header data, same applies to the other PRs.

charleshu-8: Since passthrough is essentially our misc. data, wouldn't placing it at the top distract away from our main focus of the files (e.g., controls & their results?). I'm of the opinion that placing it alongside the headers would make it less human readable, especially if the user just wants to find vulnerabilities etc. and don't care about whatever extraneous information we put in passthrough.

Amndeep7 (approved):

PR #3203, opened 2022-06-21
Updating Zap mapper formatting

charleshu-8:


Rlin232 (approved):

libs/hdf-converters/src/zap-mapper.ts
```diff
@@ -140,26 +133,42 @@ export class ZapMapper extends BaseConverter {
...
               }
             ]
           }
         ],
         sha256: ''
       }
-    ]
+    ],
+    passthrough: {
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
```
Rlin232, line 161: Avoid uses of "any" (I think this has been commented on other PR's already though)

libs/hdf-converters/src/zap-mapper.ts
```diff
@@ -140,26 +133,50 @@ export class ZapMapper extends BaseConverter {
...
       }
-    ]
+    ],
+    passthrough: {
+      transformer: (data: Record<string, unknown>): Record<string, unknown> => {
+        return {
+          auxiliary_data: [
+            {
+              name: 'OWASP ZAP',
+              data: {
```
Amndeep7, line 166: I think you could probably do ```data: _.pick(data, ['site.@port', 'site.@ssl'])```

See:

```js
const a = {a: {b: 5, c: 2, d: 3}, e: 1};
console.log(_.pick(a, ['a.b', 'a.c']));
```

which generates

```js
{
  a: {
    b: 5,
    c: 2
  }
}
```

Amndeep7 (changes requested):

libs/hdf-converters/src/zap-mapper.ts
```diff
@@ -140,26 +133,45 @@ export class ZapMapper extends BaseConverter {
               riskdesc: {path: 'riskdesc'},
               check: {transformer: checkText}
```
camdenmoors, line 134: Check and Fix are normally a description instead of a tag now

libs/hdf-converters/src/zap-mapper.ts
```diff
@@ -140,26 +133,45 @@ export class ZapMapper extends BaseConverter {
...
             source_location: {},
-            code: '',
+            title: {path: 'name'},
+            id: {path: 'pluginid'},
+            desc: {path: 'desc', transformer: parseHtml},
+            impact: {path: 'riskcode', transformer: impactMapping},
+            code: {
+              transformer: (vulnerability: Record<string, unknown>): string => {
+                return JSON.stringify(vulnerability, null, 2);
+              }
```
camdenmoors, line 145:
```suggestion
              transformer: (vulnerability: Record<string, unknown>): string => JSON.stringify(vulnerability, null, 2);
```

Amndeep7:
How does that check text look in heimdall?  It seems like it's HTML encoded.  If that's the case, maybe we can parse that too so that it looks pretty but maybe it already looks nice there?  Deploypreview is currently borked cause of the netlify injection I think so need to either push a fake commit or deploy from heroku.

charleshu-8:
> How does that check text look in heimdall? It seems like it's HTML encoded. If that's the case, maybe we can parse that too so that it looks pretty but maybe it already looks nice there? Deploypreview is currently borked cause of the netlify injection I think so need to either push a fake commit or deploy from heroku.

Text itself comes out okay. Only problem I see are the new lines.

Screenshot (13)
Screenshot (14)
Screenshot (15)

Amndeep7 (approved):

PR #3226, opened 2022-06-27
Multi os pack script

Rlin232:


libs/hdf-converters/package.json
```diff
@@ -11,9 +11,15 @@
     "main": "lib/index.js"
   },
   "scripts": {
-    "prepack": "yarn build && cp package.json package.json.orig && cat package.json.orig | jq '.main = (.publishConfig.main)' > package.json",
-    "postpack": "mv package.json.orig package.json",
-    "build": "tsc -p ./tsconfig.build.json && cp -R ./data ./lib",
+    "prepack": "run-script-os",
+    "prepack:darwin:linux": "yarn build && cp package.json package.json.orig && cat package.json.orig | jq '.main = (.publishConfig.main)' > package.json",
+    "prepack:win32": "yarn build && copy package.json package.json.orig && copy package.release.json package.json",
```
Amndeep7, line 16: Sorry that I wasn't able to meet with you - let's meet on Friday for sure.  I don't understand what's going on with the package.json moving/overwriting.  I am also particularly concerned with this package.release.json file that seems to have been introduced here, but there is no automated flow for making sure it's kept up to date as dependencies are upped.

Amndeep7 (changes requested):

camdenmoors (approved):
New `prepack.js` script looks good to me, after updating build:win32 this should be good to go

PR #3232, opened 2022-06-28
Reformatted Twistlock & Jfrog to include new passthrough format

charleshu-8:


libs/hdf-converters/src/jfrog-xray-mapper.ts
```diff
@@ -161,16 +163,16 @@ export class JfrogXrayMapper extends BaseConverter {
       }
     ],
     passthrough: {
-      raw: {
-        transformer: (
-          data: Record<string, unknown>
-        ): Record<string, unknown> => {
-          return data;
-        }
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
```
Amndeep7, line 166: don't use any since it disables the type system

libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -118,16 +120,24 @@ export class TwistlockMapper extends BaseConverter {
...
     ],
     passthrough: {
-      raw: {
-        transformer: (
-          data: Record<string, unknown>
-        ): Record<string, unknown> => {
-          return data;
-        }
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
+        const auxData = data;
```
Amndeep7, line 124: don't overwrite parameters like this if we don't need to

just do auxil_d: [{name: 'twistlock', data: _.omit(data.results[0], [blah])}]

Amndeep7 (changes requested):

libs/hdf-converters/src/jfrog-xray-mapper.ts
Rlin232: ditto on the comment about any's

Rlin232 (dismissed):
The actual formatting looks good, but there's just some potential typescript annoyances that should be addressed (which Amndeep already mentioned above)

libs/hdf-converters/src/jfrog-xray-mapper.ts
Amndeep7: please see my response there

libs/hdf-converters/src/twistlock-mapper.ts
Amndeep7: Oh I see what you mean.  The thing you need to keep in mind is that doing the parameter overwriting like we're doing here with the assignment occurring to an internal structure of the object is that the changes will propagate due to js being a pass reference by value type of language.  This is a really solid stackoverflow answer covering this with a nice example: https://stackoverflow.com/a/3638034/645647.  Ways to work around this are to make 'auxData' a full and proper clone or to manually specify the missing data in auxiliary_data which could probably be eased through judicious use of the spread operator like so:

```js
const a = {a: 5, b: [{c: 2, d: 4}]};
console.log({...a, b: _.omit(a.b[0], ['c'])});
```

libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -118,16 +120,44 @@ export class TwistlockMapper extends BaseConverter {
...
-        ): Record<string, unknown> => {
-          return data;
-        }
+      transformer: (data: Record<string, unknown>): Record<string, unknown> => {
+        return {
+          auxiliary_data: [
+            {
+              name: 'Twistlock',
+              data: {
+                results: [
```
Amndeep7, line 129: Please see my response in the earlier review (idk if it'll properly show up so i'mma copy paste the body here too).

----

Oh I see what you mean. The thing you need to keep in mind is that doing the parameter overwriting like we're doing here with the assignment occurring to an internal structure of the object is that the changes will propagate due to js being a pass reference by value type of language. This is a really solid stackoverflow answer covering this with a nice example: [stackoverflow.com/a/3638034/645647](https://stackoverflow.com/a/3638034/645647). Ways to work around this are to make 'auxData' a full and proper clone or to manually specify the missing data in auxiliary_data which could probably be eased through judicious use of the spread operator like so:

```js
const a = {a: 5, b: [{c: 2, d: 4}]};
console.log({...a, b: [_.omit(a.b[0], ['c'])]});
```

which results in

```js
{
  a: 5,
  b: [{
  d: 4
}]
}
```

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #3244, opened 2022-06-30
Updating Nikto & Burp Suite & Scout Suite mappers

charleshu-8:


libs/hdf-converters/src/burpsuite-mapper.ts
```diff
@@ -125,24 +120,37 @@ export class BurpSuiteMapper extends BaseConverter {
...
-            source_location: {},
-            code: '',
+            impact: {
+              path: 'severity',
+              transformer: impactMapping(IMPACT_MAPPING)
+            },
+            code: {
+              transformer: (vulnerability: Record<string, unknown>): string => {
+                return JSON.stringify(vulnerability, null, 2);
+              }
```
camdenmoors, line 130:
```suggestion
              transformer: (vulnerability: Record<string, unknown>): string =>  JSON.stringify(vulnerability, null, 2);
```

camdenmoors (approved):

libs/hdf-converters/src/scoutsuite-mapper.ts
```diff
@@ -220,51 +219,73 @@ export class ScoutsuiteMapper extends BaseConverter {
...
+      transformer: (data: Record<string, unknown>): Record<string, unknown> => {
+        const auxData = _.omit(data, [
+          'account_id',
+          'environment',
+          'partition',
+          'provider_code',
+          'provider_name',
+          'services'
+        ]);
+        if (auxData instanceof Object) {
```
Amndeep7, line 277: what is the purpose of this conditional?

Amndeep7 (changes requested):

libs/hdf-converters/src/scoutsuite-mapper.ts
charleshu-8: Type guard so the `auxData` variable's data type will be narrowed from `unknown` to `Object` so I can access attributes.

charleshu-8: It did when I wrote it, but I checked again and it doesn't seem to be an issue now. Removed the type guard and the mapper still works just fine.

Amndeep7 (approved):

PR #3274, opened 2022-07-05
Updating XCCDF mapper formatting

charleshu-8:


Rlin232:
> 

Not sure what to do about the SonarCloud complaining about the commented-out code. It is indeed useful to have in there for the sake of testing, so I'm of the opinion that we leave it in and ignore the SonarCloud stuff.

Amndeep7:
> I'm of the opinion that we leave it in and ignore the SonarCloud stuff.

that's exactly what we're gonna do @Rlin232

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -213,9 +218,7 @@ export class XCCDFResultsMapper extends BaseConverter {
       }
     },
     version: HeimdallToolsVersion,
-    statistics: {
-      duration: 0
-    },
+    statistics: {},
     profiles: [
       {
         name: {path: ['cdf:Benchmark.id', 'Benchmark.id']},
```
Amndeep7, line 224: @camdenmoors mentioned something about how `cdf:` is some sorta prefix that we can tell the parser to ignore.  talk to him to see what's up with that so we can ideally simplify this mapper.

Amndeep7 (changes requested):
I would appreciate a chance to pair to see what the new output looks like just to confirm that it's all good.  Github don't wanna show diffs this large lol.

Amndeep7 (approved):

PR #3314, opened 2022-07-11
Used local tsc to account for use case

Rlin232:
Signed-off-by: Rlin232 <rlin@mitre.org>


Rlin232:
Yeah, it's needed for users who don't have TSC installed globally to be able to run the scripts

em-c-rod (approved):
works in testing, including on windows. :+1:

PR #3325, opened 2022-07-12
Adding new ASFF mapper special casings

charleshu-8:
Signed-off-by: charleshu-8 <chu@mitre.org>


charleshu-8:
HANDOFF: Special cases should be fully implemented; need to check if appearance in Heimdall is satisfactory

Amndeep7:
Ignore the failing hdf converter tests - I need to update the sample results based off of the changes made here and potentially conflicts from the merge still.

Question: how do you want the inspec results to look @ejaronne?  At the moment it matches ASFF's display, but do we potentially want to combine all the same v-number controls into the same control and have the findings become subtests instead?

ejaronne:
> Ignore the failing hdf converter tests - I need to update the sample results based off of the changes made here and potentially conflicts from the merge still.
> 
> Question: how do you want the inspec results to look @ejaronne? At the moment it matches ASFF's display, but do we potentially want to combine all the same v-number controls into the same control and have the findings become subtests instead?

For the Chef InSpec results, yes, group by original requirement number, note that each "subtest" is a separate ASFF finding such as:

```
"Id": "us-east-1/380079715235/i-0f97f85f3f9b8940d/V-79001-443",
"Id": "us-east-1/380079715235/i-0f97f85f3f9b8940d/V-79001-442",
"Id": "us-east-1/380079715235/i-0f97f85f3f9b8940d/V-79001-441",
"Id": "us-east-1/380079715235/i-0f97f85f3f9b8940d/V-79001-440",
```
Also, note that a getfindings pull may have a mix of more than one target, so each would create a separate HDF results file.

ejaronne:
@Amndeep7 do you need additional (sanitized) samples for these new casings?

Amndeep7:
@ejaronne new samples would be appreciated.  if you could concoct an example of a getfindings pull with multiple targets (but same underlying product, so like multiple inspecs or multiple inspectors), I'd appreciate it.

Amndeep7:
@ejaronne please validate (using the heroku deployment, not the netlify deploy preview!) if the inspec results look good to you

libs/hdf-converters/src/asff-mapper/asff-mapper.ts
```diff
@@ -26,7 +29,10 @@ const COMPLIANCE_STATUS = 'Compliance.Status';
 
 // Sometimes certain ASFF file types require massaging in order to generate good HDF files.  These are the supported special cases and a catchall 'Default'.  'Default' files and non-special cased methods for otherwise special cased files do the pre-defined default behaviors when generating the HDF file.
 export enum SpecialCasing {
+  CMSInSpec = 'CMS Chef InSpec',
```
em-c-rod, line 32: Are these ASFF files that come from CMS running InSpec?

libs/hdf-converters/src/asff-mapper/asff-mapper.ts
```diff
@@ -37,29 +43,22 @@ export enum SpecialCasing {
...
-      /^arn:[^:]+:securityhub:[^:]+:[^:]*:product\/aws\/firewall-manager$/
-    )
-  ) {
-    return SpecialCasing.FirewallManager;
-  } else if (
-    productArn.match(
-      /^arn:[^:]+:securityhub:[^:]+:[^:]*:product\/prowler\/prowler$/
-    )
+    _.get(finding, 'ProductName') === 'Default' &&
+    _.get(finding, 'GeneratorId') === 'cms.Chef Inspec'
```
em-c-rod, line 47: Also, it seems like it is written differently in different names, `cms.Chef Inspec`, `CMS Chef InSpec`, and `CMSInSpec` at least.

libs/hdf-converters/src/asff-mapper/asff-mapper.ts
```diff
@@ -37,29 +43,22 @@ export enum SpecialCasing {
...
-      /^arn:[^:]+:securityhub:[^:]+:[^:]*:product\/aws\/securityhub$/
+      /^arn:[^:]+:securityhub:[^:]+:[^:]*:product\/aws\/firewall-manager$/
     )
   ) {
-    return SpecialCasing.SecurityHub;
+    return SpecialCasing.FirewallManager;
   } else if (
     productArn.match(
-      /^arn:[^:]+:securityhub:[^:]+:[^:]*:product\/aquasecurity\/aquasecurity$/
+      /^arn:[^:]+:securityhub:[^:]+:[^:]*:product\/aws\/guardduty$/
```
em-c-rod, line 58: Just looking at consistency - firewall-manager is separated by a dash and guardduty isn't. Maybe that is correct, but just want to make sure it is intentional.

em-c-rod (commented):
Mostly curious at why the new case is called CMS Chef InSpec and the like. What makes that case what it is? Because it is running an overlay? Or running InSpec? or what? I'm wondering if there is a more accurate name for it.

libs/hdf-converters/src/asff-mapper/asff-mapper.ts
Amndeep7: Please see internal discussion

libs/hdf-converters/src/asff-mapper/asff-mapper.ts
Amndeep7: CMSInSpec = the enum's key which we use to help out the type system more or less
CMS Chef InSpec = the enum's value - only used on definition more or less.  we try to have it be as complete a name as possible for the developer's benefit if we ever need to look that tool up again or something.
cms.Chef Inspec = the string that was placed in the generator id field of the asff file that we're using to identify what kind of file it is and then apply the special processing.

libs/hdf-converters/src/asff-mapper/asff-mapper.ts
Amndeep7: It is intentional.  It all depends on how that application registered itself as a product with AWS SecHub.  Guardduty doesn't have a dash, firewall manager does.

ejaronne:
For Chef-Inspec, it still needs to append the instance or Resources-Id to the filename json:
image

Amndeep7:
Latest commit addresses Eugene's concern

Amndeep7:
@em-c-rod this is good for review again.  if you're available, we can walk through it monday morning so that way it can be merged in time?

georgedias (approved):
Add clarification comments for each exemption referencing the AWS syntax format (schema) https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-findings-format-syntax.html

PR #3390, opened 2022-07-29
Adding support for Twistlock code repo scans

charleshu-8:
Signed-off-by: charleshu-8 <chu@mitre.org>

### DIFF
--- a/libs/hdf-converters/src/twistlock-mapper.ts
+++ b/libs/hdf-converters/src/twistlock-mapper.ts
@@ -48,16 +48,7 @@ export class TwistlockMapper extends BaseConverter {
     platform: {
       name: 'Heimdall Tools',
       release: HeimdallToolsVersion,
-      target_id: {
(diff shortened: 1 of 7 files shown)


libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -21,6 +21,23 @@ const IMPACT_MAPPING: Map<string, number> = new Map([
   ['low', 0.3]
 ]);
 
+export class TwistlockResults {
+  data: Record<string, unknown>;
+  customMapping?: MappedTransform<ExecJSON.Execution, ILookupPath>;
```
Amndeep7, line 26: if you don't use this attribute, then don't have it here.  wish we didn't put in the custom mapping stuff in the first place cause i've only seen it used once or twice across all of our mappers.

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #3504, opened 2022-09-01
Veracodemapperfixes

jweissm:
Current mapper does not address all use cases,  changing it back to include licenses and be formatted closer to original xml is important for our team.

### DIFF
--- a/libs/hdf-converters/sample_jsons/veracode_mapper/veracode-hdf.json
+++ b/libs/hdf-converters/sample_jsons/veracode_mapper/veracode-hdf.json
@@ -1715,17 +1715,240 @@
           ]
         },
         {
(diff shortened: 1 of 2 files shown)


aaronlippold:
Would love to chat about this. What do you mean by 'closer to the XML'

jweissm:
> Would love to chat about this. What do you mean by 'closer to the XML'

the current mapper switchs some things around as we thought it would match hdf format a bit more closley. It changes it form being 
component: {
    cves:[]
}
to 
cve{
   components:[]
}
this mostly works, but some components are lost, as its not 1-1, and not all data migrates as well as I first thought

Amndeep7:
gonna close this one since #3524 was opened instead

PR #3656, opened 2022-10-15
Add Database SSL Support

camdenmoors:


apps/backend/config/app_config.ts
```diff
@@ -51,9 +51,54 @@ export default class AppConfig {
...
 
   getSSLConfig() {
+    let sslKey, sslCert, sslCA;
+
+    if (typeof this.get('DATABASE_SSL_KEY') === 'string') {
+      if (this.get('DATABASE_SSL_KEY')?.indexOf('-BEGIN') !== -1) {
+        sslKey = this.get('DATABASE_SSL_KEY')
+      } else {
+        // Verify file exists
+        if (fs.statSync(this.get('DATABASE_SSL_KEY')!).isFile()) {
```
Amndeep7, line 61: code scanner thing probably throwing a fit over the exclamation mark.  think we can get typescript to accept it if we swap to something like the following:

```suggestion
        if (this.get('DATABASE_SSL_KEY') && fs.statSync(this.get('DATABASE_SSL_KEY')).isFile()) {
```

this change will need to be repeated a few times.

i think it'll also help with the fs.readFileSync call on (current) line 62 since typescript will know that that call will work

apps/backend/config/app_config.ts
```diff
@@ -51,9 +51,54 @@ export default class AppConfig {
   }
 
   getSSLConfig() {
+    let sslKey, sslCert, sslCA;
```
Amndeep7, line 54: maybe we should gate looking at all of these values on if the DATABASE_SSL is true.  if the user has garbage text in there, i don't think it matters since we say in the comments of the env-example that we only look at them if the other variable is true.  use-case i'm envisioning is someone leaving a valid sslkey/cert/ca in place but changing database_ssl to false as they do some debugging.

apps/backend/config/app_config.ts
```diff
@@ -51,9 +51,54 @@ export default class AppConfig {
...
+        if (fs.statSync(this.get('DATABASE_SSL_CA')!).isFile()) {
+          sslCA = fs.readFileSync(this.get('DATABASE_SSL_CA')!);
+        } else {
+          throw new Error('SSL CA file does not exist');
+        }
+      }
+    }
+
+
     return Boolean(this.get('DATABASE_SSL'))
```
Amndeep7, line 96: dotenv parses all variables as strings.  the Boolean function treats any string aside from '' as truthy and will therefore return true even if the string is 'false'.  I think this ternary operator will always follow the path of returning the object unless the variable is left empty basically.

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #3657, opened 2022-10-17
only pull open vulnerabilities for sonarqube2hdf as closed vulnerabilities

blake-burton:
Issue
===

Sonarqube2hdf failing with bad request caused by pulling closed vulnerabilities. The source pulls **all** the vulnerabilities, then for each vulnerability pulls the source code associated with it.  Vulnerabilities which are closed  causes an error because the file it was in no longer exists. This is the cause of the  BAD REQUEST is requesting source that no longer exists.

Solution
======

Only pull open and reopened vulnerabilities, this will ensure that the files which the vulnerabilities exists will still exist.

### DIFF
--- a/libs/hdf-converters/src/sonarqube-mapper.ts
+++ b/libs/hdf-converters/src/sonarqube-mapper.ts
@@ -130,6 +130,7 @@ export class SonarQubeResults {
           params: {
             componentKeys: this.projectId,
             types: 'VULNERABILITY',
+            statuses: 'OPEN,REOPENED',
             p: page,
             ...(this.branchName && {branch: this.branchName}),
             ...(this.pullRequestID && {pullRequest: this.pullRequestID})
--- a/libs/hdf-converters/src/sonarqube-mapper.ts
+++ b/libs/hdf-converters/src/sonarqube-mapper.ts
@@ -130,7 +130,7 @@ export class SonarQubeResults {
           params: {
             componentKeys: this.projectId,
             types: 'VULNERABILITY',
-            statuses: 'OPEN,REOPENED',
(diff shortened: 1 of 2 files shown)


blake-burton:
@Amndeep7 @camdenmoors @charleshu-8 Would it be possible to squeeze in a review for this?

Amndeep7:
Thanks for the PR, I'll take a look soon

Amndeep7 (approved):

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -130,6 +130,7 @@ export class SonarQubeResults {
           params: {
             componentKeys: this.projectId,
             types: 'VULNERABILITY',
+            statuses: 'OPEN,REOPENED',
```
Amndeep7, line 133:
```
Statuses

After creation, issues flow through a lifecycle, taking one of the following statuses:

    Open - set by SonarQube on new issues
    Confirmed - set manually to indicate that the issue is valid
    Resolved - set manually to indicate that the next analysis should Close the issue
    Reopened - set automatically by SonarQube when a Resolved issue hasn't actually been corrected
    Closed - set automatically by SonarQube for automatically created issues.
```

Shouldn't we actually be looking at all of these statuses aside from 'Closed'?

Amndeep7 (changes requested):

libs/hdf-converters/src/sonarqube-mapper.ts
blake-burton: Good call! Confirmed and  use case i have that cause this was only closed issues.

Amndeep7 (approved):

PR #3685, opened 2022-10-26
Updates for programmatic API creation use case

em-c-rod:
These changes stemmed from needing to document how to create an API Key programmatically. In doing so, we found some updates to documentation and ease of use necessary.

For reference, the new wiki to instruct how to create the API Key programmatically is [here](https://github.com/mitre/heimdall2/wiki/Programmatic-API-Key-Generation)

- Edited the error message for the setup-docker-env script if the .env file already exists
- Updated setup-docker-env script to accept "y" then enter for API Key generation
- Added in changes from Camden's branch to allow the user to be referenced by email instead of just userId

### DIFF
--- a/setup-docker-env.sh
+++ b/setup-docker-env.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [ -f .env ]; then
-	echo ".env already exists, if you would like to regenerate your secrets, please delete this file and re-run the script. Re-running this script will cause the database password to be reset within the .env file (This will not reset it in the database). If this happens, you can connect to the database directly and reset the password. Alternatively, you can remove the 'data/' folder (which will delete all data)."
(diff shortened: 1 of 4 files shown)


setup-docker-env.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
 if [ -f .env ]; then
-	echo ".env already exists, if you would like to regenerate your secrets, please delete this file and re-run the script."
+	echo ".env already exists, if you would like to regenerate your secrets, please delete this file and re-run the script. Re-running this script will cause the database password to be reset within the .env file (This will not reset it in the database). If this happens, you can connect to the database directly and reset the password. Alternatively, you can remove the 'data/' folder (which will delete all data)."
```
Amndeep7, line 4:
```suggestion
	echo ".env already exists, if you would like to regenerate your secrets, please delete this file and re-run the script. WARNING: Re-running this script will cause the database password to be reset within the .env file, but the database will still be expecting the old password.  If this happens, you can 1) change DATABASE_PASSWORD in the .env file back to the old password, 2) connect to the database directly and reset the password to the newly generated one, or 3) remove the 'data/' folder (which will delete all data)."
```

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #3693, opened 2022-10-28
Clarify XCCDF Results import scope

ejaronne:


Amndeep7 (approved):

aaronlippold:
Is there anything we can add to the fingerprinting to know if the XCCDF-Results file came from a supported tool? Is there anything in the header that ids which client produced the file?

aaronlippold:
Users don't RTFM so if our loading could say - I know you tried to give me an XCCDF-Results file but it doesn't seem to be from OpenSCAP or SCC or ACAS etc

Amndeep7:
@aaronlippold maybe?  I would feel uncomfortable relying on it due to the low number of samples that we have but please see the following:

image

image

They're in two different locations and seem to be optionally supplied.  If we have other/older samples that we can also test with to see if the behavior stays the same over time, maybe it's possible?  The SCC one in particular I dunno if it's intentional or just a fluke for how Cole set up his vm / scc run.

Amndeep7:
@mergifyio refresh

PR #3701, opened 2022-10-31
Display legacy vulnerability id

HenryXiaoHX:


package.json
```diff
@@ -35,5 +35,8 @@
   },
   "devDependencies": {
     "dotenv-cli": "^6.0.0"
+  },
+  "engines": {
+    "node": "<17.0.0"
```
Amndeep7, line 40: Note for other reviewers is that this is here since Heimdall doesn't work on node v18 (already have an issue about this).  Heroku's buildpack uses 18 by default.  The version of the buildpack that uses 16 is being deprecated.  For awareness of others / to make the Heroku build work, we need to specify a compatible version of node.  This will be removed once we get it working on v18.

apps/frontend/src/components/cards/controltable/ControlTable.vue
```diff
@@ -57,11 +57,21 @@
...
+          <v-row class="pa-3">
+            <ColumnHeader
+              text="ID"
+              :sort="sortId"
+              @input="set_sort('id', $event)"
+            />
+            <v-tooltip bottom>
+              <template #activator="{on, attrs}">
+                <v-icon style="cursor: pointer" v-bind="attrs" x-small v-on="on"
+                  >mdi-help</v-icon
```
Amndeep7, line 69: Make the icon be in a circle and in-line with the text.  Use the 'Help (?)' button on the 'load from s3 bucket' page as an example.

apps/frontend/src/components/cards/controltable/ControlTable.vue
```diff
@@ -57,11 +57,21 @@
...
+                >
+              </template>
+              <span>ID <br />(Legacy ID) </span>
```
Amndeep7, line 72: @HackerShark should add some additional text here to help explain the concept of legacy ids, but don't let that block this pr's progress.  We can make a new pr with that additional text if necessary.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -71,6 +71,11 @@
     <template #id>
       <v-card-text class="pa-2 title font-weight-bold">
         {{ control.data.id }}
+        {{
```
Amndeep7, line 74: This only puts things on a new line by a fluke on non-widescreen (1440p and up) resolutions.  My suggestion is to put the control.data.id in a div and then the legacy tag stuff in a separate div.  Make that separate div have a v-if that does the checking for if the legacy tag exists.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -71,6 +71,11 @@
...
+          control.data.tags['legacy']
+            ? '(' + control.data.tags['legacy'][0] + ')'
```
Amndeep7, line 76: Can't assume that legacy will be an array of strings.  The types say that control.data.tags is a mapping between keys that are strings and anything.  Consequently, we have to explicitly typecheck that the value is either a string or an array whose first value is a string.  For anything else or if the legacy tag doesn't exist in the first place, then don't display anything.  If someone complains, we can loosen this restriction at that time.

Amndeep7 (changes requested):

package.json
Amndeep7: https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

apps/frontend/src/components/cards/controltable/ControlTable.vue
ejaronne: Like we discussed, people who've been using older STIGs will understand. Others won't care.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
ejaronne: As long as this means the (legacy ID) will be on a new line underneath thank you!

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
ejaronne: Good catch. So, check if an array, then check if the first or only value is a string, and if not convert it internally.

ejaronne:
Instead of showing the first item in the array, please find the legacy ID that begins with "V-" and only show that in parentheses. We realized unfortunately the STIG xccdf doesn't put the vulnID and ruleID in the same order, so the array won't always have the old vulnID ("V-#") as the first item.

apps/frontend/src/components/cards/controltable/ControlTable.vue
em-c-rod: Or perhaps make it the "info" (the about example) button because it doesn't make the user think that the id is questionable.

em-c-rod: There is another example of this for the loading screen in the top right corner.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -268,5 +268,17 @@ export default class ControlRowHeader extends mixins(HtmlSanitizeMixin) {
       return {label: cci, url: '', description: this.descriptionForTag(cci)};
     });
   }
+  showLegacy(control: any) {
+    let legacyTag = control.data.tags['legacy'];
+    if (legacyTag) {
+      if (Array.isArray(legacyTag)) {
+        return '(' + legacyTag.find((ele) => ele.indexOf('V-') == 0) + ')';
```
HenryXiaoHX, line 275: Is it guaranteed we can find legacy ID that begins with "V-" if the array is present?

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -263,5 +268,22 @@ export default class ControlRowHeader extends mixins(HtmlSanitizeMixin) {
       return {label: cci, url: '', description: this.descriptionForTag(cci)};
     });
   }
+  showLegacy(control: any) {
```
Amndeep7, line 271: @HenryXiaoHX for future reference, we want to avoid using 'any' as much as possible since it essentially disables the type system and lets anything pass.  Types are great in how much safer your code ends up being, but actually finding/writing down the correct type can be frustrating.  Using 'any' as an escape should only be done in the rarest of circumstances since it weakens the typing that you've applied in all of the rest of the application.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -263,5 +268,22 @@ export default class ControlRowHeader extends mixins(HtmlSanitizeMixin) {
...
+    const legacyTag = control.data.tags['legacy'];
+    if (legacyTag) {
```
Amndeep7, line 273: @HenryXiaoHX usually it's better to have 'guard clauses' with early outs so that you can avoid nesting conditionals and have 'else' statements that are way far away from the relevant logic esp when the 'else' stuff is just like 'return default case'.  what you can do instead is make your conditional the opposite (so in this case '!legacyTag'), do the relevant else content (early out with empty string), and now the rest of the function is not indented one level.

apps/frontend/src/components/cards/controltable/ControlRowHeader.vue
```diff
@@ -263,5 +268,22 @@ export default class ControlRowHeader extends mixins(HtmlSanitizeMixin) {
...
+      if (Array.isArray(legacyTag)) {
```
Amndeep7, line 274: @HenryXiaoHX so this paradigm of 'thing or array of things' paradigm shows up a bunch where you do the exact same thing to both, but have to index into the array first.  so what we can do is similar to what i said with the guard clause concept, but in this case we check if something is not an array, make it an array, and now all the logic is consolidated.  it ends up making the code more DRY (Don't Repeat Yourself), which is good cause if the logic needs to be updated for some reason, you only need to make that update the once instead of twice or more (which could potentially result in you missing one of the occasions that you need to do that update).

Amndeep7 (approved):

PR #3718, opened 2022-11-03
handle case where 'vulnerabilities' attribute is null instead of an array

Amndeep7:


em-c-rod (approved):
LGTM :+1:

PR #3850, opened 2022-12-13
Update node version to18

em-c-rod:


.github/workflows/inspecjs-tests.yml
```diff
@@ -2,42 +2,42 @@ name: Run InspecJS Tests
...
-        bundle install --gemfile libs/inspecjs/Gemfile
-        yarn inspecjs gen-types
-        yarn inspecjs validate-types
-
-     - name: Ensure InspecJS builds cleanly
-       run: yarn inspecjs build
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
```
Amndeep7, line 17: Ruby 2.7 is EOL.  We should bump this to 3.x

.node-version
Amndeep7, line 1: should be removed post-merge since we should force the version upgrades in our cicd deployment configurations within repo and within the app settings directly for netlify and heroku

apps/frontend/jest.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     '^d3$': '<rootDir>/tests/util/d3.js',
     '^@[/](.+)': '<rootDir>/src/$1',
     '^.+\\.(css)$': '<rootDir>/tests/util/cssTransform.js',
-    '^axios$': 'axios/dist/node/axios.cjs'
+    '^axios$': 'axios/dist/node/axios.cjs',
+    '^csv-stringify/sync': 'csv-stringify/dist/cjs/sync.cjs'
```
Amndeep7, line 20: is this just to force the commonjs version over the es modules version?

apps/backend/package.json
```diff
@@ -15,7 +15,7 @@
     "build": "nest build",
     "lint": "eslint \"{src,migrations,seeders,test}/**/*.ts\" --fix",
     "lint:ci": "eslint \"{src,migrations,seeders,test}/**/*.ts\" --max-warnings 0",
-    "sequelize-cli": "ts-node --esm node_modules/.bin/sequelize",
+    "sequelize-cli": "node --loader ts-node/esm node_modules/.bin/sequelize",
```
Amndeep7, line 18: what does this do?

apps/frontend/package.json
```diff
@@ -43,18 +43,18 @@
...
     "@types/node": "^20.1.0",
-    "@types/objects-to-csv": "^1.3.0",
     "@types/prismjs": "^1.16.1",
     "@types/sanitize-html": "^2.3.1",
     "@types/triple-beam": "^1.3.2",
     "@types/uuid": "^9.0.0",
     "@types/vuelidate": "^0.7.13",
     "@types/xml-js": "^1.0.0",
-    "@vue/cli": "^5.0.3",
+    "@types/yazl": "^2.4.2",
```
Amndeep7, line 52: yazl does not seem to be maintained any longer.  we should look into a replacement lib to handle zip files

but also why do we need it at all when we also apparently are using adm-zip which is also a zip library?

admittedly it seems like adm-zip is also probably not being maintained

apps/frontend/src/views/Compare.vue
```diff
@@ -376,7 +376,7 @@ export default class Compare extends Vue {
     this.files.forEach((file) => {
       if ('passthrough' in file.evaluation.data) {
         const passthroughData = _.get(file.evaluation.data, 'passthrough');
-        if (typeof passthroughData === 'object') {
+        if (passthroughData !== null && typeof passthroughData === 'object') {
```
Amndeep7, line 379: typeof === 'object' is basically useless since most things in js are objects

just use lodash's isobject function

apps/frontend/vue.config.js
```diff
@@ -1,4 +1,5 @@
 var webpack = require('webpack');
```
Amndeep7, line 1: const

openssl.cnf
Amndeep7, line 1: what is this for?  specifically why that 'unsafe' option?

package.json
```diff
@@ -31,7 +31,7 @@
     "lerna": "^5.0.0",
     "prettier": "^2.1.2",
     "prettier-plugin-organize-imports": "^3.0.0",
-    "typescript": "~4.6.4"
+    "typescript": "^4.9.3"
```
Amndeep7, line 34: we should see what breaks if we try to version bump to 5.x

Amndeep7 (changes requested):

apps/frontend/jest.config.js
Amndeep7: jest couldn't find it, this is a resolution fix for jest specifically

apps/backend/package.json
Amndeep7: instead of calling ts-node directly, use the loader flag to still use node as the runner but with ts-node as the loader

also changing the flag name from --loader to --experimental-loader since it was apparently changed to that

Amndeep7 (approved):

PR #3922, opened 2023-01-03
Implemented OIDC external group import

akey77:
Simple change in oidc.strategy.ts to automatically import groups from the configured OIDC provider. 
This allows heimdall groups to be managed from an external OIDC provider.

A new environment variable "OIDC_EXTERNAL_GROUPS" has been added to toggle this functionality on/off (it is off by default).


Amndeep7:
Thanks for the commit @akey77!  There are some additional changes that need to be made like adding the new environment variable to apps/backend/.env-example.

Also, I'd really appreciate you sending me an email at amann@mitre.org or saf@groups.mitre.org so that we can ensure your organizational needs are being met and that our workstreams are aligned for any/all future work that's relevant to you.

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
 
     return groupToDelete;
   }
+
+  syncUserGroups(user: User, groups: Array<string>) {
```
georgedias, line 116: Can we add a comment stating where is this called from - there is when is this check is done, and a high level objective of the code!

georgedias (commented):
Great work, we really appreciate your contributions for the improvement for benefiting the community.

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -39,15 +42,21 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
...
         if (email_verified) {
-          const user = authnService.validateOrCreateUser(
+          authnService.validateOrCreateUser(
             email,
             given_name,
             family_name,
             'oidc'
-          );
-          return done(null, user);
+          ).then((user) => {
```
Amndeep7, line 52: I personally prefer using `async/await` over `then` if possible.

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -3,21 +3,24 @@ import {PassportStrategy} from '@nestjs/passport';
...
 import {AuthnService} from './authn.service';
+import {GroupsService} from '../groups/groups.service';
 
 interface OIDCProfile {
   _json: {
     given_name: string;
     family_name: string;
     email: string;
     email_verified: boolean;
+    groups: Array<string>;
```
Amndeep7, line 14:
```suggestion
    groups: string[];
```

My preference

apps/backend/src/groups/groups.service.ts
```diff
@@ -25,6 +27,23 @@ export class GroupsService {
...
 
+  async findByName(name: string): Promise<Group> {
+    return this.findOneBang({
+      where: {
+        name
+      }
+    });
+  }
+
+  async findOneBang(options: FindOptions | undefined): Promise<Group> {
```
Amndeep7, line 38:
```suggestion
  async findOneBang(options: FindOptions?): Promise<Group> {
```

Amndeep7, line 38: I'd move this function definition above findbyname

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
 
     return groupToDelete;
   }
+
+  syncUserGroups(user: User, groups: Array<string>) {
+    // Check if user is any existing groups that they should not be
+    user.$get('groups', {include: [User]}).then((currentGroups) => {
```
Amndeep7, line 118: prefer async/await - see example in removeuserfromgroup

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
...
+      currentGroups.filter(group => !groups.includes(group.name)).forEach((groupToLeave) => {
```
Amndeep7, line 119: https://2ality.com/2021/01/looping-over-arrays.html

prefer `for of` over `foreach`

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
...
+        this.removeUserFromGroup(groupToLeave, user);
```
Amndeep7, line 120: shouldn't this be awaited to confirm that the user is removed from the group? especially since there's a potential for an error to be thrown

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
...
+      })
+    })
+
+    groups.forEach((group) => {
+      this.findByName(group).then((existingGroup) => {
+        // Check if the user is already in that group
+        user.$get('groups', {include: [User]}).then((groups) => {
```
Amndeep7, line 127: variable shadowing (repeating the name of groups) makes things unclear

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
...
+          const groupMap = groups.map((group) => new GroupDto(group));
+
+          if(!groupMap.includes(existingGroup)) {
+            this.addUserToGroup(existingGroup, user, "member");
+          }
+        });
+      }).catch((err) => {
```
Amndeep7, line 134: prefer full try/catch around an await

apps/backend/src/groups/groups.service.ts
```diff
@@ -93,4 +112,38 @@ export class GroupsService {
...
+        if(err instanceof NotFoundException) {
+
+          const createGroup: CreateGroupDto = {
+            name: group,
+            public: false
+          };
+
+          this.create(createGroup).then((newGroup) => {
+            this.addUserToGroup(newGroup, user, "member");
```
Amndeep7, line 143: would prefer to use a 'guard clause' type of control flow where we get the list of groups that the user is in, create any groups that don't already exist, and then go through and add the user to any groups they are not already in

Amndeep7 (changes requested):

apps/backend/.env-example
```diff
@@ -52,6 +52,7 @@ OIDC_TOKEN_URL=<Your OIDC token endpoint, for example: https://sample.us.auth0.c
 OIDC_USER_INFO_URL=<Your OIDC user info endpoint, for example: https://sample.us.auth0.com/userinfo>
 OIDC_CLIENTID=<Your OIDC Client ID>
 OIDC_CLIENT_SECRET=<Your OIDC Client Secret>
+OIDC_EXTERNAL_GROUPS=<If groups should be imported from the OIDC provider (if nothing is provided, defaults to false)>
```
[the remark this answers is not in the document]
akey77: Line to add to the 'Open ID Connect' section of the Environment-Variables-Configuration.md wiki page:

 - ``OIDC_EXTERNAL_GROUPS``: Import groups from external provider via OpenID Connect (OIDC) - true or false (defaults to false if not provided)

apps/backend/src/groups/groups.service.ts
akey77: This change appears to fail the build:

src/groups/groups.service.ts:31:56 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(this: ModelStatic<Group>, options: NonNullFindOptions<any>): Promise<Group>', gave the following error.
    Argument of type 'FindOptions<any> | null' is not assignable to parameter of type 'NonNullFindOptions<any>'.
      Type 'null' is not assignable to type 'NonNullFindOptions<any>'.
  Overload 2 of 2, '(this: ModelStatic<Group>, options?: FindOptions<any> | undefined): Promise<Group | null>', gave the following error.
    Argument of type 'FindOptions<any> | null' is not assignable to parameter of type 'FindOptions<any> | undefined'.
      Type 'null' is not assignable to type 'FindOptions<any> | undefined'.

31     const group = await this.groupModel.findOne<Group>(options);

akey77:
@Amndeep7 I think I covered everything in that last commit, as well as preventing crashing when trying to remove an owner from a group.

apps/backend/src/groups/groups.service.ts
Amndeep7: That might be because I put the question mark in the wrong place haha.  Can you try with `findonebang(options?: findoptions)`

aaronlippold:
@Amndeep7 can you review that @akey77 did resolve your concerns and we should pull in current main to keep this PR up to date in the codebase so we can get it merged.

Rlin232:
#4196 is a blocker for this PR

Amndeep7:
@akey77 this PR is now unblocked after https://github.com/mitre/heimdall2/pull/4403 gets merged which will resolve https://github.com/mitre/heimdall2/issues/4196

@Rlin232 will handle the merge conflicts that might arise from merging master into this branch, then it should be a final round of testing and this should be good to merge as well

apps/backend/.env-example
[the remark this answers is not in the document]
Amndeep7: use this instead "Synchronize user groups from external OIDC provider; Groups are not created automatically, users are only mapped into existing groups - true or false (defaults to false if not provided)" (to match env sample)

apps/backend/src/groups/groups.service.ts
```diff
@@ -142,4 +171,46 @@ export class GroupsService {
...
+        }
+      }
+    }
+
+    // Add user to existing groups that they aren't in yet
+    for (const existingGroup of existingGroups) {
+      if (!currentGroups.some((group) => group.name === existingGroup.name)) {
+        this.addUserToGroup(existingGroup, user, 'member');
+      }
+    }
```
Amndeep7, line 214: the addusertogroup promises that are being created in this for loop are not being awaited here nor explicitly returned for the caller to await

could be rewritten in a more functional / less-blocking manner as follows:

```suggestion
    // Add user to existing groups that they aren't in yet
    await Promise.all(existingGroups.filter(existingGroup => !currentGroups.some((group) => group.name === existingGroup.name)).map(this.addUserToGroup(existingGroup, user, 'member')));
```

but you could also just put an `await` in front of the 'addusertogroup' function call

apps/backend/src/groups/groups.service.ts
```diff
@@ -142,4 +171,46 @@ export class GroupsService {
...
+  // This method ensures that the passed in user is in all of the
+  // passed in groups, as long as the group already exists.
+  // It will additionally remove the user from any groups not in the list.
+  // Called from oidc.strategy.ts, if OIDC_EXTERNAL_GROUPS is enabled
+  async syncUserGroups(user: User, groups: string[]) {
+    const currentGroups = await user.$get('groups', {include: [User]});
+    const groupsToLeave = currentGroups.filter(
+      (group) => !groups.includes(group.name)
+    );
+    const existingGroups: Group[] = [];
```
Amndeep7, line 184: should be instantiated closer to where it's being used (the find existing groups block)

Amndeep7 (changes requested):
@akey77 after these last couple fixes, can we schedule a meeting to do a final round of live testing?

apps/backend/src/groups/groups.service.ts
Amndeep7: based off of the ci this fails cause a) the argument passed into the map isn't a function (we just need to turn it into one) and b) the linter

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -31,22 +34,31 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
...
             given_name,
             family_name,
             'oidc'
           );
+
+          if (
+            configService.get('OIDC_EXTERNAL_GROUPS') === 'true' &&
+            groups !== undefined
+          ) {
+            groupsService.syncUserGroups(user, groups);
```
Amndeep7, line 59: this function call needs to be awaited

apps/backend/src/groups/groups.service.ts
```diff
@@ -142,4 +171,42 @@ export class GroupsService {
 
     return groupToDelete;
   }
+
+  // This method ensures that the passed in user is in all of the
+  // passed in groups, as long as the group already exists.
+  // It will additionally remove the user from any groups not in the list.
+  // Called from oidc.strategy.ts, if OIDC_EXTERNAL_GROUPS is enabled
+  async syncUserGroups(user: User, groups: string[]) {
```
Amndeep7, line 179: at the end of this function, we should run the 'ensuregrouphasowner' function against any group that has been impacted by changes / potentially just against all groups entirely

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #3977, opened 2023-01-18
removed puppeteer dependencies since it was no longer being used

Amndeep7:
it seems like puppeteer was originally used for doing the ui testing for heimdall server, but that code was transitioned out to a top level directory (/tests as opposed to /apps or /libs) when the ui was deduped between server and lite, and then was also converted to being cypress


wdower (approved):
LGTM.

PR #3986, opened 2023-01-19
Added env var for disabling update checks

ebauman:
This PR addes an environment config item for disabling checks for Heimdall2 updates.

This is useful in cases where Heimdall2 is being used as a component of a larger system, and that system's maintainers do not wish this update banner to appear for end users.

It also is useful in airgapped usecases where Heimdall2 may not have access to GitHub to check for a newer version.

### DIFF
--- a/apps/backend/.env-example
+++ b/apps/backend/.env-example
@@ -25,6 +25,7 @@ ONE_SESSION_PER_USER=<If users are only allowed to be logged in on one browser, 
 CLASSIFICATION_BANNER_TEXT=<If a sensitivity classification banner should be shown to users, for example FOUO (if nothing is provided, no banner is shown)>
 CLASSIFICATION_BANNER_TEXT_COLOR=<The color of the text on the sensitivity classification banner, if enabled (defaults to white)>
 CLASSIFICATION_BANNER_COLOR=<The color of the sensitivity classification banner, if enabled (defaults to red)>
+DISABLE_UPDATE_CHECK=<Disable checking GitHub for an updated version of Heimdall2 (defaults to false)>
 
 
 # Oauth Client IDs and Secrets, If a variable does not have client id values assigned then the feature is disabled.
--- a/apps/backend/src/config/config.service.ts
+++ b/apps/backend/src/config/config.service.ts
@@ -56,7 +56,8 @@ export class ConfigService {
       oidcName: this.get('OIDC_NAME') || '',
       ldap: this.get('LDAP_ENABLED')?.toLocaleLowerCase() === 'true' || false,
       registrationEnabled: this.isRegistrationAllowed(),
-      localLoginEnabled: this.isLocalLoginAllowed()
+      localLoginEnabled: this.isLocalLoginAllowed(),
+      disableUpdateCheck: !!this.get('DISABLE_UPDATE_CHECK')
     });
   }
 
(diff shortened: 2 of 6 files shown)


apps/backend/.env-example
```diff
@@ -25,6 +25,7 @@ ONE_SESSION_PER_USER=<If users are only allowed to be logged in on one browser,
 CLASSIFICATION_BANNER_TEXT=<If a sensitivity classification banner should be shown to users, for example FOUO (if nothing is provided, no banner is shown)>
 CLASSIFICATION_BANNER_TEXT_COLOR=<The color of the text on the sensitivity classification banner, if enabled (defaults to white)>
 CLASSIFICATION_BANNER_COLOR=<The color of the sensitivity classification banner, if enabled (defaults to red)>
+DISABLE_UPDATE_CHECK=<Disable checking GitHub for an updated version of Heimdall2 (defaults to false)>
```
Amndeep7, line 28: please match other variable names

```suggestion
UPDATE_CHECK_DISABLED=<Disable checking GitHub for an updated version of Heimdall2 (defaults to false)>
```

apps/backend/src/config/config.service.ts
```diff
@@ -56,7 +56,8 @@ export class ConfigService {
       oidcName: this.get('OIDC_NAME') || '',
       ldap: this.get('LDAP_ENABLED')?.toLocaleLowerCase() === 'true' || false,
```
Amndeep7, line 57:
```suggestion
      ldap: this.get('LDAP_ENABLED')?.toLowerCase() === 'true',
```

apps/backend/src/config/config.service.ts
```diff
@@ -56,7 +56,8 @@ export class ConfigService {
...
       registrationEnabled: this.isRegistrationAllowed(),
-      localLoginEnabled: this.isLocalLoginAllowed()
+      localLoginEnabled: this.isLocalLoginAllowed(),
+      disableUpdateCheck: !!this.get('DISABLE_UPDATE_CHECK')
```
Amndeep7, line 60: This will only check if there's a non-empty string.  Also please propagate moving the 'disable' part to the end of the name.

```
$ node
Welcome to Node.js v18.12.1.
Type ".help" for more information.
> false
false
> 'false'
'false'
> !'false'
false
> !!'false'
true
```

```suggestion
      updateCheckDisabled: this.get('UPDATE_CHECK_DISABLED')?.toLowerCase() === 'true'
```

apps/frontend/src/store/app_info.ts
```diff
@@ -54,7 +55,7 @@ export class AppInfo extends VuexModule implements IAppInfoState {
 
   @Action
   public async CheckForUpdates() {
-    if (this.checkedForUpdates === false) {
+    if (!ServerModule.disableUpdateCheck && this.checkedForUpdates === false) {
```
Amndeep7, line 58: Please move this into its own guard clause where we can set 'SET_VERSION' to the deployed instance's version and 'SET_CHECKED_FOR_UPDATES' to true.

Amndeep7 (changes requested):

aaronlippold:
It seems like this PR has been reviewed, can the author resolve the review items so we can push ahead with the PR.

@ebauman @Amndeep7

PR #4071, opened 2023-02-10
Setup script fixes

Amndeep7:
added more vim files to the .gitignore, made new lines show up properly in the setup scripts, made the 'y/n' behavior the same between the two scripts, made the password that the user was typing in hidden like password prompts ought to be tho it doesn't really matter too much cause we throw it in that .env file anyways

### DIFF
--- a/setup-dev-env.sh
+++ b/setup-dev-env.sh
@@ -17,7 +17,8 @@ if ! grep -qF "DATABASE_USERNAME" .env; then
 fi
 
 if ! grep -qF "DATABASE_PASSWORD" .env; then
-	read -p 'Enter DATABASE_PASSWORD (leave blank to not set a password): ' psswd
+	read -s -p 'Enter DATABASE_PASSWORD (leave blank to not set a password): ' psswd
+	echo ""
 	echo "DATABASE_PASSWORD=$psswd" >> .env
 fi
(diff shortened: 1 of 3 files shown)


setup-dev-env.sh
```diff
@@ -17,7 +17,8 @@ if ! grep -qF "DATABASE_USERNAME" .env; then
 fi
 
 if ! grep -qF "DATABASE_PASSWORD" .env; then
-	read -p 'Enter DATABASE_PASSWORD (leave blank to not set a password): ' psswd
+	read -s -p 'Enter DATABASE_PASSWORD (leave blank to not set a password): ' psswd
+	echo ""
```
Hookwitz, line 21: Just curious. I assume the echo "" is for readability when running the script. Is it to provide visual separation to the DATABASE environment variables from JWT, API_KEY_SECRET, FQDN/Hostname/IP Address, and SSL Self-Signed Cert section?

Hookwitz (approved):

setup-dev-env.sh
Amndeep7: Talked it out over call, but basic answer is yes since some of the 'read' flags make it so that newlines aren't applied properly.

PR #4094, opened 2023-02-17
Updated COPY commands

catorreMC:
Added the built-in chmod/chown flags to improve build time. Removed standalone chmod and chown RUN commands, and removed all chmod/chown operations from the builder container. Will and I elected to keep the chmod and chown commands in the app stage of the image because they serve as a precautionary measure, and removing them makes no positive or negative impact on performance or space, and do not impact the legibility of the file.

On my machine, this improved build time from 2604 seconds to 309 - an 88 percent decrease. Pending further testing on other machines, as mine appeared to run it dramatically slower than others.

Dockerfile.lite still needs to be updated accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,43 +11,37 @@ USER 0
 COPY package.json yarn.lock lerna.json tsconfig.json .prettierrc ./
 COPY apps ./apps
 COPY libs ./libs
-RUN chmod 0500 -R apps libs
 RUN apk add python3 make g++
 RUN sed -i s^https://registry.yarnpkg.com^$YARNREPO^g yarn.lock
 RUN yarn --frozen-lockfile --production --network-timeout 600000
 
 RUN yarn run build
-RUN chown -R node apps libs package.json
-RUN chmod 0400 package.json libs/interfaces/package.json tsconfig.json .prettierrc
-RUN chmod 0500 -R apps libs
 
 ### Production image
 
 FROM $BASE_CONTAINER as app
 
 WORKDIR /app
-
-COPY package.json ./
-COPY --from=builder /src/apps/backend/package.json apps/backend/
-COPY --from=builder /src/apps/frontend/package.json apps/frontend/
-COPY --from=builder /src/libs/interfaces/package.json libs/interfaces/
-COPY --from=builder /src/libs/password-complexity/ libs/password-complexity/
-COPY --from=builder /src/apps/backend/node_modules apps/backend/node_modules
-COPY --from=builder /src/apps/backend/dist apps/backend/dist
-COPY --from=builder /src/dist/ dist/
-COPY --from=builder /src/apps/backend/.sequelizerc /app/apps/backend/
-COPY --from=builder /src/apps/backend/db /app/apps/backend/db
-COPY --from=builder /src/apps/backend/config /app/apps/backend/config
-COPY --from=builder /src/apps/backend/migrations /app/apps/backend/migrations
-COPY --from=builder /src/apps/backend/seeders /app/apps/backend/seeders
-RUN chown node package.json libs libs/interfaces libs/interfaces/package.json libs/password-complexity apps/backend/node_modules apps/backend/seeders apps/backend/config apps/backend/db apps/backend/migrations apps/backend/seeders apps/backend/package.json apps/backend apps/frontend apps/frontend/package.json
-RUN chmod 0400 libs/interfaces/package.json apps/frontend/package.json
+#Note for posterity: the permissions setting is to ensure that the user cannot modify the files accidentally and can't be read, executed or changed by another account.
+#Removing these permissions doesn't improve image size or performance meaningfully, but keeping them adds a potential safeguard.
+COPY --chown= --chmod=0400 package.json ./
+COPY --from=builder --chown= --chmod=0400 /src/apps/frontend/package.json apps/frontend/
+COPY --from=builder --chown= --chmod=0400 /src/libs/interfaces/package.json libs/interfaces/
+COPY --from=builder --chown= --chmod=0500 /src/libs/password-complexity/ libs/password-complexity/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/package.json apps/backend/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/node_modules apps/backend/node_modules
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/dist apps/backend/dist
+COPY --from=builder --chown= /src/dist/ dist/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/.sequelizerc /app/apps/backend/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/db /app/apps/backend/db
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/config /app/apps/backend/config
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/migrations /app/apps/backend/migrations
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/seeders /app/apps/backend/seeders
 
 
 EXPOSE 3000
 
-COPY cmd.sh /usr/local/bin/
-RUN chmod 755 /usr/local/bin/cmd.sh
+COPY --chmod=755 cmd.sh /usr/local/bin/
 
 USER node
 


Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
 RUN yarn --frozen-lockfile --production --network-timeout 600000
 
 RUN yarn run build
-RUN chown -R node apps libs package.json
-RUN chmod 0400 package.json libs/interfaces/package.json tsconfig.json .prettierrc
-RUN chmod 0500 -R apps libs
 
 ### Production image
 
 FROM $BASE_CONTAINER as app
```
Amndeep7, line 22: following along with the snyk blog should we be setting a node_env variable?  i checked out setup env scripts and we're doing that in the dev env script (though setting the var to development) but not doing it at all in the docker env script.  i think the idea was that it would get passed through in the docker-compose (which is where we do have that var set to production), but then that means that it is not present when one tries using the image directly (e.x. running it locally or on kubernetes / non docker-compose system).  furthermore, should it be in both stages?  just this one?

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
 COPY package.json yarn.lock lerna.json tsconfig.json .prettierrc ./
```
Amndeep7, line 11: why do we have to copy all these separately if they're not being renamed and put in different directories?  if there are things we don't want to have copied over (such as a lot of config stuff in the top level of the repo), it seems like we should expand the .dockerignore accordingly.  and even then, why is it necessary to have some of these config files like the prettier one?  would the build fail due to some sorta linting issue if it wasn't present?

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
-COPY --from=builder /src/apps/backend/dist apps/backend/dist
-COPY --from=builder /src/dist/ dist/
-COPY --from=builder /src/apps/backend/.sequelizerc /app/apps/backend/
-COPY --from=builder /src/apps/backend/db /app/apps/backend/db
-COPY --from=builder /src/apps/backend/config /app/apps/backend/config
-COPY --from=builder /src/apps/backend/migrations /app/apps/backend/migrations
-COPY --from=builder /src/apps/backend/seeders /app/apps/backend/seeders
-RUN chown node package.json libs libs/interfaces libs/interfaces/package.json libs/password-complexity apps/backend/node_modules apps/backend/seeders apps/backend/config apps/backend/db apps/backend/migrations apps/backend/seeders apps/backend/package.json apps/backend apps/frontend apps/frontend/package.json
-RUN chmod 0400 libs/interfaces/package.json apps/frontend/package.json
+#Note for posterity: the permissions setting is to ensure that the user cannot modify the files accidentally and can't be read, executed or changed by another account.
```
Amndeep7, line 25:
```suggestion
# The permissions setting is to ensure that the user cannot modify the files accidentally and can't be read, executed or changed by another account.  Removing these permissions doesn't improve image size or performance meaningfully, but keeping them adds a potential safeguard.
```

lines can be long lol.  that's what wrapping is for

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
+#Removing these permissions doesn't improve image size or performance meaningfully, but keeping them adds a potential safeguard.
```
Amndeep7, line 26:
```suggestion
```

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
+COPY --chown= --chmod=0400 package.json ./
+COPY --from=builder --chown= --chmod=0400 /src/apps/frontend/package.json apps/frontend/
+COPY --from=builder --chown= --chmod=0400 /src/libs/interfaces/package.json libs/interfaces/
+COPY --from=builder --chown= --chmod=0500 /src/libs/password-complexity/ libs/password-complexity/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/package.json apps/backend/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/node_modules apps/backend/node_modules
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/dist apps/backend/dist
+COPY --from=builder --chown= /src/dist/ dist/
```
Amndeep7, line 34: why isn't this being chmod'd?

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/.sequelizerc /app/apps/backend/
```
Amndeep7, line 35: why do the copies start doing absolute paths for the copy instead of relative like we have been?

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
 COPY package.json yarn.lock lerna.json tsconfig.json .prettierrc ./
 COPY apps ./apps
 COPY libs ./libs
-RUN chmod 0500 -R apps libs
 RUN apk add python3 make g++
 RUN sed -i s^https://registry.yarnpkg.com^$YARNREPO^g yarn.lock
 RUN yarn --frozen-lockfile --production --network-timeout 600000
```
Amndeep7, line 16:
```suggestion
RUN yarn install --frozen-lockfile --production --network-timeout 600000
```

would feel better if we explicitly labeled the action being done here

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/dist apps/backend/dist
+COPY --from=builder --chown= /src/dist/ dist/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/.sequelizerc /app/apps/backend/
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/db /app/apps/backend/db
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/config /app/apps/backend/config
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/migrations /app/apps/backend/migrations
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/seeders /app/apps/backend/seeders
 
 
 EXPOSE 3000
```
Amndeep7, line 42: expose makes a zero-size layer but still impacts caching for lower layers.  since it doesn't change, let's just throw it above the copies in this stage esp since it'll also allow the copy of the cmd.sh to be closer to the other copies.

Amndeep7 (changes requested):

Dockerfile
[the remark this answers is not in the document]
catorreMC: I'll give this a look. Surprised I hadn't seen it before.

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
 COPY package.json yarn.lock lerna.json tsconfig.json .prettierrc ./
 COPY apps ./apps
 COPY libs ./libs
-RUN chmod 0500 -R apps libs
 RUN apk add python3 make g++
```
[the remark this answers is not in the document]
catorreMC: Good point, it also looks like it'll help me do my job better. I'll give it a look.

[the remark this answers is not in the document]
catorreMC: Very true, though we also need to be wary of updates potentially breaking compatibility with parts of the application.

aaronlippold:
It seems to me that this container should be running as either a heimdall user or a heimdallapp or nodeapp or something like that - and thus in the final build layer there should be commands making sure all the data on the image is owned by that user:

```
COPY --from=build /usr/bin/dumb-init /usr/bin/dumb-init
USER node
WORKDIR /usr/src/app
COPY --chown=node:node --from=build /usr/src/app/node_modules /usr/src/app/node_modules
COPY --chown=node:node . /usr/src/app
```
`chmod,chwon` commands can be joined and they can also be recessive. So I can `chown -R user:group <path>` 
next
The files and folders on the system, I would think, only need at max: `00640 for -type f` and `00750 -type d` although enabling the stickly bit on all directory structures in the base layers with the right mask may save some time an effort here. However this may be showing a bug that would mean we may have to be more hands on with managing this.

<https://forums.docker.com/t/permissions-issue-with-compose-build-vs-docker-build/7775>

Dockerfile
Amndeep7: no like, just make it a 'yarn install' instead of a 'yarn' - see the commit sugggestion i've provided.  it doesn't need a comment.

Amndeep7:
> It seems to me that this container should be running as either a heimdall user or a heimdallapp or nodeapp or something like that - and thus in the final build layer there should be commands making sure all the data on the image is owned by that user:

That's why we have that 'node' user.  They've basically done the work to properly create a different user account from root and we can do whatever we want with it now as opposed to having to manually create a 'heimdall' user.

Dockerfile
catorreMC: Missed that on the codespace GUI - yeah, I second that, because I didn't initially know what that line was doing, either.

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/config /app/apps/backend/config
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/migrations /app/apps/backend/migrations
+COPY --from=builder --chown= --chmod=0500 /src/apps/backend/seeders /app/apps/backend/seeders
 
 
 EXPOSE 3000
 
-COPY cmd.sh /usr/local/bin/
-RUN chmod 755 /usr/local/bin/cmd.sh
+COPY --chmod=755 cmd.sh /usr/local/bin/
```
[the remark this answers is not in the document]
catorreMC:
```suggestion
COPY --chmod=555 cmd.sh /usr/local/bin/
```
Can't see a reason not to - cmd.sh shouldn't need to be overwritten in the course of execution as far as I can tell, so I'll commit this change.

Dockerfile
```diff
@@ -11,43 +11,37 @@ USER 0
...
-COPY --from=builder /src/apps/backend/.sequelizerc /app/apps/backend/
-COPY --from=builder /src/apps/backend/db /app/apps/backend/db
-COPY --from=builder /src/apps/backend/config /app/apps/backend/config
-COPY --from=builder /src/apps/backend/migrations /app/apps/backend/migrations
-COPY --from=builder /src/apps/backend/seeders /app/apps/backend/seeders
-RUN chown node package.json libs libs/interfaces libs/interfaces/package.json libs/password-complexity apps/backend/node_modules apps/backend/seeders apps/backend/config apps/backend/db apps/backend/migrations apps/backend/seeders apps/backend/package.json apps/backend apps/frontend apps/frontend/package.json
-RUN chmod 0400 libs/interfaces/package.json apps/frontend/package.json
+#Note for posterity: the permissions setting is to ensure that the user cannot modify the files accidentally and can't be read, executed or changed by another account.
+#Removing these permissions doesn't improve image size or performance meaningfully, but keeping them adds a potential safeguard.
+COPY --chown= --chmod=0400 package.json ./
```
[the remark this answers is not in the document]
catorreMC: I left that as-is from the previous version; I could certainly change it for consistency, it is definitely easier to read if that group of lines is consistent.

```suggestion
COPY --from-builder --chown= --chmod=0400 ./package.json ./
```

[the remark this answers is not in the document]
catorreMC: As far as setting it to nothing - that is an error on my part - I should have put node into all the chown flags.

That said, according to commit history, back in 2022 there's another commit to "optimize the dockerfile" that got rid of the -R flag on line 43 and added the chown -R node on line 20 into effect, so I believe it *was* added for a reason. I'm leaning towards keeping them owned by node, personally, but I do agree that research could be done to definitively find the best course of action.

Dockerfile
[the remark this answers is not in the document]
catorreMC: I've read that article for my previous project- I hadn't consulted it since it was for python, not node, but that's a fair point. 
As far as size, the multi-stage build should (or possibly could, if it isn't) offset the bloat, but time spent building dependencies is definitely a valid concern.

To definitively answer what the best choice is, though, we have to find what dependencies are being built that could be pulled if we used another image, find how much space they account for, and how much space/time we'd save by switching to that other image. Additional research could be done for this, but I'm beginning to run short on time on this project myself.

Dockerfile
catorreMC: I'll agree prettier probably shouldn't need to be there - the rest seem to be updated regularly. Package.json, yarn.lock, and tsconfig.json are dependencies or config files for dependencies, which would probably/definitely break things if removed. Lerna seems to be for the monorepo, which I'm not experienced with, and .prettierc can probably be removed and added to the .gitignore without issue.

As far as copying them all separately from everything else, they could theoretically be in their own folder named 'dependencies' but then they'd probably simply be copied to the root of the container anyways, so I don't see much value in doing that for the sake of Dockerfile performance. Repo housekeeping is a different story.

Dockerfile
catorreMC: it wasn't being chmod'd before - that could be changed, after all I can't think of a reason for anything in /src to be overwritten over the course of app execution.
```suggestion
COPY --from=builder --chown --chmod=0500= /src/dist/ dist/
```

Dockerfile
catorreMC: Good question - I didn't change that part of the line, but I don't have any rationale for keeping it that way. That could be changed for consistency.

Dockerfile
catorreMC: Good point. I'll make that change.

Dockerfile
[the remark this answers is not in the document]
catorreMC: I agree with you here - I've added them to both stages. Build speed is affected, but better to be slightly slower than less secure.

PR #4114, opened 2023-02-22
Remove advanced codeql configuration

Amndeep7:
this workflow is failing since we also have codeql running from the 'default' setup (i.e. repo setting option).  we're not doing anything fancy, so i'd rather use the default over our 'advanced' configuration.

https://github.com/mitre/heimdall2/actions/runs/4246770128
https://github.blog/2023-01-09-default-setup-a-new-way-to-enable-github-code-scanning/


wdower (approved):
LGTM

PR #4143, opened 2023-03-01
Fixed Compare Results Bug

### LINKED ISSUE #3902: [BUG] Profiles that are loaded in at the same time as results cause a type error
**Describe the bug**
When a profile and results file are loaded in at the same time, a type error appears at delta_util:

image

Causes at least the following issues:
- 'display only changed results' button in the comparison view doesn't do anything

**To Reproduce**
1. Load one or more results files
2. Load a profile
3. See error pop up in terminal
4. (Optional) Go to comparison view and see how bits and pieces of the UI don't work

**Expected behavior**
The comparison view works properly / there's no type error

**Screenshots**

Before loading the profile:

image

After loading the profile:

image

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Firefox
 - Version: 108.0

**Additional context**
If you could also submit the file that would also be great!

ChrisHinchey:
Resolves #3902


apps/frontend/src/utilities/delta_util.ts
```diff
@@ -155,9 +155,11 @@ export function get_eval_start_time(
   ev: ContextualizedEvaluation
 ): string | null {
   for (const prof of ev.contains) {
-    for (const ctrl of prof.contains) {
-      if (ctrl.hdf.segments!.length) {
-        return ctrl.hdf.segments![0].start_time;
+    if ('contains' in prof) {
```
Amndeep7, line 158: prof.contains is a array of contextualizedprofiles which themselves implement the contains interface but with item being an array of contextualizedcontrols so why is the type system throwing a fit for no good reason???????

Amndeep7 (changes requested):

Amndeep7 (approved):

Issue #3020, opened 2022-05-11
Exports in profile view


Amndeep7:
You can export a profile.json as csv or html, but you cannot export one to xccdf or asff (they crash or return empty files).  Exporting as json just returns the original profile.json.

Should all the exporters return valid files (requires modifications to the mappers to support profile jsons and not just hdf files) or should we remove non-functional exporters from that view (i.e. the export as drop down will not contain options for asff/xccdf/etc)?

@ejaronne 

Thanks to @em-c-rod for making me think of the profile view.

ejaronne:
For now, just keep/show the export json choice.

PR #4172, opened 2023-03-07
We should only export profile jsons in the profile view, i.e. no XCCDF/CAAT/CSV/HTML/whatever

Amndeep7:
Resolves https://github.com/mitre/heimdall2/issues/3020

### DIFF
--- a/apps/frontend/src/views/Results.vue
+++ b/apps/frontend/src/views/Results.vue
@@ -20,7 +20,7 @@
             </v-btn>
           </template>
           <v-list class="py-0">
-            <v-list-item class="px-0">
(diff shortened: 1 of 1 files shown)


wdower (approved):
LGTM

Amndeep7:
image

Successfully only showing the JSON exporter.  Tested it just to make sure and got the same profile out as I put in (though the keys were in different order, which was easily fixed by running `jq --sort-keys '.'` on both of them before running the diff.

ejaronne:
Why did you kill off csv and html? Or is this a matter of circling back and tailoring those to be relevant for guidance content only?

PR #4212, opened 2023-03-15
BUG Fix for Heimdall CKL Export

Hookwitz:
set vid that maps to Vuln_Num to use tags.gid if available otherwise use control id


ejaronne:
Based on this analysis and conclusion on row 11, I approve the proposed change
image

ejaronne (approved):
LGTM!

PR #4242, opened 2023-03-21
swap clipboard library

Amndeep7:
added types override to deal with the fact that the types only work with vue3 even though the library works fine with vue2

should have also included https://github.com/mitre/heimdall2/commit/e6440200cdfa127e73245a7b7706820ac213f7f3 but I screwed up and pushed that straight to master


Amndeep7:
also i recreated the yarn lock file

PR #4243, opened 2023-03-21
netsparker is now invicti

Amndeep7:


ejaronne (approved):
LGTM!

libs/hdf-converters/src/netsparker-mapper.ts
```diff
@@ -139,87 +139,102 @@ function formatMessage(response: unknown): string {
...
+      platform: {
+        name: 'Heimdall Tools',
+        release: HeimdallToolsVersion,
+        target_id: {path: `${toolname}-enterprise.target.url`}
+      },
+      version: HeimdallToolsVersion,
+      statistics: {},
+      profiles: [
+        {
+          name: `${toolname.replace(/^./, (firstLetter) =>
```
Hookwitz, line 155: If we are controlling 'toolname' could we use toolname[0].toUpperCase() + toolname.substring(1). I don't think there is a performance issue, but regex to return the first character seems over-engineered. And would it be useful to save this as a const variable above the return since the upperCase is performed multiple times?

Hookwitz (changes requested):

libs/hdf-converters/src/netsparker-mapper.ts
Amndeep7: I can put it into its own variable.  I find 'replace' as a function making more sense to me as to what's going on than arbitrary string concatenation.

Hookwitz: Approving with condition of change

Hookwitz (approved):

PR #4254, opened 2023-03-23
Add error check in xccdf_results2hdf mapper

em-c-rod:
- [ ] add check that the text field exists before calling .match on it
- [ ] also add logger
- [ ] Test that logging not displays as expected


aaronlippold:
#4255 also seems to update the CCI linking ... maybe worth a look

Issue #4194, opened 2023-03-13
[BUG] `@mitre/hdf-converters` fails converting a valid xccdf file to hdf: TypeError: Cannot read properties of undefined (reading 'match')


candrews:
**Describe the bug**
`@mitre/hdf-converters` cannot convert a valid xccdf file to hdf.

**To Reproduce**
1. Given this [report.xml](https://github.com/mitre/saf/files/10937540/report.xml.txt) (which is a valid xccdf report produced by `openscap --results`) run:
`npx @mitre/saf@1.2.7 convert xccdf_results2hdf -i report.xml -o report.hdf`

**Actual behavior**
```
$ npx @mitre/saf@1.2.7 convert xccdf_results2hdf -i report.xml -o report.hdf
(node:86425) NOTE: We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023.

Please migrate your code to use AWS SDK for JavaScript (v3).
For more information, check the migration guide at https://a.co/7PzMCcy
(Use `node --trace-warnings ...` to show where the warning was created)
    TypeError: Cannot read properties of undefined (reading 'match')
[1]
```

**Expected behavior**
The conversion should complete successfully.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: n/a
 - Version: `@mitre/hdf-converters@2.6.38`

candrews:
The only place where `match` is read/called is https://github.com/mitre/heimdall2/blob/v2.6.41/libs/hdf-converters/src/xccdf-results-mapper.ts#L190 so that must be where this error occurs.

The problem appears to be that `@mitre/hdf-converters` expects the xccdf file to always have `xccdf:ident` elements, and this file does not have any `xccdf:ident` elements.

Note that according to page 23 of the [XCCDF 1.2 Specification](https://csrc.nist.gov/CSRC/media/Publications/nistir/7275/rev-4/final/documents/nistir-7275r4_updated-march-2012_clean.pdf), `xccdf:ident` elements are optional (may have 0 count):
![image](https://user-images.githubusercontent.com/194713/224588996-3975d2a0-b50b-40a2-adf5-be2cafd3cf7a.png)

PR #3859, opened 2022-12-14
Checklist Mapper

Hookwitz:
Checklist Mapper created with tests WIP:

- [x] Functionality
    - [x] Added JSONIX library to hdf-converters
    - [x] Added parsing functionality to base-converter.ts for inherited functionality
    - [x] ChecklistMapper.ts will parse a checklist xml and map to hdf format utilizing much of the intermediate object/functionality
    - [x] Added a passthrough object for Heimdall frontend checklist viewer ease of access
    - [x] Custom parser to create multiple ControlResult objects from findingDetails using the existing Export as Checklist functionality from Heimdall
    - [x] Create logic to parse multiple hdf object from multiple iSTIG checklists
    - [x] Create logic to generate a wrapper and add parent_profile to existing hdf profiles
    - [x] Update Sample function to include ckl - loadText function assumes hdf
 
- [x] UI
    - [x] Add ckl samples to Samples

- [x] Tests
    - [x] Add tests/results for hdf converters

Signed-off-by: David Markowski <dmarkowski@mitre.org>

Issue #3084, opened 2022-05-27
Remove Namespaces for XCCDF Import


camdenmoors:
See https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v4/2.XMLparseOptions.md#removensprefix, update mapper to not expect namespaces.

PR #4255, opened 2023-03-24
Improve xccdf_results_mapper when converting XCCDF Results to HDF Results

candrews:
* Support XCCDF generated using ComplianceAsCode: https://github.com/ComplianceAsCode/content/
* Support nested XCCDF groups
* Set `code_desc` to an appropriate value (not empty string)
* Refactor to remove global variables

Bug: https://github.com/mitre/saf/issues/1163
Fixes: https://github.com/mitre/heimdall2/issues/4194


aaronlippold:
Hi the team @Amndeep7 @em-c-rod @wdower @Hookwitz @georgedias are looking over this and seeing if some of our other PR work can be rolled into this as well. 

A couple questions:
- Given some of the changes that were make, would we be able to make some smaller unit tests that cover actual function @em-c-rod @Amndeep7 @candrews 
- Should we add test data for other XCCDF profiles -windows, Ubuntu, DB, IIS, etc or would a subset of XCCDF content element likely be sufficient
- Have we seen any corner cases - nesting, tags inside tags, etc - that we should make specific tests for?

aaronlippold:
@candrews looks like SonarCloud is also unhappy about something - which we may want to poke at as well as doing a rebase from master/main

aaronlippold:
Question: Is any of the work here also useful or overtaken by this PR? https://github.com/mitre/heimdall2/pull/2958

candrews:
> looks like SonarCloud is also unhappy about something - which we may want to poke at as well as doing a rebase from master/main

I fixed the SonarCloud findings. The only finding now is in the test for a comment (the same comment is in every other test, so I think it should be in this one as well) and I believe that finding should be suppressed.

candrews:
> Question: Is any of the work here also useful or overtaken by this PR? #2958

This PR completely supersedes #2958 - it does everything that PR does and more.

candrews:
Can I do something to help this along?

I'd really like to have these issues fixed soon as it's blocking progress on a project which is import to my corporate overlords.

Amndeep7:
Sorry for the delay @candrews - could you address the findings for the HDF-Converters tests?  It seems like some of the other XML mappers such as the Veracode are now having issues.

Amndeep7:
Another point that I wanted to bring up was that we were intending on eventually rewriting this mapper to use jsonix to be able to match against the XCCDF schema properly instead of using the fast-xml-parser which has issues with (for example) determining the difference between singular nodes and arrays of length one.  This isn't a request for you to rework this entire PR to use that other library, just a heads up about part of the delay in dealing with this mapper since we're still exploring how to use jsonix and write a mapper properly with it in our checklist mapper work (https://github.com/mitre/heimdall2/pull/3859).

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -23,79 +23,70 @@ const IMPACT_MAPPING: Map<string, number> = new Map([
...
-      testResult,
-      pathRuleResult
-    ) as Record<string, unknown>[] | undefined;
-    if (ruleResult === undefined) {
+function getRuleResult(
+  ruleId: string,
+  benchmark: Record<string, unknown>
+): Record<string, unknown> | undefined {
+  for (const pathRuleResult of [
+    'cdf:TestResult.cdf:rule-result',
```
Amndeep7, line 41: After we'd written this, we saw that there might be ways to make this easier.

https://github.com/mitre/heimdall2/issues/3084

The blocker for this is updating the version of the fast-xml-parser, which I could look into hopefully soon.

Amndeep7 (changes requested):

Amndeep7:
Still need to finish reading/reviewing the code, which I'll try to do next week.

Also @candrews - it would be really nice if we could arrange a meeting to discuss how you and your corporate overlords are using HDF-Converters / the SAF.  We don't do any tracking, so we gotta go the old fashioned way and actually talk to our users in order to learn how/why they're using our stuff, and what we could do to deepen the partnership.

If you could email me at amann@mitre.org with some dates/times that you/your team/your management chain are available, I'd appreciate it a lot.

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: That got merged

candrews: I've updated this MR and this change certainly does make the task much easier and less error prone!

candrews:
> This pull request has a conflict. Could you fix it @candrews?

I've addressed the conflict.

candrews:
> Sorry for the delay @candrews - could you address the findings for the HDF-Converters tests? It seems like some of the other XML mappers such as the Veracode are now having issues.

As far as I can tell, all tests pass. If there is a problem, can you please point me to it?

candrews:
@Amndeep7 do you have an update as to when you'll be able to review and (hopefully) merge this request?

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +307,141 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-              severity: {path: 'Rule.severity'},
+              severity: {path: 'severity'},
               description: {
-                path: ['Rule.description.text', 'Rule.description'],
-                transformer: (description: string) =>
-                  JSON.stringify(
-                    _.pickBy(convertEncodedXmlIntoJson(description), _.identity)
-                  )
+                path: ['description.text', 'description'],
+                transformer: (description: string): string => {
```
Amndeep7, line 325: you don't need a wrapping pair of braces + return on an arrow function if it's a single expression

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +307,141 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-                  const paths = {
-                    id: ['id'],
-                    description: ['description.text', 'description'],
-                    title: ['title.text', 'title']
-                  };
-                  return getProfiles(profiles, pathsSelect, paths);
+              ident: {path: 'ident'},
+              reference: {path: 'reference'},
+              selected: {path: 'selected'},
+              version: {path: ['id', 'version']}, // dunno why the field is called version when it's just an old id
```
Amndeep7, line 354: i think i had id in here as a bug, so we should be safe to make it only 'version' unless you've found id to also be relevant here.  i fixed it when i did the xml parser update.

libs/hdf-converters/test/mappers/forward/xccdf_mapper.spec.ts
```diff
@@ -54,6 +54,30 @@ describe('xccdf_mapper', () => {
...
   describe('OpenSCAP', () => {
+    it('ubuntu1804', () => {
+      const mapper = new XCCDFResultsMapper(
+        fs.readFileSync(
+          'sample_jsons/xccdf_results_mapper/sample_input_report/xccdf-results-openscap-ComplianceAsCode-ubuntu1804.xml',
+          {encoding: 'utf-8'}
+        )
+      );
+
+      // fs.writeFileSync(
```
Amndeep7, line 65: ignore the sonarqube codesmell warning for the commented out code here.  if it has any other findings though, please do fix them.

libs/hdf-converters/sample_jsons/xccdf_results_mapper/sample_input_report/xccdf-results-openscap-ComplianceAsCode-ubuntu1804.xml
Amndeep7, line 1: If you check out the deploy preview (https://deploy-preview-4255--heimdall-lite.netlify.app), it does not seem to parse + convert the XCCDF file in a reasonable amount of time.  The other files still completed within a few seconds.  This makes me feel like it's probably something to do with the recursion step.

If you see the results of running the tests on this pr, the xccdf unit tests took 155.569 s (https://github.com/mitre/heimdall2/actions/runs/4692756126/jobs/8318869062?pr=4255).  On some random other pr, it took 19.37 s (https://github.com/mitre/heimdall2/actions/runs/4921274476/jobs/8790946044).

Amndeep7, line 1: Please make sure to add/update the sample files in the saf cli as well.

Amndeep7 (changes requested):

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -21,73 +21,56 @@ const IMPACT_MAPPING: Map<string, number> = new Map([
...
+): Record<string, unknown> | undefined {
+  const ruleResults = asArray(
+    _.get(benchmark, 'TestResult.rule-result') as
+      | Record<string, unknown>
+      | Record<string, unknown>[]
+  );
+  const matchingRuleResult = ruleResults.find(
+    (element) => _.get(element, 'idref') === ruleId
+  );
+  if (matchingRuleResult) {
```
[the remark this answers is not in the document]
candrews: Good point! I've made this change :+1:

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -97,65 +80,63 @@ function convertEncodedXmlIntoJson(
   return parseXml(encodedXml);
```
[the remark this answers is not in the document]
candrews: I used to have it doing more... I didn't notice it became a useless wrapper :)

I've made the suggested change :+1:

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -164,19 +145,14 @@ interface IIdent {
...
-    inputArray = input;
-  } else {
-    inputArray = [input];
-  }
+  const inputArray = asArray(input);
 
   const CCI_REGEX = /CCI-(\d*)/;
 
   const output: string[] = [];
   inputArray.forEach((element) => {
```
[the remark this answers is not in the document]
candrews: Sure, I've made this change :+1:

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: I've made that change (and the same change in other such places) :+1:

libs/hdf-converters/test/mappers/forward/xccdf_mapper.spec.ts
candrews: This is the only finding reported by sonarqube; I'll continue to ignore it :)

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -21,73 +21,56 @@ const IMPACT_MAPPING: Map<string, number> = new Map([
...
+      return ExecJSON.ControlResultStatus.Error;
+    case 'unknown':
+      return ExecJSON.ControlResultStatus.Error;
+    case 'notapplicable':
+      return ExecJSON.ControlResultStatus.Skipped;
+    case 'notchecked':
+      return ExecJSON.ControlResultStatus.Skipped;
+    case 'notselected':
+      return ExecJSON.ControlResultStatus.Skipped;
+    case 'informational':
```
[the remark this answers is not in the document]
candrews: I've implemented this change :+1:

libs/hdf-converters/sample_jsons/xccdf_results_mapper/sample_input_report/xccdf-results-openscap-ComplianceAsCode-ubuntu1804.xml
candrews: I'm assuming you mean at https://github.com/mitre/saf/tree/main/test/sample_data

That data would be updated as part of the PR that updates the version of the `@mitre/hdf-converters` dependency at https://github.com/mitre/saf/blob/73fa1def42f5eb39602d64debbe8a57a2ca5e2a2/package-lock.json#L223 which can only be done after this PR is merged and an updated version of this project (`@mitre/hdf-converters`) is released, right?

Amndeep7: you can just make a draft pr in that repo and say it's blocked by this one getting merged/released

candrews: I found the culprit and fixed it: https://github.com/mitre/heimdall2/pull/4255/commits/c2640f507bf4862cd5496e24b01a79cf2f282200

The tests now run in 13 seconds; without this change, they took 53 seconds.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: Removing `id` does change the output resulting in "version" not being set in some examples.

Perhaps we can consider that in a separate issue from this one?

candrews:
@Amndeep7 I believe I've addressed all feedback and that all tests pass - can you please take another look?

candrews:
As far as I can tell, the ["Run E2E Backend + Frontend Tests / build (pull_request)"](https://github.com/mitre/heimdall2/actions/runs/4952309550/jobs/8859791454?pr=4255) test failure is unrelated to this PR - please correct me if I'm wrong.

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: Sure we can make this a separate issue.  Maybe we change the order of attempted paths such that it attempts to find 'version' before using 'id' as a fallback?

candrews: > Maybe we change the order of attempted paths such that it attempts to find 'version' before using 'id' as a fallback?

Sure, I've added a commit that does so.

Amndeep7:
@candrews did you see my message on teams?  It seems like there are a handful of comments that you did not see/address yet.  Github's done this to me before where people haven't received some of the feedback that I wrote despite me nominally having submitted the review.  Do you see the below comment?

image

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: Looks great.  We can probably remove the comment there too :)

candrews: I've added a commit that remove that comment :)

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -164,19 +145,14 @@ interface IIdent {
...
-  }
+  const inputArray = asArray(input);
 
   const CCI_REGEX = /CCI-(\d*)/;
 
   const output: string[] = [];
   inputArray.forEach((element) => {
-    const text = _.get(element, 'text');
-    if (text.match(CCI_REGEX)) {
+    const text = String(_.get(element, 'text'));
```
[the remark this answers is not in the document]
candrews: I don't know `String` is there... I agree it shouldn't be there.

I've added a commit removing it.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -223,7 +252,7 @@ export class XCCDFResultsMapper extends BaseConverter {
           path: 'Benchmark',
           transformer: (input: Record<string, unknown>): string => {
             const descriptionPaths = [
-              ['description.text', 'description'],
+              ['description.text'],
```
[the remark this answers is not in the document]
candrews: Maybe? I'd have to study more xccdf files and see. However, this seems to be good for now, at least.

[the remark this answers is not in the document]
candrews: Because here, it's the path is used as an "and" not an "or"

The value used here is the combination of the list items (not a fallback, as is the case elsewhere). So if we used `description` and `description.text` here, then the value used would have the description twice, which is not desirable. That's observable by making this change then running the tests - the duplication is evident in the output.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -192,6 +168,59 @@ function nistTag(input: IIdent | IIdent[]): string[] {
   );
 }
 
+/**
+ * Given a group, returns all the rules within it (including rules in groups nested within the group).
+ */
+function getRulesInGroup(
```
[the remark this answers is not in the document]
candrews: I added tsdoc to this function to clarify how it works

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -164,19 +145,14 @@ interface IIdent {
...
+  const inputArray = asArray(input);
 
   const CCI_REGEX = /CCI-(\d*)/;
 
   const output: string[] = [];
   inputArray.forEach((element) => {
-    const text = _.get(element, 'text');
-    if (text.match(CCI_REGEX)) {
+    const text = String(_.get(element, 'text'));
+    if (!!text && text.match(CCI_REGEX)) {
```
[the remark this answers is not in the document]
candrews: Since the String conversion has been removed, the `!!text` part still serves a purpose (catching null / undefined)

candrews:
> did you see my message on teams?

Yes, I did

> It seems like there are a handful of comments that you did not see/address yet. Github's done this to me before where people haven't received some of the feedback that I wrote despite me nominally having submitted the review. Do you see the below comment? There are a few others as well.

I found https://github.com/mitre/heimdall2/pull/4255/files/88cd988c1d5771c7e5e0c5be16c010ae607cad6a..6992f6d7234f03ef14aab5d6d1ce027f981db659#r1187829883 and addressed it.

I also found and addressed these:
* https://github.com/mitre/heimdall2/pull/4255/files#r1187864828
* https://github.com/mitre/heimdall2/pull/4255/files#r1187868617
* https://github.com/mitre/heimdall2/pull/4255/files#r1187840749
* https://github.com/mitre/heimdall2/pull/4255/files#r1187830802

I _think_ that's everything... please let me know. Thank you!

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
+                  parseXml(description),
+                  'ProfileDescription',
+                  description
+                ) as string
             },
             descriptions: [
               {
                 data: {
-                  path: 'Rule.check.check-content-ref.name',
+                  path: ['check.check-content-ref.name'],
```
Amndeep7, line 391: image

check text and references are falling back to the correspondingly named tags when they are left as empty string / empty array.  this causes [Object] to show up in heimdall when looking at the details tab.

please update descriptions.check and the refs array to contain this information instead.  refs array will probably be pretty easy - just take the data that looks like this (
image
) and put each one of those objects as that 'ref' attribute (https://github.com/mitre/heimdall2/blob/master/libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts#L245).  check text will be harder (
image
) - easiest thing to do that i'm fine with is just running a JSON.stringify and dumping it out.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-                  profiles: Record<string, unknown>[]
-                ): Record<ProfileKey, unknown>[] => {
-                  const pathsSelect = ['select'];
-                  const paths = {
-                    id: ['id'],
-                    description: ['description.text', 'description'],
-                    title: ['title.text', 'title']
-                  };
-                  return getProfiles(profiles, pathsSelect, paths);
+              ident: {path: 'ident'},
```
Amndeep7, line 341: image

seems like there are more attributes that don't exist in the output coming from complianceascode.  we should probably toss those in the transformer as a conditionally provided attribute (see the version example).

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
                 transformer: extractCci
               },
               nist: {
-                path: 'Rule.ident',
+                path: ['ident', 'reference'],
                 transformer: nistTag
               },
-              severity: {path: 'Rule.severity'},
+              severity: {path: 'severity'},
               description: {
```
Amndeep7, line 315: image

requires more work to handle the embedded html tags in there - "theso" came from "place a file in the <html blob> so that it" but the html blob and surrounding whitespace was removed for some reason

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
             tags: {
               cci: {
-                path: 'Rule.ident',
+                path: ['ident', 'reference'],
                 transformer: extractCci
               },
               nist: {
-                path: 'Rule.ident',
+                path: ['ident', 'reference'],
                 transformer: nistTag
```
Amndeep7, line 312: not producing correct output at times.  see that same "crypto policies are means of" rule.  
image
image
the nist tag list should consist of those 5

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-                  path: 'Rule.check.check-content-ref.name',
+                  path: ['check.check-content-ref.name'],
                   transformer: parseHtml
                 },
                 label: 'check'
               },
               {
                 data: {
-                  path: 'Rule.fixtext.text',
+                  path: ['fixtext', 'fix'],
```
Amndeep7, line 398: image
fix text not being found

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: feel like it's cause we only look for cci's and then the corresponding nist tag.  if there are no cci's provided then we stuff the default nist tags in there (sa-11/ra-5) as a (usually correct but probably not this time) reasonable enough default.

however, if the references list provides actual nist tags as well then we should use those in addition to / instead of the cci ones.

another example of this happening is 'the installed operating system is fips 140-2 certified".  there are two cci's identified from the references list (
image
) and so we get 2 nist tags, but the references section itself includes a hell of a lot more (
image
)

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-              }
+              path: ['description.text', 'description'],
+              transformer: (description: string): string =>
+                _.get(
+                  parseXml(description),
+                  'ProfileDescription',
+                  description
+                ) as string
             },
             descriptions: [
```
Amndeep7, line 388: image

rationale should probably be added as a conditional subdescription

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
         status: 'loaded',
         controls: [
           {
-            path: 'Benchmark.Group',
+            path: 'Benchmark',
+            pathTransform: getRulesInBenchmark,
             key: 'id',
             tags: {
```
Amndeep7, line 305: image

warning should be added as a conditionally supplied tag or subdescription

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
+                transformer: (description: string): string =>
+                  _.get(
+                    parseXml(description),
+                    'VulnDiscussion',
+                    description
+                  ) as string
               },
-              group_id: {path: 'id'},
-              group_title: {path: ['title.text', 'title']},
+              group_id: {path: 'group.id'},
```
Amndeep7, line 324: image
image

doubly nested groups only seem to show the immediate parent group's information.  i feel like all levels of group information should be shown here, but am not sure how best to reflect that information such that it shows up well within heimdall.  the also showcases other problems like how certain group level attributes don't get surfaced properly such as that 'value' block.  probably worthwhile setting up a brief call with you/me/aaron/eugene to figure out how to handle this one.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,149 @@ export class XCCDFResultsMapper extends BaseConverter {
...
               },
               value: {
-                path: '$.Benchmark.Value',
-                transformer: (values: Record<string, unknown>[]): unknown => {
-                  return _.find(values, (value: Record<string, unknown>) => {
-                    const id = _.get(value, 'id');
-                    return id && id === valueIdTracker;
-                  });
-                }
+                path: ['values']
```
Amndeep7, line 365: this seems to get highly borked at times

for "configure auditd max_log_file_action upon reaching maximum log size"

image

image

Seems like it's extracting out the 'br' stuff from being embedded in the text content and putting it separate whereas the original document is using it as html-encoded formatting :(

Amndeep7 (changes requested):

libs/hdf-converters/sample_jsons/xccdf_results_mapper/xccdf-openscap-ComplianceAsCode-ubuntu1804-hdf.json
```diff
@@ -32860,13 +32776,16 @@
                 "text": "SRG-OS-000059-VMM-000280",
                 "href": ""
               },
+              {
```
Amndeep7: example of a validate duplicate showing up in the hdf file cause the original sample file has it duplicated as well for some reason

image

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,155 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-                  path: 'Rule.check.check-content-ref.name',
+                  path: ['check.check-content-ref.name'],
                   transformer: parseHtml
                 },
                 label: 'check'
               },
               {
                 data: {
-                  path: 'Rule.fixtext.text',
+                  path: ['fixtext', 'fix'],
```
Amndeep7, line 398: image

example of multiple fix texts existing that i don't think the code would be able to handle even if the path were correctly pointing to it

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -392,8 +392,21 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-              path: ['severity'],
-              transformer: impactMapping(IMPACT_MAPPING)
+              transformer: (vulnerability: Record<string, unknown>): number => {
+                const ruleResult = _.get(vulnerability, 'ruleResult') as Record<
+                  string,
+                  unknown
+                >;
+                if (ruleResult) {
+                  const result = _.get(ruleResult, 'result') as string;
+                  if (result == 'notapplicable' || result == 'informational') {
```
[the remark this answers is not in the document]
candrews: :+1: Done

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: making this a separate issue to figure out a good solution for multiply-nested groups.  for now, only showing the parent group is sufficient.

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: whereas the ref/reference stuff is not overlapped, the check subdescription and check tag do have overlap where if the subdescription is available it hides the tag.  for now, we have stringified the tag so it looks good in heimdall, but we should go and make the subdescription find things properly and behave well with the original data being an array for example.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -278,142 +299,155 @@ export class XCCDFResultsMapper extends BaseConverter {
...
             },
             results: [
               {
                 status: {
-                  path: '$.Benchmark.TestResult',
+                  path: ['ruleResult.result'],
                   transformer: getStatus
                 },
-                code_desc: '',
+                code_desc: {
```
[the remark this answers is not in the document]
candrews: This problem appears to be caused by shortcoming the XML parser used by this project: https://github.com/NaturalIntelligence/fast-xml-parser/issues/558

I don't know what, if anything, I can do about this one... any suggestions?

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: I believe I have now address this point.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: This problem appears to be caused by shortcoming the XML parser used by this project: https://github.com/NaturalIntelligence/fast-xml-parser/issues/558

I don't know what, if anything, I can do about this one... any suggestions?

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: I believe I have now addressed this point.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: I believe I have now addressed this point.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: This problem appears to be caused by shortcoming the XML parser used by this project: https://github.com/NaturalIntelligence/fast-xml-parser/issues/558

I don't know what, if anything, I can do about this one... any suggestions?

candrews:
@Amndeep7 I believe that I have addressed all the feedback.

As you know, I'm not terribly familiar with the heimdall tool, but with that disclaimer, I think the output looks good and is definitely improvement over when we last looked at this together.

Can you please take another look and let me know what else needs to be done?

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: Yeah, we rewrite it using JSONIX which hopefully just leaves random embedded blocks as they are.  So long as the with-raw option is provided, this isn't a blocker cause the user could at the end of the day retrieve this information.  I'm not gonna make this a blocker.

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: not considered a blocker - same as the other fast-xml-parser comment

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -375,7 +421,19 @@ export class XCCDFResultsMapper extends BaseConverter {
                 )
               })
             },
-            refs: [],
+            refs: [
```
Amndeep7, line 424: from xccdf-scc-rhel8-hdf.json

image

your sample outputs don't match the schema for some reason

https://github.com/mitre/heimdall2/blob/master/libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts#L244

should probably look something like `refs: [ref: [{publisher: blah, ...}]]`

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -392,16 +450,34 @@ export class XCCDFResultsMapper extends BaseConverter {
...
                 },
                 label: 'check'
               },
               {
                 data: {
-                  path: ['fixtext', 'fix'],
-                  transformer: parseHtml
+                  path: ['fixtext.text', 'fixtext', 'fix.text', 'fix'],
+                  transformer: (text: string | string[]) =>
+                    asArray(text).map(parseHtml).join('\n') || undefined
```
Amndeep7, line 462: the '|| undefined' is dangerous cause it makes it not match the schema

xccdf-scc-rhel8-hdf.json

image

schema: https://github.com/mitre/heimdall2/blob/master/libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts#L224

if a description is provided it needs to have both a label and data.  the or defined makes it so that sometimes only a label is provided.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -151,11 +160,18 @@ function extractCci(input: IIdent | IIdent[]): string[] {
...
+    extractCci(input),
     DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS,
     false
+  ).concat(
+    asArray(input)
+      .filter((x) => !!x)
+      .map((x) => x.text)
+      .map(parse_nist)
+      .filter((x) => !!x)
+      .filter(is_control)
```
Amndeep7, line 173: seems like your sample is not updated properly still?  or this function is borked lol

xccdf-scc-rhel7-hdf.json

image

seems like the nist tag array is including the stig control ids, which it shouldn't do

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7, line 1: none of the original or hdf samples open up in heimdall i think cause of the schema issues

attempt at raw file:  
image

attempt with generated hdf file:  
image

this is it complaining about the controldescription type not being matched:
image

Amndeep7 (changes requested):
Feeling like this should probably be able to be merged in after this final set of changes is made.  Thanks for bearing with the process for so long.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: That data structure is a bit confusing... but I think I matched the intent now.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: Updated! :+1:

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: `is_control(parse_nist('SV-86515'))` return true... so as far as inspectjs is concerned, 'SV-86515' is a nist control and therefore should be included.

Am I misunderstanding `nist.js`? How should I determine if an item should be included or excluded from this list?

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: I tested a bunch and they load now.

If you have one that doesn't, please give me the name so I can take a look.

candrews:
Using `stopNodes` would be a great way to fix the parsing problem that occurs with nodes that contain HTML.

However... `stopNodes` doesn't work for XML documents that use namespaces: https://github.com/NaturalIntelligence/fast-xml-parser/blob/v4.2.4/src/xmlparser/OrderedObjParser.js#L308 Attempting to use it results in an "Unexpected end of <tag>" error.

If this issue in the xml parser is fixed, then modifying `xccdf-results-mapper.ts` to say:
```javascript
  constructor(scapXml: string, withRaw = false) {
    super(
      parseXml(scapXml, {
        stopNodes: [
          '*.fixtext',
          '*.fix',
          '*.rationale',
          '*.warning',
          '*.title',
          '*.description'
        ]
      })
    );
```

would solve the html parsing problem.

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -481,37 +481,41 @@ export class XCCDFResultsMapper extends BaseConverter {
...
-                label: 'check'
-              },
+                path: ['check.check-content-ref.name'],
+                transformer: (
+                  data: string | string[]
+                ): ExecJSON.ControlDescription => ({
+                  data: asArray(data).join('\n'),
+                  label: 'check'
+                })
+              } as any,
```
Amndeep7, line 491: would prefer to avoid using 'any' if possible?  how does it complain if you don't cast it?

if we can't make it happy by clearing up the types directly, then i usually prefer doing 'as unknown as whatever_it_ought_to_be' over 'as any'

Amndeep7 (changes requested):

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: this feels like a failure on inspecjs' part in not removing things that are clearly wrong. parse_nist should throw errors/explicitly fail when they are not nist controls/control families. it not doing so confuses the type system by mistakenly typing that string. is_control more or less is only used to narrow down types and doesn't do much actual validation.

anyways as shown there and in 'raw_nist.ts', nist controls look like 'two_letter_family-small_number possibly followed by letters and numbers in parentheses' whereas stig ids are 'S?V-5__or_6_digit_number'.

the issue that arose here is that it thinks 'sv' is a control family and then processes the numbers after it as like 'control "big number" from within that control family' but doesn't take into account the reality of the situation where 'sv' is not a valid control family nor is there a control with an index as high as '86515' in existence.

looking briefly, it seems like the regexes that parse_nist uses are vastly broader than the actually allowed set of inputs. i will experiment to see if it's possible to restrict it to allowed control families.

libs/inspecjs/src/nist.ts
```diff
@@ -2,9 +2,10 @@ import {ControlStatus} from './compat_wrappers';
 import {ALL_NIST_CONTROL_NUMBERS, ALL_NIST_FAMILIES} from './raw_nist';
 
 // Regexes.
-const NIST_FAMILY_RE = /^[A-Z]{2}$/;
+// Matches against only currently existing NIST Control families instead of allowing any two adjacent uppercase letters
```
Amndeep7, line 5: image

seems like it is no longer falsely validating sv numbers as controls but

image

these undefineds are getting passed through which ain't good cause they're not valid json.  idk where they're coming from tho - don't *think* it should have anything to do with the change that i made

libs/inspecjs/src/nist.ts
```diff
@@ -2,9 +2,10 @@ import {ControlStatus} from './compat_wrappers';
...
+const NIST_FAMILY_RE = /^(?:A[CPRTU]|C[AMP]|D[IM]|I[APR]|M[AP]|P[ELMS]|RA|S[ACEI]|TR|U[LM])$/;
```
Amndeep7, line 6: would appreciate you pleasing the linter on my behalf and also regenerating the samples

libs/hdf-converters/src/xccdf-results-mapper.ts
```diff
@@ -164,34 +145,93 @@ interface IIdent {
...
-    if (text.match(CCI_REGEX)) {
+    if (!!text && CCI_REGEX.exec(text)) {
       output.push(text);
     }
-  });
+  }
   return output;
 }
 
 function nistTag(input: IIdent | IIdent[]): string[] {
```
Amndeep7, line 162: image

I think the function in that utility file is expecting the nist tag to only be ex. 'SI-2' whereas we are passing in ex. 'SI-2 (b)'.  Considering that the "800-53 Controls & CCIs" column in the table can show the full thing, I think this is a bug with the utility file where it's not dealing with the full string properly.  If you have the time to debug it, I'd appreciate it, otherwise I'll get to it myself in the next couple days.

Amndeep7 (changes requested):

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: The type of `descriptions` is `ControlDescription` and that has two mandatory fields: `data` and `label`. Neither of those fields are directly specified (they are set by the transformer, which typescript has no understanding of).

I implemented the `as unknown as ControlDescription` approach.

libs/inspecjs/src/nist.ts
candrews: Done! The linting issues have been addressed and samples regenerated.

libs/inspecjs/src/nist.ts
candrews: I don't see `undefined` in any of the `libs/hdf-converters/sample_jsons/xccdf_results_mapper/*.json` files. Can you link me to where I can see it or provide the file name that I can look at to see this problem?

Amndeep7: I got these by doing the yarn run test.  you can see them in the logs on github as well.

libs/hdf-converters/src/xccdf-results-mapper.ts
candrews: Have you had a chance to address this issue?

libs/inspecjs/src/nist.ts
candrews: I don't see this in any of the github logs, nor in the output of `yarn test`, nor or any of the files.

Is it possible you're looking at an old build or old code?

If you can find it, can you please link me to it, or give me the path of a file and line number to took at?

candrews:
@Amndeep7 do you have a guess as to when we'll be able to progress this PR?

Amndeep7:
Sorry I've been swamped this week.  I'll try to take a look next week to fix that bug in Heimdall that I mentioned.

libs/hdf-converters/src/xccdf-results-mapper.ts
Amndeep7: Turns out that it was not a problem with the specification but with the zero padding that was causing issues.  Anyways, I've made the fix.  I will merge if there are no other issues that arise during the CI.

libs/inspecjs/src/nist.ts
Amndeep7: Realized what was happening: the test is comparing the output of the file (JSON) with the output from running the toHdf function (JS).  consequently it sees the 'attribute: unknown' in there cause that's still valid in js (and is functionally a no-op) but it gets removed when turned into json.  when you get rid of any other differences in the files (ex. the differently formatted / duplicate nist tags), then it seems like jest is smart enough to realize that they are equal and doesn't surface anything into the console.

Amndeep7 (approved):

Amndeep7:
Thank you for your patience and hard work on this PR @candrews.  You did some absolutely fantastic work here that'll be appreciated across many of the users of Heimdall.

PR #4285, opened 2023-03-29
Update dockerfile to:

Amndeep7:
1) Use ubi8 as the base instead of alpine
2) Ideally resolve the malformed dependencies issue by including the
   tsconfigs as well as the package jsons before building
3) Specify a default value for NODE_ENV
4) Get rid of pointless permissions setting (esp in the builder)
5) Some stylistic improvements


Amndeep7:
TODO: apply similar changes to dockerfile.lite
TODO: re-add arm support for that one workflow

PR #4383, opened 2023-04-24
Added variable in store/config to track if api keys are enabled

Rlin232:


apps/frontend/src/components/global/UserModal.vue
```diff
@@ -370,22 +373,24 @@ export default class UserModal extends Vue {
...
+      this.apiKeyTableLoading = true;
+      axios
+        .create()
+        .get<IApiKey[]>(`/apikeys`, {params: {userId: this.user.id}})
+        .then(({data}) => {
+          this.apiKeys = data;
+        })
+        .catch((error) => {
+          if (error.response) {
+            if (error.response.status === 403) {
```
Amndeep7, line 386: we should make the snackbar thing pop up regardless of which particular error response code we get (i.e. get rid of this nested if, the top level one is sufficient)

apps/frontend/src/components/global/UserModal.vue
```diff
@@ -370,22 +373,24 @@ export default class UserModal extends Vue {
...
+              SnackbarModule.failure('Unable to get API Key');
```
Amndeep7, line 387:
```suggestion
              SnackbarModule.failure('Unable to get API Keys');
```

Amndeep7 (changes requested):

apps/frontend/src/components/global/UserModal.vue
Amndeep7: fixed

Amndeep7 (approved):

PR #4403, opened 2023-04-26
Reassign group owners upon user deletion

Rlin232:


Amndeep7 (changes requested):

apps/backend/src/groups/groups.service.ts
```diff
@@ -58,6 +54,26 @@ export class GroupsService {
     const groupUser = await GroupUser.findOne({
       where: {groupId: group.id, userId: updateGroupUser.userId}
     });
+    // If there are no more owners, set an admin to owner
+    const owners = (await group.$get('users')).filter(
```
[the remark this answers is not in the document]
Rlin232: See e457d32d9d03a48e2a499e759d26ea96faf081cb

Rlin232:
The current failing unit test is regarding the groups service. The test expects a ForbiddenException as per the previous default behavior for trying to remove the last owner of a group. This PR changes this default behavior (instead promoting the default admin, if possible, to the new group owner). Thus, the test should be re-written to reflect this change.

apps/backend/src/groups/groups.service.ts
```diff
@@ -58,19 +61,47 @@ export class GroupsService {
...
+    }
     return groupUser?.update({role: updateGroupUser.groupRole});
   }
 
   async removeUserFromGroup(group: Group, user: User): Promise<Group> {
+    await this.setDefaultToOwner(group, user.id);
+    return group.$remove('user', user);
+  }
+
+  async setDefaultToOwner(group: Group, id: string) {
```
Amndeep7, line 75: this function name is misleading.  default what?  why is ownership being changed on every user role modification / user removal?

i would do something closer to 'ensureGroupHasOwner' which actually mentions the relevant verb (ensuring that something is the case and making changes to have it be so if necessary), mentions the relevant noun (the group), and the relevant condition (groups must have owners).  only thing we're missing is the fact that an admin becomes the owner by default?  atm i'm not sure if this counts as an implementation detail or something worthwhile surfacing to the function name.

alternatively we could separate out the conditional from the work being done so that there would be a 'isLastOwner' function and then a 'setDefaultOwnerForGroup' function that might take an argument for which user to assign ownership to by default and then as fallback goes up the admin chain.

https://softwareengineering.stackexchange.com/questions/161753/naming-a-do-x-if-needed-method
https://softwareengineering.stackexchange.com/questions/365310/how-to-name-a-method-which-may-or-may-not-perform-an-action-depending-on-a-condi

^ interesting material to read through.  might be worthwhile having a go-back during standup or as a question posted in dev-chat on what best practices we should follow

Amndeep7, line 75: 'id' of what?  i would do 'userId' here if forced to only have the attribute, but it seems like the current practice is to pass through these model objects in their entirety and then use attributes from them as necessary, so i would actually probably make this 'user' and then change the conditional on line 80 to get the attribute

apps/backend/src/users/users.service.ts
```diff
@@ -140,6 +142,16 @@ export class UsersService {
         'Cannot destroy only administrator account, please promote another user to administrator first'
       );
     }
+    // Clean up groups owned by user
+    await Promise.all(
```
Amndeep7, line 146: https://www.techiediaries.com/promise-all-map-async-await-example/

example basically doing exactly what we're doing here.  no change needed in the code, just wanted to show you

apps/backend/src/users/users.service.ts
```diff
@@ -140,6 +142,16 @@ export class UsersService {
         'Cannot destroy only administrator account, please promote another user to administrator first'
       );
     }
+    // Clean up groups owned by user
```
Amndeep7, line 145: can you simplify the code ni lines 121-136?  there's no point doing a try-catch to catch an exception that you yourself are throwing.

i would probably do something like:

`if(abac.cannot && !(await compare)) { throw new ForbiddenException('your password was wrong') }`

apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -240,11 +240,17 @@ export default class GroupModal extends Vue {
...
-    ]);
+    const removedUserPromises = toRemove.map((user) => {
+      const removeUserDto: IRemoveUserFromGroup = {
+        userId: user.id
+      };
+      return axios.delete(`/groups/${group.id}/user`, {data: removeUserDto});
+    });
+    return Promise.all([...addedUserPromises, ...updatedUserPromises]).then(
+      () => {
+        return removedUserPromises;
```
Amndeep7, line 251: removedUserPromises is an array of promises.  this function being wrapped in a 'then' means that the function's return value is immediately a promise, so doing an await against it will return the array of promises that themselves haven't been awaited, i.e. line 191 isn't doing anything worthwhile.  this line should probably be `return Promise.all(removedUserPromises);`

async is hard to do right.  make sure to walk through the logic every time.

apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -173,15 +173,22 @@ export default class Users extends Vue {
...
   }
 
-  deleteUserConfirm(): void {
+  @Emit()
+  deleteUserConfirm(): boolean {
+    let saveable = true;
+    const userToDelete = this.currentUsers.indexOf(this.getEditedUser());
+    if (
+      this.currentUsers[userToDelete].groupRole === 'owner' &&
+      this.numberOfOwners() <= 1
```
Amndeep7, line 182: we've been using `< 2` everywhere else though I agree that `<= 1` is clearer but let's keep it consistent with the `< 2` unless you wanna go and fix it everywhere

apps/frontend/src/components/global/groups/Users.vue
```diff
@@ -173,15 +173,22 @@ export default class Users extends Vue {
...
+    ) {
+      saveable = false;
+    }
     if (this.editedUserID !== '0') {
```
Amndeep7, line 186: why did we choose '0' as the default id?  isn't it possible for the default admin to be an edited user, and doesn't the default admin have an id of 0?

yarn.lock
Amndeep7, line 1: unless you're intentionally wanting to add/update dependencies, I would probs not do it in this pr esp since that is what causes some pointless merge conflicts

Amndeep7 (changes requested):

apps/frontend/src/components/global/groups/Users.vue
Rlin232: No, the default admin has an id of 1.

apps/backend/src/users/users.service.ts
Rlin232: Done

Amndeep7 (approved):

PR #4410, opened 2023-04-27
Tenable Frontend

georgedias:
What is provided by this PR
 - [x] Add the capability of querying a Tenable.SC instances, display available scans for user selection into Heimdall for visualization
 - [x] Notify user when selecting an invalid file (non json) from a S3 bucket
 - [x] Display `Heimdall Server` on the load modal when in server mode


apps/frontend/src/components/global/upload_tabs/FileReader.vue
```diff
@@ -27,6 +27,7 @@
             <li>Static Analysis Results Interchange Format (SARIF)</li>
             <li>Scoutsuite</li>
             <li>Snyk</li>
+            <li>Tenable Export (nessus)</li>
```
ejaronne, line 30: Change "Tenable Export (nessus)" to "Tenable (Nessus)". Export is confusing because the user might confuse that with the different Heimdall export options and we don't need to get into all of that at this time.

georgedias: Should I use Tenable.SC (Nessus)

apps/frontend/src/components/global/upload_tabs/FileReader.vue
```diff
@@ -27,6 +27,7 @@
             <li>Static Analysis Results Interchange Format (SARIF)</li>
             <li>Scoutsuite</li>
             <li>Snyk</li>
+            <li>Tenable (Nessus)</li>
```
aaronlippold, line 30: Tenable.sc API?

georgedias: @ejaronne suggested that we do not use the .sc here

apps/frontend/src/components/global/HelpModal.vue
```diff
@@ -70,12 +70,18 @@
 
 <script lang="ts">
 import {AppInfoModule} from '@/store/app_info';
-import Vue from 'vue';
-import Component from 'vue-class-component';
+import ServerMixin from '@/mixins/ServerMixin';
+import Component, {mixins} from 'vue-class-component';
 
 @Component
-export default class HelpModal extends Vue {
```
Hookwitz, line 77: Not sure if this was supposed to be renamed to AboutModal as the HelpModal is used in TopbarDropdown and HelpFooter so these could break.

Hookwitz (changes requested):

apps/frontend/src/components/global/HelpModal.vue
georgedias: Too funny, done

Hookwitz (approved):

apps/frontend/src/components/global/upload_tabs/aws/FileList.vue
```diff
@@ -78,6 +79,14 @@ export default class FileList extends Vue {
     // Fetch it from s3, and promise to submit it to be loaded afterwards
     await fetch_s3_file(this.auth.creds, file.Key!, this.formBucketName).then(
       (content) => {
+        try {
+          JSON.parse(content);
```
Amndeep7, line 83: I don't think we should add this restriction such that only JSON files can be processed from S3 buckets.  It'd be nice to support any filetype (ex. any of the XML based results files) that could be stored in a bucket and then converted on load.

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
```diff
@@ -49,6 +49,31 @@
             <v-icon pr-2>mdi-github-circle</v-icon>
             Splunk Configuration
           </v-btn>
+          <br />
+          <span>
+            The Splunk platform accepts any type of data. In particular, it
+            works with all IT streaming
```
Amndeep7, line 55: what is IT streaming?

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
```diff
@@ -49,6 +49,31 @@
...
+            <br />
```
Amndeep7, line 56: why are we manually putting breaks in the text?  the component should be able to handle word wrapping responsively

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
```diff
@@ -49,6 +49,31 @@
...
+          </span>
+          <br />
+          <span>
+            <br />
+            "A Splunk index is a repository for Splunk data."
+            <br />
+            Data that has not been previously added to Splunk is referred to as
+            raw data.
+            <br />
+            When the data is added to Splunk, it indexes the data (uses the data
```
Amndeep7, line 71: the parenthetical clause doesn't add any new information/clarification.

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
```diff
@@ -49,6 +49,31 @@
...
+            to update its indexes),
+            <br />
+            creating event data. Individual units of this data are called
```
Amndeep7, line 74:
```suggestion
            consequently creating event data. Individual units of this data are called
```

apps/frontend/src/components/global/upload_tabs/tenable/AuthStep.vue
```diff
@@ -0,0 +1,111 @@
...
+  accesskey = '';
+  secretkey = '';
+  hostname = '';
+
+  /** Form required field rules. Maybe eventually expand to other stuff */
+  reqRule = (v: string | null | undefined) =>
+    (v || '').trim().length > 0 || 'Field is Required';
+
+  async login(): Promise<void> {
+    if (!/^https?:\/\//.test(this.hostname)) {
```
Amndeep7, line 77: this is checking to see if the url starts with https right?  what happens if it's on http for some reason, like it's on some purely internal server where they didn't feel the need to set up certs and stuff?

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
```diff
@@ -0,0 +1,262 @@
+<template>
+  <span>
+    <div class="d-flex flex-row-reverse" style="cursor: pointer">
+      <v-btn icon @click="updateSearch">
+        <v-icon
+          b-tooltip.hover
```
Amndeep7, line 6: isn't this b-tooltip stuff like bootstrap specific functionality?  we're not using bootstrap here

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
```diff
@@ -0,0 +1,262 @@
...
+  ];
+
+  async updateSearch() {
+    this.loading = true;
+
+    // Get todays date
+    const dateNow = new Date();
+
+    // Convert date to epoch number value
+    const endTime = Number(dateNow.getTime().toString().substring(0, 10));
```
Amndeep7, line 151: what is the point of doing the substring?  looks like it's cutting off the last three digits of time *for now* so is the intent just to get the epoch time but at second resolution?  if so, then we can do something like: `Math.trunc(dateNow.getTime()/1000)` and avoid having to do the cast to string, latent bug string manipulation, and then recast back into number.

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
```diff
@@ -0,0 +1,262 @@
...
+
+    // Set the start time to midnight (start of the day), converting to epoch
+    dateNow.setHours(0, 0, 0, 0);
+    const startTime = Number(
+      dateNow
+        .setDate(dateNow.getDate() - this.scanDays)
+        .toString()
+        .substring(0, 10)
```
Amndeep7, line 159: same as above here

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
```diff
@@ -0,0 +1,262 @@
...
+    this.$emit('got-files', files);
+  }
+
+  async mounted() {
+    await this.updateSearch();
+  }
+
+  epochToDate(date: string): string {
+    return Number(date) !== -1
+      ? new Date(Number(date) * 1000).toLocaleString('en-US', {
```
Amndeep7, line 245: is the reason we're supplying a specific locale due to needing to supply the options as well?  is it possible to provide like an empty string/undefined/null/whatever such that it falls back to default behavior (i.e. uses the host's locale)?

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
```diff
@@ -0,0 +1,262 @@
...
+          timeZone: 'UTC',
+          hour12: false
+        })
+      : '';
+  }
+
+  formatNumberOfScans(value: string | undefined): string {
+    return Number(value) !== -1
+      ? Number(value).toLocaleString('en-US').toString()
```
Amndeep7, line 254: same as above

apps/frontend/src/components/global/upload_tabs/tenable/TenableReader.vue
```diff
@@ -0,0 +1,102 @@
...
+          :tenable-config="tenableConfig"
+          @signOut="onSignOut"
+          @got-files="got_files"
+        />
+      </v-stepper-content>
+    </v-stepper-items>
+    <v-overlay
+      :opacity="50"
+      absolute="absolute"
+      :value="errorCount >= 3 || errorCount < 0"
```
Amndeep7, line 32: why is it possible for the errorcount to be less than 0?

why are we checking for if it's greater than or equal to 3 here, but on line 36 for the v-if on that span we only check if it's greater than 0?

apps/frontend/src/components/global/upload_tabs/tenable/TenableReader.vue
```diff
@@ -0,0 +1,102 @@
...
+    >
+      <div class="text-center">
+        <p>
+          <span v-if="errorCount > 0" style="color: red; font-weight: bold">
+            It seems you may be having trouble connecting to Tenable.sc. Ensure
+            <br />
+            the following are properly configured: access and secrete key, URL?
```
Amndeep7, line 39:
```suggestion
            that the access key, secret key, and host url are properly configured.
```

apps/frontend/src/components/global/upload_tabs/tenable/TenableReader.vue
```diff
@@ -0,0 +1,102 @@
...
+    <v-overlay
+      :opacity="50"
+      absolute="absolute"
+      :value="errorCount >= 3 || errorCount < 0"
+    >
+      <div class="text-center">
+        <p>
+          <span v-if="errorCount > 0" style="color: red; font-weight: bold">
+            It seems you may be having trouble connecting to Tenable.sc. Ensure
+            <br />
```
Amndeep7, line 38: same as above w/r to the breaks

apps/frontend/src/components/global/upload_tabs/tenable/TenableReader.vue
```diff
@@ -0,0 +1,102 @@
...
+            px-0
+          >
+            <v-icon pr-2>mdi-github-circle</v-icon>
+            Tenable Configuration
+          </v-btn>
+          <br />
+          <span>
+            API key authorization requires Tenable.sc 5.13.x or later.
+            <br />
+            Unique set of API keys can be generated for each user account.
```
Amndeep7, line 59:
```suggestion
            A unique set of API keys can be generated for each user account.
```

apps/frontend/src/utilities/export_util.ts
```diff
@@ -1,6 +1,5 @@
-import concat from 'concat-stream';
+import ZipFile from 'adm-zip';
```
Amndeep7, line 1: why did we need to change zip libraries?

Amndeep7 (changes requested):

apps/frontend/src/components/global/upload_tabs/aws/FileList.vue
georgedias: Currently the code only processes JSON files, once we lift that we can change it here. The problem is that if you have any other format in the S#, there isn't any notification to the user of any problems, and the file doesn't load.

apps/frontend/src/utilities/export_util.ts
georgedias: Because the previous packages did not provide unzipping in memory.

apps/frontend/src/components/global/upload_tabs/tenable/TenableReader.vue
georgedias: so we can display the error span tag

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
georgedias: It is when you create a pipe to an endpoint that continuously pushes its output.

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
georgedias: So the linter wont bark

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
georgedias: yes it does, it informs the user how indexes are created.

apps/frontend/src/components/global/upload_tabs/splunk/SplunkReader.vue
georgedias: the word consequently is not needed.

apps/frontend/src/components/global/upload_tabs/tenable/AuthStep.vue
georgedias: we should not have an unsecure instances running.

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
georgedias: it works

apps/frontend/src/components/global/upload_tabs/tenable/FileList.vue
georgedias: yes, if you check epoch values are based on 10 digital numeric values, the getTime, returns eccessive digits (3) that when passed to the endpoint will not match the date time format on the server. The implemented solution is valid, no need to change it.

georgedias (commented):
made changes base on requests

README.md
```diff
@@ -268,6 +266,20 @@ In order to allow Heimdall to Connect to your AWS S3 bucket, you need to [add a
 ]
 ```
 
+### Tenable.SC
+
+In order to allow Heimdall to Connect to a Tenable.SC instance, the hosting services should be configured with an allowlist including the calling domain (where Heimdall resides) is a trusted domain to perform CORS requests.
```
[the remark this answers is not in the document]
georgedias: update with proposed content

PR #4480, opened 2023-05-16
Don't handle every array item within each array item

candrews:
In handleArray, the array v is looped over. For each item, the entire array is looped over again, resulting in an array v of length n resulting in an output array of length n^2, making handleArray's complexity O(n^2).

However, that n^2 looping is unnecessary. Removing it brings the complexity of handleArray down to O(n), drastically reducing execution time.


candrews:
@Amndeep7 as discussed, I discovered this problem while working on https://github.com/mitre/heimdall2/pull/4255 but since it's unrelated to that effort I've created a new PR for this change.

Amndeep7 (approved):
Excellent catch.  Extended the implication from where you found savings with my own commits.

PR #4516, opened 2023-05-24
Conveyor mapper

jweissm:


aaronlippold:
@Amndeep7 @em-c-rod do we have a status on this PR? Are we waiting on input from the author?

libs/hdf-converters/src/conveyor-mapper/conveyor-mapper.ts
em-c-rod, line 1: Did you mean to commit this file?

em-c-rod:
Simple fix - can you add this to the list in the README? https://github.com/mitre/heimdall2/blob/master/libs/hdf-converters/README.md

libs/hdf-converters/test/mappers/forward/conveyor_mapper.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('conveyor_mapper', () => {
+  it('Successfully converts Conveyor reports', () => {
+    const mapper = new ConveyorResults(
+      fs.readFileSync(
+        'sample_jsons/conveyor_mapper/sample_input_report/large-results.json',
+        {encoding: 'utf-8'}
+      )
+    );
+
+    expect(omitVersions(mapper.toHdf()['Moldy'])).toEqual(
```
em-c-rod, line 13: Do you have a check for Stigma?

em-c-rod:
Are you able to see the pipeline results? You can run `npm test` and `npm run lint` locally to check both of those and also look at the PR pipeline results to fix any of the others.

jweissm:
> @jweissm Are you able to see the pipeline results? You can run `npm test` and `npm run lint` locally to check both of those and also look at the PR pipeline results to fix any of the others.

i can see it thanks! looks like i need to handle undefined cases bettter

package-lock.json
em-c-rod, line 1: Can you delete this file, too? This repository is just storing the yarn.lock file, so we don't need this. If you use `yarn install` instead of `npm install` you can avoid generating this file locally as well, although either should work for local testing.

jweissm: got it, i deleted the file

libs/hdf-converters/sample_jsons/conveyor_mapper/conveyor-hdf.json
em-c-rod, line 1: Are you using this file?

libs/hdf-converters/sample_jsons/conveyor_mapper/empty.json
em-c-rod, line 1: Are you using this file?

libs/hdf-converters/sample_jsons/conveyor_mapper/conveyor-hdf.json
jweissm: kind of...i thought it would be good to have an example of the full result(not just moldy or stigma)

em-c-rod: Does it make sense to add a test for that case, then?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,211 @@
...
+      platform: {
+        name: 'Heimdall Tools',
+        release: HeimdallToolsVersion
+      },
+      version: HeimdallToolsVersion,
+      statistics: {},
+      profiles: [
+        {
+          name: this.type,
+          version: 'na',
```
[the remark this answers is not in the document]
jweissm: ok talked to conveyor, added what should suffice for a version number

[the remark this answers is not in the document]
jweissm: ok version added correctly now, have what the conveyor team thinks works

libs/hdf-converters/README.md
```diff
@@ -8,23 +8,24 @@ HDF Converters supplies several methods to convert various types of security too
...
-12. [**sarif-mapper**] - SARIF JSON file
-13. [**scoutsuite-mapper**] - ScoutSuite results from a Javascript object
-14. [**snyk-mapper**] - Snyk results JSON file
-15. [**sonarqube-mapper**] - SonarQube vulnerabilities for the specified project name and optional branch or pull/merge request ID name from an API
-16. [**splunk-mapper**] - Splunk instance
-17. [**twistlock-mapper**] - Twistlock CLI output file
-18. [**veracode-mapper**] - Veracode Scan Results XML file
-19. [**xccdf-results-mapper**] - SCAP client XCCDF-Results XML report
-20. [**zap-mapper**] - OWASP ZAP results JSON
+4.  [**conveyor-mapper**] -- Conveyor JSON file
```
Amndeep7, line 11: single hyphen

libs/hdf-converters/README.md
```diff
@@ -8,23 +8,24 @@ HDF Converters supplies several methods to convert various types of security too
...
+5.  [**dbprotect-mapper**] - DBProtect report in "Check Results Details" XML format
+6.  [**fortify-mapper**] - Fortify results FVDL file
+7.  [**ionchannel-mapper**] - SBOM data from Ion Channel
+8.  [**jfrog-xray-mapper**] - JFrog Xray results JSON file
+9.  [**nessus-mapper**] - Nessus XML results file
+10.  [**netsparker-mapper**] - Netsparker XML results file
```
Amndeep7, line 17: extra space after the 10

Amndeep7 (changes requested):

Amndeep7 (changes requested):
is there a conveyor UI (or is it just a results file)?  can we compare what results look like on there vs in heimdall?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+export class ConveyorMapper extends BaseConverter {
+  data: Record<string, unknown>;
+  type: string;
+  defaultMapping(): MappedTransform<ExecJSON.Execution, ILookupPath> {
+    return {
+      platform: {
+        name: 'Heimdall Tools',
+        release: HeimdallToolsVersion
+      },
+      version: HeimdallToolsVersion,
```
Amndeep7, line 184: make this the conveyor version

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+      statistics: {},
+      profiles: [
+        {
+          name: this.type,
+          version: {path: 'api_server_version'},
```
Amndeep7, line 189: make this the tool's version (ex. moldy)

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    title: {path: 'filename'},
+    desc: {path: 'cve_summary'},
+    impact: {
+      path: 'result.score',
+      transformer: (value) => {
+        return value / 1000;
+      }
+    },
+    refs: [],
+    tags: {
```
Amndeep7, line 158: should probably be passing through more of the data in the results file here as tags

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    },
+    source_location: {
+      ref: ''
+    },
+    results: [
+      {
+        path: 'result.sections',
+        status: {path: 'status'},
+        code_desc: {path: 'code_desc'},
+        start_time: {path: 'start_time'}
```
Amndeep7, line 169: can add run_time as well i think

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+      nist: DEFAULT_NIST_TAG
+    },
+    source_location: {
+      ref: ''
+    },
+    results: [
+      {
+        path: 'result.sections',
+        status: {path: 'status'},
+        code_desc: {path: 'code_desc'},
```
Amndeep7, line 168: maybe consider adding title_text as part of the code_desc

Amndeep7 (changes requested):

Amndeep7 (changes requested):
appreciate making a table depicting the mapping between the various conveyor fields to hdf.  we've been getting sponsor requests to showcase this information.  also useful for dev purposes in order to determine where gaps in the mapping occurred.  not everything needs a map since passthrough exists, but *enough* should be mapped to have a solid understanding of the results.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+const DEFAULT_NIST_TAG = ['SI-2', 'RA-5'];
+
+function createDescription(
+  data: Record<string, unknown>,
+  score: number,
+  date: string,
+  type: string
+): Record<string, unknown> {
+  let desc = '';
+  if (type == 'Moldy') {
```
Amndeep7, line 14: lets make an enum for all the possible tools and do comparisons against that instead of stringly typed stuff

Amndeep7, line 14: also reminder to use triple equals for comparisons

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
+import {ExecJSON} from 'inspecjs';
+import _ from 'lodash';
+import {version as HeimdallToolsVersion} from '../package.json';
+import {BaseConverter, ILookupPath, MappedTransform} from './base-converter';
+const DEFAULT_NIST_TAG = ['SI-2', 'RA-5'];
```
Amndeep7, line 5: use the DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS exported variable

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    _.set(data, 'api_response.results', conveyor);
+    super(data);
+    this.type = type;
+    this.data = data;
+    this.setMappings(this.defaultMapping());
+  }
+}
+
+export class ConveyorResults {
+  data: unknown;
```
Amndeep7, line 221: i think this could be whatever the type is for the data coming in (maybe record<string, unknown>?)

Amndeep7: ask to see if they have a schema so that we don't need to do this cause we could just use types based off of their schema instead

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  meta: Record<string, unknown>;
```
Amndeep7, line 222: since you're basing this off of the asff mapper structure, then i would combine meta into the data field unless you're having additional information come from outside the results file.  for the asff mapper, it was possible to pass in additional context that could better tune the results for some of the specialcasings, but i don't think that's what's happening here.  based off of talking with you, it's mostly metadata information from within the results file itself which we could just pass into the data field and use where necessary with the typical path/recursive structure instead of grabbing it from some arbitrary variable.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  constructor(ConveyorJson: string) {
```
Amndeep7, line 223: follow camelCase please

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    const data = JSON.parse(ConveyorJson);
+    const mapped = shafileMapper(data);
+    const newres = arrayifyObject(data, mapped);
+    this.data = newres;
+    this.meta = _.omit(data, 'api_response.results');
+  }
+
+  toHdf(): Record<string, ExecJSON.Execution> {
+    const mapped = _.mapValues(
```
Amndeep7, line 232: could be replaced by something that looks like the following.  it'll be pretty similar to what you already have here but using more native syntax instead of lodash.

toHdf(): {
return Object.entries(this.data).map(([key, val]) => return new ConveyorMapper(stuff).toHdf());
}

Amndeep7 (changes requested):

Amndeep7 (changes requested):
testing in heimdall shows that there's some typeerror (cannot read properties of udnefined reading toLowerCase on data_filters.ts:106:36) when doing free text search (the buttons all work fine).  using title:"vault.moldy" works fine but free text search for that is what is causing the failure.  imo feels like the freetext search is looking for an attribute that isn't there when it's doing its search.  unsure if this is a bug in the freetext search where we make an assumption or if it's a bug in the mapper.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+      const temp: Record<string, unknown> = _.get(value, 'children') || {};
+      _.forEach(childrenfinder(temp), function (value) {
+        arr.push(value);
+      });
+    }
+  });
+  return arr;
+}
+function shafileMapper(
+  output: Record<string, unknown>
```
Amndeep7, line 76: naming could be more accurate - the parameter passed in isn't an output

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+): Record<string, unknown> {
+  const toplevel = _.get(output, 'api_response.file_tree') as Record<
```
Amndeep7, line 78: would appreciate more camelcase usage -> topLevel, shaMappings, etc.

Amndeep7 (changes requested):

Amndeep7 (changes requested):
some more questions about the format: how is the 'score' determined within the file_tree structure?  seems like some of the inner children have scores that are not 0, but the score for the parent does not reflect that.  ex. the score for that zip in its entirety was 0 even tho it had a child in there somewhere that had a score of 100.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    string,
+    unknown
+  >;
+  const shamappings = {};
+  _.forEach(toplevel, (value) => {
```
Amndeep7, line 83: this top level code and the helper function seem pretty similar in structure.  this is just based off of a skim so it might not be possible, but it seems like it could be possible to just call the helper function directly from here instead of doing the first iteration in here and then having that iteration call the helper function.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    if (_.has(value, 'name')) {
+      const temp: string = _.get(value, 'name[0]') || '';
+      _.set(shamappings, temp, _.get(value, 'sha256'));
+    }
+    if (_.has(value, 'children')) {
+      const temp: Record<string, unknown> = _.get(value, 'children') || {};
+      const arr = childrenfinder(temp);
+      _.forEach(arr, (value) => {
+        const temp1: string = value[0] || '';
```
Amndeep7, line 92: talking with you said that this was the sha and the one below was the filename so lets continue to make variable names more useful

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  const shamappings = {};
+  _.forEach(toplevel, (value) => {
+    if (_.has(value, 'name')) {
+      const temp: string = _.get(value, 'name[0]') || '';
+      _.set(shamappings, temp, _.get(value, 'sha256'));
+    }
+    if (_.has(value, 'children')) {
+      const temp: Record<string, unknown> = _.get(value, 'children') || {};
+      const arr = childrenfinder(temp);
+      _.forEach(arr, (value) => {
```
Amndeep7, line 91: attempt to avoid variable names that are as generic as 'value'.  additionally, this variable is shadowing the 'value' from the outer scope, which we usually try to avoid since it sometimes causes confusion about what is actually within it.

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7, line 83: like below, i would recommend using native syntax as much as possible when it's possible.  in this case doing an object.entries and then either a map or a for..of loop depending on if you wanted to do a functional or procedural style.  since we're building up the object like this, i would probably do the procedural style.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    if (_.has(value, 'children')) {
+      const temp: Record<string, unknown> = _.get(value, 'children') || {};
+      _.forEach(childrenfinder(temp), function (value) {
+        arr.push(value);
+      });
+    }
+  });
+  return arr;
+}
+function shafileMapper(
```
Amndeep7, line 75: sha2filenameMapper as a potential alternative name that helps explain what's going on.  i would also make the return type Record<string, string> since it's mappings between sha->filename.  would also be useful to have a comment explaining why this is necessary.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  return shamappings;
+}
+function determineStatus(score: number): ExecJSON.ControlResultStatus {
+  if (score == 0) {
+    return ExecJSON.ControlResultStatus.Passed;
+  }
+  return ExecJSON.ControlResultStatus.Failed;
+}
+
+function arrayifyObject(
```
Amndeep7, line 107: ideally we minimize the amount of pre-processing we need to do since it makes it harder for maintainers to come back later on and understand how the mapping works since the paths/data that the mapper uses are not actually analogous to the original data.

you said that some of this was necessary because the controls block was expecting 'code_desc' and other attributes at that path which is strange to me since that shouldn't be the case.  when you're working on resolving this issue, reach out to me and we can do a working session.

Amndeep7 (changes requested):

Amndeep7 (changes requested):
fingerprinting still needs to be done

Amndeep7 (changes requested):
please make a draft pr against the saf cli supporting the new mapper

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7: this applies to the results and mapper constructors as well

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  constructor(
+    conveyor: Record<string, unknown>,
+    meta: Record<string, unknown>,
+    type: string
+  ) {
+    const data = meta;
+    _.set(data, 'api_response.results', conveyor);
+    super(data);
+    this.type = type;
+    this.data = data;
```
Amndeep7, line 215: not necessary since super(data) assigns this.data

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+        start_time: {path: 'start_time'}
+      }
+    ]
+  };
+}
+
+export class ConveyorMapper extends BaseConverter {
+  data: Record<string, unknown>;
+  type: string;
+  defaultMapping(): MappedTransform<ExecJSON.Execution, ILookupPath> {
```
Amndeep7, line 178: since we aren't passing in any parameters or anything, we could just provide the 'mappings' object which is specified in baseconverter and get rid of the need to do the manual this.setmappings

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+function controlMappingConveyor(): MappedTransform<
+  ExecJSON.Control & ILookupPath,
+  ILookupPath
+> {
+  return {
+    id: {path: 'sha256'},
+    title: {path: 'filename'},
+    desc: {path: 'cve_summary'},
```
Amndeep7, line 150: you said cve_summary was empty all the time so this probably should not be set.  would be nice to have data here if at all possible though.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+      path: 'result.score',
+      transformer: (value) => {
+        return value / 1000;
+      }
+    },
+    refs: [],
+    tags: {
+      nist: DEFAULT_NIST_TAG
+    },
+    source_location: {
```
Amndeep7, line 161: source_location can be an empty object

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
+import {ExecJSON} from 'inspecjs';
+import _ from 'lodash';
+import {version as HeimdallToolsVersion} from '../package.json';
+import {BaseConverter, ILookupPath, MappedTransform} from './base-converter';
+const DEFAULT_NIST_TAG = ['SI-2', 'RA-5'];
+
+function createDescription(
```
Amndeep7, line 7: probably worthwhile to reorder functions so that related functions are closer together.

Amndeep7 (changes requested):

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixe

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: enum created

jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: errors out when line 215 is omitted

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: this returns an array, does it still make sense to change it to this method, to then have to transform the result again?

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: removed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: on line 67 had to change to .forEach, as sonarcube doesnt like using .map

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+function sha2filenameMapper(
+  results: Record<string, unknown>
+): Record<string, unknown> {
+  const toplevel = _.get(results, 'api_response.file_tree') as Record<
+    string,
+    unknown
+  >;
+  const shaMappings = {};
+  const arr = childrenfinder(toplevel)
+  arr.forEach((value) => {
```
Amndeep7, line 90: for of

please fix everywhere possible

Amndeep7, line 90: use decomposition to extract variable names instead of doing value[0] etc.

please fix everywhere possible

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    const sha: string = value[0] || '';
+    const filename: string = value[1] || '';
+    _.set(shaMappings, sha, filename);
+  });
+  return shaMappings;
+}
+function determineStatus(score: number): ExecJSON.ControlResultStatus {
+  if (score == 0) {
```
Amndeep7, line 98: triple equals

please fix everywhere possible

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7: for future reference, we tried implementing this and realized we needed access to the parent attribute in order to actually make this work.  currently the baseconverter doesn't support this behavior (however reversebaseconverter does which is ever more reason to combine those classes) so we are forced to do the implementation shown here.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    return ExecJSON.ControlResultStatus.Passed;
+  }
+  return ExecJSON.ControlResultStatus.Failed;
+}
+
+function arrayifyObject(
```
Amndeep7, line 104: better name please?  what type of object are we arrayify-ing?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  parsed: Record<string, unknown>,
+  mappings: Record<string, unknown>
+): Record<string, unknown> {
+  const results = _.get(parsed, 'api_response.results') as Record<string, unknown>;
+  const newout: Record<string, unknown>[] = [];
+  _.forEach(results, (result) => {
```
Amndeep7, line 110: for of for lodash.foreach too

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    const temp = result as Record<string, unknown>;
```
Amndeep7, line 111: why have 'temp' when the only other reference to result is in the 'sha' initialization?  also they're two pointers to the same object so any changes made to the one happen to the other.

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+        createDescription(
+          val as Record<string, unknown>,
+          _.get(temp, 'result.score') as number,
+          _.get(temp, 'response.milestones.service_started') as string,
+          _.get(temp, 'response.service_name') as string,
+          _.get(temp, 'response.milestones.service_completed') as string
+        )
+    );
+    if (descriptions.length == 0) {
+      descriptions.push({
```
Amndeep7, line 126: wait what's going on here, why are we creating this new result?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+  return ExecJSON.ControlResultStatus.Failed;
+}
+
+function arrayifyObject(
+  parsed: Record<string, unknown>,
+  mappings: Record<string, unknown>
+): Record<string, unknown> {
+  const results = _.get(parsed, 'api_response.results') as Record<string, unknown>;
+  const newout: Record<string, unknown>[] = [];
```
Amndeep7, line 109: not sure what this name is supposed to mean

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+function controlMappingConveyor(): MappedTransform<
+  ExecJSON.Control & ILookupPath,
+  ILookupPath
+> {
+  return {
+    id: {path: 'sha256'},
+    title: {path: 'filename'},
+    desc: '',
```
Amndeep7, line 148: do none of the results ever have a value that could serve as an overarching description for the control?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+    impact: {
+      path: 'result.score',
+      transformer: (value) => {
+        return value / 1000;
```
Amndeep7, line 152: probably should convert the 1000 into a constant that in the name explains that it's the highest possible score coming from conveyor

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7: i seem to have misplaced that mapping file that you put together.  could you link it to me again?

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7: why was this marked as resolved when my suggestion wasn't implemented nor an explanation given as to why you were gonna skip it?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+      }
+    },
+    refs: [],
+    tags: {
+      archive_ts: {path: 'archive_ts'},
+      classification: {path:'classification'},
+      expiry_ts: {path: 'expiry_ts'},
+      size: {path:'size'},
+      type: {path:'type'},
+      nist: DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS
```
Amndeep7, line 162: don't forget to add the CCI tags as well which can be derived from the admittedly default/constant nist tags being used here.

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7: this was added in

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+            }
+          ],
+          sha256: ''
+        }
+      ]
+  }
+  constructor(
+    conveyor: Record<string, unknown>,
+    data: Record<string, unknown>,
+    type: string
```
Amndeep7, line 216: should probably rename to also be 'scannerName' or something like that

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
...
+              path: 'api_response.results',
+              ...controlMappingConveyor()
+            }
+          ],
+          sha256: ''
+        }
+      ]
+  }
+  constructor(
+    conveyor: Record<string, unknown>,
```
Amndeep7, line 214: should probably have better names here in the constructor to explain what 'conveyor' is and what the difference is to 'data'

libs/hdf-converters/src/saved properties in control.txt
Amndeep7, line 1: this file shouldn't be in the source code.  clean up the formatting a bit and post it in a wiki page / more formal docs section.  remind me and i can set up a page for you to dump it in.

libs/hdf-converters/src/utils/fingerprinting.ts
```diff
@@ -25,6 +26,7 @@ export enum INPUT_TYPES {
 // Fields to look for inside of JSON structures to determine type before passing to hdf-converters
 const fileTypeFingerprints: Record<INPUT_TYPES, string[]> = {
   [INPUT_TYPES.ASFF]: ['Findings', 'AwsAccountId', 'ProductArn'],
+  [INPUT_TYPES.CONVEYOR]: [],
```
Amndeep7, line 29: we should probs have some type of fingerprinting going on lol

yarn.lock
Amndeep7, line 1: there are conflicts apparently.  i don't see why you needed to update the yarn.lock file in this pr anyways.  easiest thing to do would be to delete your variant here and just pull in the version of the yarn.lock that's in master.

Amndeep7 (changes requested):

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -0,0 +1,244 @@
+import {ExecJSON} from 'inspecjs';
+import _ from 'lodash';
+import {version as HeimdallToolsVersion} from '../package.json';
+import {BaseConverter, ILookupPath, MappedTransform} from './base-converter';
+import {
+  DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS,
+} from './utils/global';
+enum scannerType {
+  Moldy = 'Moldy',
+  Stigma = 'Stima',
```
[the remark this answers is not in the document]
jweissm: fixed

libs/hdf-converters/src/conveyor-mapper.ts
jweissm: fixed

Amndeep7 (approved):

PR #4549, opened 2023-06-01
Remove Splunk SDK usage

charleshu-8:
Substitute Splunk SDK usage in Splunk mapper w/ direct HTTP requests using Axios.


libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
+    const hostname = config.port
+      ? `${config.scheme}://${config.host}:${config.port}`
+      : `${config.scheme}://${config.host}:8089`;
     logger.info(
-      `Logging into Splunk Service: ${config.host} with user ${config.username}`
+      `Logging into Splunk instance at ${hostname} with user ${config.username}`
     );
-    logger.verbose('Got Execution: ' + filename);
+
+    logger.verbose(`Got execution: ${filename}`);
```
em-c-rod, line 442: Can you update this verbose comment to be something more specific?

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -152,42 +151,45 @@ function consolidateFilePayloads(
...
-  } else {
-    service = new splunkjs.Service(config);
-  }
+  config: SplunkConfig
+): Promise<string> {
+  const hostname = config.port
+    ? `${config.scheme}://${config.host}:${config.port}`
+    : `${config.scheme}://${config.host}:8089`;
+  const username = (config.username ??= '');
+  const password = (config.password ??= '');
```
em-c-rod, line 160: What does ??= do?

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -152,42 +151,45 @@ function consolidateFilePayloads(
...
+      .then(
+        (response) => resolve(response.data.sessionKey),
+        (error) => {
+          try {
+            if (error.response.status === 401) {
+              reject('Incorrect Username or Password');
+            }
+          } catch (error) {
+            reject(
+              'Failed to login: Please check your CORS configuration or validate that you have inputted the correct domain'
```
em-c-rod, line 188:
```suggestion
              'Failed to login: Please check your CORS configuration and validate that your input has the correct domain'
```

em-c-rod:
How did you test this? Can you add some information about the results?

libs/hdf-converters/src/splunk-mapper.ts
charleshu-8: [Nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment). Only assigns the username/password if it is null.

charleshu-8:
@em-c-rod Manual testing by passing files back and forth from a Splunk container and a local Heimdall Lite instance and verifying that the files passed through intact. The goal was mainly to ensure that nothing broke preexisting features, but I'm not sure if that's the particular results you're asking about.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: Would something along the lines of `Found designated file to transfer` be more appropriate?

wdower (approved):
LGTM

libs/hdf-converters/src/splunk-mapper.ts
em-c-rod: Nice

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: Sure, here's the output for pushing to Splunk (it gives you a log) and pulling from Splunk (it doesn't give a log, but shows the content of the instance instead).

To Splunk
From Splunk

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -343,66 +344,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
-          (err: any) => {
-            if (err) {
-              reject(err);
-            }
-            logger.verbose(`Successfully uploaded ${chunk.length} control(s)`);
-            resolve();
-          }
-        );
-      });
+      for (const chunk of _.chunk(splunkData.controls, UPLOAD_CHUNK_SIZE)) {
```
Amndeep7, line 389: these promises potentially don't complete.  they need to be wrapped in a promise.all or something and awaited.

keep an eye out for similar situations where promises are just created in order to do work within that same function instead of being returned for some user to deal with.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
+    const hostname = config.port
+      ? `${config.scheme}://${config.host}:${config.port}`
+      : `${config.scheme}://${config.host}:8089`;
     logger.info(
-      `Logging into Splunk Service: ${config.host} with user ${config.username}`
+      `Logging into Splunk instance at ${hostname} with user ${config.username}`
     );
-    logger.verbose('Got Execution: ' + filename);
+
+    logger.verbose(`Found designated file to transfer: ${filename}`);
```
Amndeep7, line 442: I appreciate the work fixing up the logging

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
-    logger.verbose('Using GUID: ' + guid);
-    return new Promise((resolve) => {
-      if (config.username && config.password) {
-        service.login((error: any) => {
-          this.handleSplunkError(error);
-        });
+    logger.verbose(`Using GUID: ${guid}`);
+
+    // Attempt to authenticate using given credentials
+    const username = (config.username ??= '');
```
Amndeep7, line 447: i feel like i've seen bits and pieces of these lines spread throughout the file.  is it possible to set up some utility/factory functions to do this stuff for you that would then for ex. return an object that can be parsed out with all of the stuff you could need?

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
+    );
+    this.axiosInstance.defaults.headers.common[
+      'Authorization'
+    ] = `Bearer ${authResponse.data.sessionKey}`;
+
+    // Request all available indexes
+    const indexResponse = await this.axiosInstance.get(
+      `${hostname}/services/data/indexes`,
+      {
+        params: {count: 0}
```
Amndeep7, line 464: why 0?

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
       }
-      service.indexes().fetch(async (error: any, indexes: any) => {
-        if (error) {
-          this.handleSplunkError(error);
-        } else if (!indexes) {
+    );
+
+    return new Promise((resolve) => {
```
Amndeep7, line 468: when using promise + resolve/reject syntax you need to always make sure to handle the potential error case.  ex. throwing/catching errors like this definitely isn't gonna work out how you want it to

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
           );
-          if (availableIndexes.includes(config.index)) {
-            const targetIndex = indexes.item(config.index);
-            logger?.verbose(`Have index ${targetIndex.name}`);
+          logger.verbose(`Available indexes: ${indexNames.join(', ')}`);
+
+          // Parse available indexes for user desired index
+          if (indexNames.includes(config.index)) {
+            const targetIndex = indexes.filter(
+              (index: {name: string}) => index.name == config.index
```
Amndeep7, line 485: triple equals

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -432,48 +431,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
+            )[0];
+            logger.verbose(`Found index: ${targetIndex.name}`);
+
+            // Post given file(s) to identified index
             const splunkData = this.createSplunkData(guid, filename);
-            this.uploadSplunkData(targetIndex, splunkData)
+            this.uploadSplunkData(config, targetIndex, splunkData)
```
Amndeep7, line 491: there's no guarantee that this promise resolves by the end of the function.  clearer to deal with this issue using async/await.

libs/hdf-converters/package.json
```diff
@@ -26,7 +26,6 @@
   },
   "dependencies": {
     "@aws-sdk/client-config-service": "^3.95.0",
-    "@mitre/splunk-sdk-no-env": "^1.10.0",
```
Amndeep7, line 29: once this pr is merged we should have a plan of action to deprecate this library

apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
```diff
@@ -59,7 +59,7 @@ import FileList from '@/components/global/upload_tabs/aws/FileList.vue';
 import {SnackbarModule} from '@/store/snackbar';
 import {LocalStorageVal} from '@/utilities/helper_util';
 import {checkSplunkCredentials} from '@mitre/hdf-converters/src/splunk-mapper';
-import {SplunkConfig} from '@mitre/splunk-sdk-no-env';
+import {SplunkConfig} from '@mitre/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper';
```
Amndeep7, line 62: feels weird to import types and functions from the mapper for use here.  it would probably be better to make a utility file that includes these things and import from there instead.

```
hdfconverters -> src -> splunk -> utilities
                               -> mapper (depends on ./utilities)
                        converters-from-hdf -> splunk -> reverse-mapper (depends on ../../splunk/utilities)
```

Amndeep7 (changes requested):

Amndeep7 (changes requested):
also the e2e tests were failing but idk if that was cause of something actually going wrong or if they were just being flaky

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -303,33 +299,69 @@ export class SplunkMapper {
...
-                }
+      // Ping Splunk instance every 50 ms on status of search job
+      const awaitJob = setInterval(() => {
+        this.axiosInstance
+          .get(`${this.hostname}/services/search/jobs/${job}`)
+          .then(
+            async (response) => {
+              // If search job is complete, kill interval loop and ping Splunk for search job results
+              if (
+                response.data.entry[0].content.dispatchState == 'DONE' &&
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -152,42 +151,45 @@ function consolidateFilePayloads(
 }
 
 export async function checkSplunkCredentials(
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -303,33 +299,69 @@ export class SplunkMapper {
...
+                const queryJob = await this.axiosInstance.get(
+                  `${this.hostname}/services/search/jobs/${job}/results`,
+                  {
+                    params: {count: 100000, output_mode: 'json_rows'}
+                  }
+                );
+                resolve(this.parseSplunkResponse(query, queryJob.data));
+                // If search job returns a bad state result, kill interval loop and fail the query
+              } else if (
+                response.data.entry[0].content.dispatchState == 'PAUSE' ||
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: Changed as requested.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: (=^ェ^=)

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: > The maximum number of results to return. If value is set to 0, then all available results are returned. 
[Per Splunk SDK docs](https://docs.splunk.com/Documentation/Splunk/9.0.5/RESTREF/RESTsearch#search.2Fv2.2Fjobs.2F.7Bsearch_id.7D.2Fresults). Also found in implemented in SDK.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -303,33 +299,69 @@ export class SplunkMapper {
...
+              if (
+                response.data.entry[0].content.dispatchState == 'DONE' &&
+                response.data.entry[0].content.isDone
+              ) {
+                clearInterval(awaitJob);
+
+                const queryJob = await this.axiosInstance.get(
+                  `${this.hostname}/services/search/jobs/${job}/results`,
+                  {
+                    params: {count: 100000, output_mode: 'json_rows'}
```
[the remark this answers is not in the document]
charleshu-8: Actually incorrect value. Should've been 0 per [SDK docs](https://docs.splunk.com/Documentation/Splunk/9.0.5/RESTREF/RESTsearch#search.2Fv2.2Fjobs.2F.7Bsearch_id.7D.2Fresults):

> The maximum number of results to return. If value is set to 0, then all available results are returned. 

Also the endpoint was deprecated, so I changed that as well.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: See https://github.com/mitre/heimdall2/pull/4549#discussion_r1228155934

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -343,66 +344,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
-    targetIndex: splunkjs.Index,
+    config: SplunkConfig,
+    targetIndex: any,
     splunkData: SplunkData
   ): Promise<void> {
     return new Promise((resolve, reject) => {
+      const hostname = config.port
+        ? `${config.scheme}://${config.host}:${config.port}`
+        : `${config.scheme}://${config.host}:8089`;
+      this.axiosInstance.defaults.params['sourcetype'] = MAPPER_NAME;
```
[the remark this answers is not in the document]
charleshu-8: I'm not sure if I understand what you're trying to say here. Feel free to correct me if I'm not answering your question.

Is this an issue with these variables being constant? If so, wouldn't that not matter since the values are handled by the garbage collector the moment the function call ends and they leave scope? By the time we enter another checkSplunk call, the variables from the previous function call would've already been flushed out with the calll stack change. And since these variables are generated off of whatever is passed in for `config`, they're independent of each other anyways.

libs/hdf-converters/src/splunk-mapper.ts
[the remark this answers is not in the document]
charleshu-8: Changed as requested. Went with a set since it had a straightforward implementation.

apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
charleshu-8: Good call. I'll centralize some commonly used functions/types among the Splunk mappers into a single utility & types file.

Tangent:

Looking at the directory tree for hdf-converters, it seems that there a bit of incoherence going on for what goes where specifically when it comes to type files. Instead of expanding it with more paths, I propose reusing and consolidating what we already have to make this more maintainable and heuristically manageable.

So the directory structure would look something like this:

hdfconverters
├── src
│   ├── converters-from-hdf
│    │   └── Splunk
│    │       └── {ONLY REVERSE MAPPER RELEVANT STUFF, MOVE TYPE FILES TO TYPES DIRECTORY}
│   └── utils
│       └── {UTILITIES GO HERE, INCLUDING UTILITIES FOR SPLUNK}
└── types
    └── {ALL TYPE DEFINING FILES GO HERE}

I'm not sure if there's a reason why a lot of these type files are in the same directory as the mappers aside from convenience, so correct me if I'm missing something.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: See https://github.com/mitre/heimdall2/pull/4549#discussion_r1228426104

aaronlippold:
@charleshu-8 #4582 would we be able to perhaps address some of these usability issues while we remove the SDK?

charleshu-8:
> @charleshu-8 #4582 would we be able to perhaps address some of these usability issues while we remove the SDK?

We could, but I would suggest addressing that in another PR since this one is primarily focused on backend integration with Splunk. I want to get the Splunk SDK dependency removed first and foremost without tying it down with extra frontend fixes.

libs/hdf-converters/package.json
charleshu-8: Did we publish this library to a public registry anywhere (like npm)?

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -303,33 +299,69 @@ export class SplunkMapper {
   }
 
   async queryData(query: string): Promise<any[]> {
```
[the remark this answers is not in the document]
charleshu-8: There's a number of conflicting downstream typing that gets broken if anything more specific is used. Could try and change it but would have to unravel some mapper logic.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -152,42 +151,45 @@ function consolidateFilePayloads(
...
+          username: username,
+          password: password
+        }),
+        {params: {output_mode: 'json'}}
+      )
+      .then(
+        (response) => resolve(response.data.sessionKey),
+        (error) => {
+          try {
+            if (error.response.status === 401) {
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -152,42 +151,45 @@ function consolidateFilePayloads(
...
+      .post(
+        `${hostname}/services/auth/login`,
+        new URLSearchParams({
+          username: username,
+          password: password
+        }),
+        {params: {output_mode: 'json'}}
+      )
+      .then(
+        (response) => resolve(response.data.sessionKey),
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -198,46 +200,40 @@ function unixTimeToDate(unixTime: string): Date {
...
-          }
+      // Post to {host}/services/search/jobs endpoint to queue search job for given query
+      // Will return unique search ID (SID) assigned to that search job for future reference
+      this.axiosInstance
+        .post(
+          `${this.hostname}/services/search/jobs`,
+          `exec_mode=blocking&search=${query}`
+        )
+        .then(
+          (response) => resolve(response.data.sid),
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -343,66 +344,64 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
   }
 
   async uploadSplunkData(
-    targetIndex: splunkjs.Index,
+    config: SplunkConfig,
+    targetIndex: any,
     splunkData: SplunkData
   ): Promise<void> {
     return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: Changed as requested.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: Changed as requested.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
charleshu-8: Changed as requested.

libs/hdf-converters/src/splunk-mapper.ts
```diff
@@ -198,46 +200,40 @@ function unixTimeToDate(unixTime: string): Date {
...
-            } else {
-              resolve(createdJob);
-            }
-          }
+      // Post to {host}/services/search/jobs endpoint to queue search job for given query
+      // Will return unique search ID (SID) assigned to that search job for future reference
+      this.axiosInstance
+        .post(
+          `${this.hostname}/services/search/jobs`,
+          `exec_mode=blocking&search=${query}`
```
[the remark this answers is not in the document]
Amndeep7: @ejaronne could you provide your take on this?  are we/heimdall responsible for vetting user input before sending it on to splunk if the original SDK does not do so?

if we were to add it in, would we put it here in the mapper itself or allow this footgun to exist, and instead put the pre-processing in heimdall?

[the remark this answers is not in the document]
Amndeep7: @em-c-rod would like your opinion on this as well

libs/hdf-converters/src/splunk-mapper.ts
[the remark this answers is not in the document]
Amndeep7: I'm ok with that occurring.  I would like an overhaul of this mapper as a whole while we're working on it.  For ex. there are some foreach's still in there that I would like transitioned into for..of's, etc.  Take ownership of everything you touch.

libs/hdf-converters/src/splunk-mapper.ts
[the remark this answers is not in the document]
Amndeep7: I'll take your word that it returns an array since that doesn't seem to come through clearly from the XML version of the output.  If it's returning an array, I guess I'm just unsure why it's only an array of length 1?  Is there a way to force it to return an array of length larger than 1?  If we get it to be a larger array, is it possible for the dispatchState to be 'done' in some but not all of them (which could be the source of a possible error)?

In any case, it can't hurt to do defensive ~~driving~~ parsing.

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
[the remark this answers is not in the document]
Amndeep7: I think I lost track of overarching scope and found a potential shape of a problem and made a comment about it.  Basically my concern was that we were setting state on a global object within some random function that could've been potentially called more than once.  But now I'm seeing it doesn't actually get called more than once probably lol.

libs/hdf-converters/package.json
Amndeep7: https://www.npmjs.com/package/@mitre/splunk-sdk-no-env

apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
Amndeep7: Works fine by me.

> I'm not sure if there's a reason why a lot of these type files are in the same directory as the mappers aside from convenience

Cause people are doing things willynilly lol

libs/hdf-converters/src/splunk-mapper.ts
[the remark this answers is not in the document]
charleshu-8: It appears that the array is a quirk of how Splunk responds to this request when asking for a json formatted response. The SDK doesn't indicate that there are multiple fields in `entry` and the docs don't show any parameters which could be used to prompt such. 

Just in case, I put in some handling to ensure that we're always using the first entry item and to throw an error if there are every any instance of hacing >1 or <1 items in the array since that's unexpected.

libs/hdf-converters/src/splunk-mapper.ts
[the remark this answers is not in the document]
charleshu-8: Moved to https://github.com/mitre/heimdall2/issues/4614. Should be addressed outside of this PR.

libs/hdf-converters/src/splunk-mapper.ts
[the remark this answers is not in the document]
charleshu-8: Changed as requested

Amndeep7 (changes requested):

libs/hdf-converters/src/converters-from-hdf/splunk/reverse-splunk-mapper.ts
```diff
@@ -370,13 +371,19 @@ export class FromHDFToSplunkMapper extends FromAnyBaseConverter {
...
+            .post(
+              `${hostname}/services/receivers/simple`,
+              chunk.map((control) => JSON.stringify(control)).join('\n')
+            )
+            .then(() =>
+              logger.verbose(`Successfully uploaded ${chunk.length} control(s)`)
+            );
+        }
+      );
+      Promise.all(controlEvents);
```
Amndeep7: don't forget to await

apps/frontend/src/components/global/UploadNexus.vue
```diff
@@ -108,7 +108,7 @@ const localTab = new LocalStorageVal<string>('nexus_curr_tab');
 export default class UploadNexus extends mixins(ServerMixin, RouteMixin) {
   @Prop({default: true}) readonly visible!: boolean;
   @Prop({default: false}) readonly persistent!: boolean;
-  activeTab: string = localTab.get_default('uploadtab-local');
+  activeTab: string = localTab.getDefault('uploadtab-local');
```
Amndeep7, line 111: ideally we have cypress tests to validate these changes, but it's unfortunately not the case.  can you attest that you've spun up a heimdall instance and done manual testing on the ui to hit the various components that were modified?

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #4574, opened 2023-06-07
arrayTransformer requires pathVal

candrews:
Like transform, arrayTransformer expects pathVal.

If the pathVal is run through convertInternal, the data is removed from it, resulting in arrayTransformer getting an array of the expected length but with all empty elements.

@Amndeep7 this explains why `refs` is always a list of empty objects in https://github.com/mitre/heimdall2/pull/4480

### DIFF
--- a/libs/hdf-converters/src/base-converter.ts
+++ b/libs/hdf-converters/src/base-converter.ts
@@ -319,23 +319,24 @@ export class BaseConverter {
             pathVal = pathTransform(pathVal, file);
           }
           if (Array.isArray(pathVal)) {
-            v = pathVal.map((element: Record<string, unknown>) => {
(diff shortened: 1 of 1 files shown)


Amndeep7:
@candrews this change in behavior seems to have borked quite a few of the mappers

candrews:
Nevermind, I misunderstood this code.

BTW, some documentation (like what is a `arrayTransformer`?)`, code comments (tsdoc, comments), and stronger typing (not using so many `unknown` and `any` types would be great) would all go a long way towards making this code easier to grok. Cause it's _really_ hard to understand as it it is right now.

Amndeep7:
I was getting a start on revising the baseConverter in one of the other PRs but hadn't had an opportunity to take it to completion.  I definitely agree that it's a mess.

PR #4586, opened 2023-06-12
Refactor Export HTML Feature

charleshu-8:
- Change CSS styling to use TailwindCSS over Bootstrap & jQuery
- Adjustments/enhancements to HTML export file appearance/functionality
- Rework of HTML export backend generation functions
- Add print info button


apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -247,14 +285,29 @@ export default class ExportHTMLModal extends Vue {
         StatusCountModule.countOf(this.filter, 'PassingTestsFailedControl') +
         StatusCountModule.countOf(this.filter, 'FailedTests')
     };
+    this.outputData.severity = {
+      low: SeverityCountModule.low(this.filter),
+      medium: SeverityCountModule.medium(this.filter),
+      high: SeverityCountModule.high(this.filter),
+      critical: SeverityCountModule.critical(this.filter)
+    };
+    this.outputData.compliance =
```
Amndeep7, line 294: image

should probably truncate this to two decimal points

see https://github.com/mitre/heimdall2/pull/4655

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Should be good now

aaronlippold:
Let's see if we can get the Issues from Sonar Cloud updated so we have passing checks

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,638 @@
...
+              data-te-collapse-item
+              data-te-collapse-show
+              aria-labelledby="control-accordion-header-{{controlID}}"
+            >
+              <!-- Control description -->
+              {{data.desc}}
+              <hr class="my-3" />
+
+              <!-- Control tables -->
+              <div>
```
Amndeep7, line 425: try out using that html element that leaves pre-styled text alone so that everything isn't scrunched into one block.  the issue is that it might not look good on smaller/mobile/print views cause it might force itself out of its container so we'll play it by ear for that stuff.  maybe on the xl breakpoint, we can allow it to be formatted but on the other break points it isn't formatted.

edit: saw that you were able to get the code block to wrap, so maybe my concerns can be alleviated if we can force this to still wrap even if it's pre-formatted.

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,638 @@
...
+                <!-- Tests -->
+                <h5 class="text-lg font-bold">Test Results</h5>
+                <table
+                  class="table min-w-full text-left"
+                  aria-describedby="Control test description and result"
+                >
+                  <!-- Generate all available test results -->
+                  {{#hdf.segments}}
+                  <thead>
```
Amndeep7, line 434: maybe instead of using a table header, we can play around with spacing / adding a vertical divider bar.  'name' and 'value' don't really add a lot of clarification to the content of the table so it's basically not worthwhile even having it imo (other than maybe have it be hidden so that it's still there for accessibility purposes idk).

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -118,72 +176,180 @@ export default class ExportHTMLModal extends Vue {
   // If we are exporting a profile we can remove the test/results table
   @Prop({type: String, required: true}) readonly fileType!: string;
 
+  // Generated injectable HTML for icons
+  private iconHTMLStore = {
+    // Passed; green
+    circleCheck: this.iconDatatoSVG(
+      mdiCheckCircle,
+      'rgb(76, 176, 79)',
```
Amndeep7, line 184: check out the colors in the color hack file to make sure they're the same.  ideally we can just grab them from there but at least we should make a reference here and there that the colors must stay aligned

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -118,72 +176,180 @@ export default class ExportHTMLModal extends Vue {
...
+      'A green circle with a check'
```
Amndeep7, line 185: make sure the descriptions actually have meaning associated with them not just the physical description.  as in, green circle with a check is not that useful when the actual meaning is that this was a successful control or whatever.  maybe we don't even need to have a description at all since the picture is fully supplementary to what's going on and there's always a text label that explains in full what's happening.  i would look up documentation on how to write good alt text - maybe ask on #accessibility or #uiux.

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -118,72 +176,180 @@ export default class ExportHTMLModal extends Vue {
...
+      circleHigh: this.iconHTMLStore.circleHigh,
+      // Critical severity
+      circleCritical: this.iconHTMLStore.circleCritical
     }
   };
 
+  // Configures outputData object's report type based on user input
   @Watch('exportType')
   onFileChanged(newValue: string, _oldValue: string) {
     switch (newValue) {
```
Amndeep7, line 320: there should always be a default on a switch statement if we're not doing it on something with predefined values like an enum - here we have something that's stringly-typed but at the end of the day is still a string so anything could be substituted in.  if possible, i would actually like to replace this with a straight up actual enum if at all possible.

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -231,12 +398,60 @@ export default class ExportHTMLModal extends Vue {
     });
   }
 
+  // Takes all available existing file data to use as default settings/data for outputData object
   resetOutputData() {
+    // Total control count
+    const controlCnt =
```
Amndeep7, line 404: just use full words

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -231,12 +398,60 @@ export default class ExportHTMLModal extends Vue {
...
+      StatusCountModule.countOf(this.filter, 'Profile Error');
+
+    // Calculate & set compliance level from control statuses
+    const MAX_DECIMAL_PRECISION = 2;
+    let compliance = 0;
+    // If controls exist, calculate compliance level
+    if (controlCnt > 0) {
+      // Formula: compliance = Passed/(Passed + Failed + Not Reviewed + Profile Error) * 100
+      // Truncate to hundredths decimal place
+      const calculatedCompliance =
```
Amndeep7, line 418: i'm already calculating this in heimdall2/apps/frontend/src/components/cards/compliancechart.vue - in order to ensure that that value is the same, maybe put the math in a utility function that can be called from both places or maybe even just have it be its own thing in the vuex store somewhere.

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -247,14 +462,91 @@ export default class ExportHTMLModal extends Vue {
...
+      high: SeverityCountModule.high(this.filter),
+      critical: SeverityCountModule.critical(this.filter)
+    };
     this.outputData.files = [];
     this.outputData.controlSets = [];
   }
 
+  // Replace all found illegal characters in string with compliant string equivalent
+  replaceIllegalChar(text: string): string {
+    for (const illegalChar in ILLEGAL_CHAR_SET) {
```
Amndeep7, line 478: i think if you swapped over to using for..of instead of for..in, you should be able to not need the type cast stuff going on in what's currently line 481

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -247,14 +462,91 @@ export default class ExportHTMLModal extends Vue {
...
   addDetails(
     control: ContextualizedControl,
     controlLevels: ContextualizedControl[]
-  ): ContextualizedControl & {details: Detail[]} {
+  ): ContextualizedControl & {details: Detail[]} & {controlID: string} & {
+    controlStatus: ControlStatus;
+  } & {controlSeverity: ControlSeverity} & {controlTags: string[]} {
+    // Check status of individual control to assign corresponding icon
+    let statusColor;
+    switch (control.root.hdf.status) {
```
Amndeep7, line 496: file an issue to come back and use the enum/better string typing that's currently in the checklistviewer pr as opposed to the string typing that we're doing here

applies to both status and severity.  that pr also includes having better and consistent aesthetic presentation of all of the things so you can get it from an authoritative source.

apps/frontend/src/components/global/ExportHTMLModal.vue
Amndeep7, line 1: file an issue to go and move this html export stuff into hdfconverters so that we can do it from within both heimdall and the saf cli

Amndeep7 (changes requested):

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,638 @@
...
-      <div class="col-md">
-        <div class="card bg-warning tall-card">
-          <div class="card-body">
-            <div class="card-title">
-              <div class="h4"> {{{icons.circleAlert}}} <span id="nrCount">Not Reviewed:
-                  {{statistics.notReviewed}}</span> </div>
-              <p class="card-text"> Can only be tested manually at this time </p>
+
+          <!-- Compliance level -->
+          <div class="w-64 px-2 py-2">
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Done.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Done.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/package.json
```diff
@@ -15,6 +15,7 @@
   "scripts": {
     "prebuild": "rimraf ../../dist/frontend",
     "build": "vue-cli-service build",
+    "build:tailwind": "npx tailwindcss -i ./src/tailwind.css -o ./public/static/export/style.css --minify --watch",
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

apps/frontend/public/static/export/style.css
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,638 @@
...
+    <div class="px-4 py-6" id="profileInfo">
+      <div
+        class="min-w-full rounded-xl border border-gray-300 bg-gray-100 px-4 py-2 print:border-black print:bg-white"
+      >
+        <h5 class="mb-2 text-xl font-bold">Profile Info</h5>
+        <div class="grid grid-cols-2 md:grid-cols-3 xl:grid-cols-4">
+          <!-- Generate an info report for every profile -->
+          {{#files}}
+          <div class="mx-2 my-2 break-words">
+            <span>
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,637 @@
...
+              </span>
+              <!-- ID -->
+              <span class="hidden w-28 pl-4 sm:block lg:w-32">ID</span>
+              <!-- Severity indicator -->
+              <span class="hidden w-28 pl-3 lg:block">Severity</span>
+              <!-- Title -->
+              <span class="w-1/2 pl-8 sm:block lg:pl-9">Title</span>
+              <!-- Control tags -->
+              <span class="hidden lg:block lg:w-36 xl:w-52"
+                >800-53 Controls & CCIs</span
```
Amndeep7, line 298: please find/fix all instances of

```
<tag
  >content inside</tag
>
```

it should be
```
<tag>
  content inside
</tag>
```

or a one liner if it fits according to the linter

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,637 @@
...
+            >
+              <!-- Control description -->
+              {{data.desc}}
+              <hr class="my-3" />
+
+              <!-- Control tables -->
+              <div>
+                <!-- Tests -->
+                <h5 class="text-lg font-bold">Test Results</h5>
+                <table
```
Amndeep7, line 417: usually you wanna move away from using tables for layout.  i think the situation we're in is currently like a 50/50 sorta situation where i can see why you might wanna use it, but could also see why we shouldn't.

anyways, make sure that the styling for the columns is consistent:
image

here you can see how the left side is centered vertically whereas the right side is bottom aligned

i think a reason to move away from tables is we could make it more vertically oriented so that it'd look like the following:
image

cause look at how much horizontal space is wasted in the mobile view port just to have that left column

image

i'm not opposed to the zebra style shading still being a thing tho

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -118,72 +192,134 @@ export default class ExportHTMLModal extends Vue {
   // If we are exporting a profile we can remove the test/results table
   @Prop({type: String, required: true}) readonly fileType!: string;
 
+  // Generated injectable HTML for icons
+  // Icons colors should align with Vue colors used in general dashboard
```
Amndeep7, line 196: maybe give a path to where those colors are defined?

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -118,72 +192,134 @@ export default class ExportHTMLModal extends Vue {
...
-  iconDatatoSVG(
+  // Generate SVG HTML for icons for injection into export template
+  iconDataToSVG(
     iconData: string,
     fill: string,
     widthPx = 24,
     heightPx = 24
   ): string {
-    return `<svg style="width:${widthPx}px; height:${heightPx}px" viewBox="0 0 ${widthPx} ${heightPx}"><path fill="${fill}" d="${iconData}"/></svg>`;
+    return `<svg style="width:${widthPx}px; height:${heightPx}px" viewBox="0 0 ${widthPx} ${heightPx}" role="img" aria-hidden="true"><path fill="${fill}" d="${iconData}"/></svg>`;
```
Amndeep7, line 319: instead of always setting it to be hidden here, i would make it respond to a parameter / passed in alt-text with the default behavior being to set it to hidden.  and then we can leave a comment where we define all of our icons to say that since they're purely supplementary we don't need to provide alt text.

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -59,55 +62,126 @@ import {InspecDataModule} from '../../store/data_store';
...
-  value: string;
-  class?: string;
+// Illegal characters which are not accepted by HTML id attribute
+// Generally includes everything that is not alphanumeric or characters [-,_]
+// Expand as needed
+const ILLEGAL_CHARACTER_SET = [['\\.', '___PERIOD___']];
+
+// All selectable export types for an HTML export
+const enum FileExportTypes {
+  executive = 'executive',
```
Amndeep7, line 78: https://www.typescriptlang.org/docs/handbook/enums.html

enums are usually (first letter also capitalized) camel case

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -118,72 +192,134 @@ export default class ExportHTMLModal extends Vue {
...
+      // High severity
+      circleHigh: this.iconHTMLStore.circleHigh,
+      // Critical severity
+      circleCritical: this.iconHTMLStore.circleCritical
     }
   };
 
+  // Configures outputData object's report type based on user input
   @Watch('exportType')
   onFileChanged(newValue: string, _oldValue: string) {
```
Amndeep7, line 292: the usage of the enum should extend as fully up the chain as possible.  so in this case, 'new value' should also be an instance of fileexporttypes.

Amndeep7, line 292: if old values aren't ever used, let's get rid of it as a parameter entirely.  js allows you to supply additional parameters to functions.

```
$ node
Welcome to Node.js v16.19.0.
Type ".help" for more information.
> let a = (b) => console.log(b);
undefined
> a
[Function: a]
> a()
undefined
undefined
> a(1)
1
undefined
> a(1, 2)
1
undefined
>
```

Amndeep7, line 292: i don't think files are actually changing.  this function should probably be renamed to something more accurately reflecting what's going on.

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
```diff
@@ -1,271 +1,637 @@
...
+      <!-- Heimdall report -->
+      <a href="#" class="px-3 py-2 text-lg font-medium">Heimdall Report</a>
+      <!-- Profile info link -->
+      <button role="button" aria-label="Jump to profile info card">
+        <a href="#profileInfo" class="px-3 py-2 text-sm" aria-current="page"
+          >Profile Info</a
+        >
+      </button>
+      <!-- Control sets dropdown menu -->
+      {{#showControlSets}}
```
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
Amndeep7: Let's get other folks' perspective on this.  @ChrisHinchey @wdower do yall have opinions?

Let's bring this up at standup as well (or ask around in the lab), but if Chris or Will have opinions that we like before we bring it up to the big group, we can probs save some time.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested x2.

apps/frontend/public/static/export/template.html
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
```diff
@@ -33,21 +33,21 @@
...
       <div class="inline-block" data-te-dropdown-ref>
         <!-- Button def -->
         <button
           type="button"
           class="inline-flex px-3 py-2 text-sm"
           data-te-dropdown-toggle-ref
-          href="#controlSetsList"
+          href="#resultSetsList"
           role="button"
-          aria-controls="controlSetsList"
```
Amndeep7: search replace issue here - i think it needs to stay 'aria-controls'

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -313,10 +325,15 @@ export default class ExportHTMLModal extends Vue {
   iconDataToSVG(
     iconData: string,
     fill: string,
+    description = 'None',
```
Amndeep7, line 328: would probably make this a purely an optional parameter instead of one that has a default value (would also probably reorder the parameters so this one goes last in line).  if the user intentionally supplies a description, then we should use that.  otherwise, we can move on when we see that it's undefined.

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -313,10 +325,15 @@ export default class ExportHTMLModal extends Vue {
...
     widthPx = 24,
     heightPx = 24
   ): string {
-    return `<svg style="width:${widthPx}px; height:${heightPx}px" viewBox="0 0 ${widthPx} ${heightPx}" role="img" aria-hidden="true"><path fill="${fill}" d="${iconData}"/></svg>`;
+    if (description === 'None') {
+      return `<svg style="width:${widthPx}px; height:${heightPx}px" viewBox="0 0 ${widthPx} ${heightPx}" role="img" aria-hidden="true"><path fill="${fill}" d="${iconData}"/></svg>`;
```
Amndeep7, line 333: could probably simplify this so we don't need a wrapping conditional, or at least not one with as much duplication as is going on here (DRY etc)

```
const label = ` aria-label="${description}"`;
return `<svg style="width:${widthPx}px; height:${heightPx}px" viewBox="0 0 ${widthPx} ${heightPx}" role="img" title="${description}"${description ? label : ''}><path fill="${fill}" d="${iconData}"/></svg>`;
```

apps/frontend/public/static/export/template.html
Amndeep7, line 1: image

this was output when i opened up the exported file.  it also caused the system to be unresponsive in the beginning until that error eventually popped up

Amndeep7 (changes requested):
I liked the changes you made w/r to going from a table to the divs

apps/frontend/src/utilities/compliance_util.ts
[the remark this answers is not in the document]
Amndeep7: the pr: https://github.com/mitre/heimdall2/pull/4818

Amndeep7 (changes requested):

apps/frontend/public/static/export/template.html
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/components/global/ExportHTMLModal.vue
charleshu-8: Changed as requested.

apps/frontend/src/utilities/compliance_util.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// Format all final compliance level results to hundredths place percentage of compliance level
+// Returns string typed compliance level
+export function formatCompliance(rawCompliance: number): string {
+  let truncatedCompliance =
+    Math.trunc(Math.pow(10, MAX_DECIMAL_PRECISION) * rawCompliance) /
+    Math.pow(10, MAX_DECIMAL_PRECISION);
+
+  // Check if calculated compliance is valid
+  if (truncatedCompliance < 0) {
```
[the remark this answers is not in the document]
charleshu-8: No, but can't hurt to be careful.

apps/frontend/src/utilities/compliance_util.ts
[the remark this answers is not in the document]
charleshu-8: Changed as requested.

apps/frontend/public/static/export/template.html
charleshu-8: Checked and the issue seems to stem from Tailwind Elements. Will look into it further but it seems rather odd.

Amndeep7 (approved):

PR #4600, opened 2023-06-14
testing backend results with forced updated of sequelize dependency

Hookwitz:


apps/backend/src/database/database.service.ts
```diff
@@ -12,9 +12,9 @@ export class DatabaseService {
   }
 
   async cleanAll(): Promise<void> {
-    Object.values(this.sequelize.models).forEach((model) => {
-      model.destroy({where: {}});
-    });
+    for (const model of Object.values(this.sequelize.models)) {
+      await model.destroy({where: {}});
```
Amndeep7, line 16: instead of awaiting every iteration within this loop, it is better to make a collection of all of the promises, make them a single promise with promise.all, and then await that.

Amndeep7 (changes requested):

apps/backend/src/database/database.service.ts
Hookwitz: Added Promises.all to manage collection of promises.

apps/backend/src/database/database.service.ts
```diff
@@ -12,9 +12,11 @@ export class DatabaseService {
   }
 
   async cleanAll(): Promise<void> {
-    Object.values(this.sequelize.models).forEach((model) => {
-      model.destroy({where: {}});
-    });
+    const promises: Promise<number>[] = [];
+    for (const model of Object.values(this.sequelize.models)) {
```
Amndeep7, line 16: could be simplified by just doing a map against the object.values

i.e. something like

await Promise.all(Object.values(this.sequelize.models).map(model => model.destroy(blah)))

Amndeep7 (changes requested):

apps/backend/src/database/database.service.ts
Hookwitz: Good suggestion. I disregarded `.map` because it also is not able to handle promises gracefully.

Amndeep7 (approved):

Issue #4523, opened 2023-05-25
Rename package from Heimdall Tools to HDF Converters


em-c-rod:
As an example:
```    
import {version as HeimdallToolsVersion} from '../package.json';
...
platform: {
      name: 'Heimdall Tools',
      release: HeimdallToolsVersion
    },
    version: HeimdallToolsVersion,
```

Recommend changing to:
```    
import {version as HDFConvertersVersion} from '../package.json';
...
platform: {
      name: 'HDF Converters',
      release: HDFConvertersVersion
    },
    version: HDFConvertersVersion,
```

Example below, but present in many converters:
https://github.com/mitre/heimdall2/blob/master/libs/hdf-converters/src/burpsuite-mapper.ts#L82

Amndeep7:
some considerations we might need to keep in mind for some of the converters that have to handle like full-cycle situations (format -> hdf -> format / hdf -> format -> hdf) ex. ASFF

Amndeep7:
talked it out with @ejaronne, and "MITRE SAF OHDF Converters" is better with "MITRE SAF OHDF Converters Version" for the ckl export (and maybe other exports)

PR #4601, opened 2023-06-14
Update DBProtect mapper

charleshu-8:
- Standardize DBProtect mapper fields so it's in line with other updated/newer mappers.
- Add testing for DBProtect mapper.
- Add passthrough attribute for withraw passing.


charleshu-8:
NOTE: SonarCloud is failing from some commented out code which exports HDF converted files to their respective directories for testing the tests. From my understanding, it was agreed upon (and is currently practiced in the other mapper tests) to leave those statements there for further testing, so this is a false positive for now.

libs/hdf-converters/src/dbprotect-mapper.ts
```diff
@@ -99,53 +103,53 @@ function idToString(id: unknown): string {
...
+
+  mappings: MappedTransform<
+    ExecJSON.Execution & {passthrough: unknown},
+    ILookupPath
+  > = {
     platform: {
       name: 'Heimdall Tools',
-      release: HeimdallToolsVersion,
-      target_id: ''
+      release: HeimdallToolsVersion
```
em-c-rod, line 114: I'm not sure if it would be an easy find and replace or more complicated, but should we change this to whatever it is now? Because Heimdall Tools is deprecated?

libs/hdf-converters/test/mappers/forward/dbprotect_mapper.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+describe('dbprotect_mapper_check', () => {
+  it('Successfully converts DBProtect check scan targeted at a local/cloned repository data', () => {
+    const mapper = new DBProtectMapper(
+      fs.readFileSync(
+        'sample_jsons/dbprotect_mapper/sample_input_report/DbProtect-Check-Results-Details-XML-Sample.xml',
+        {encoding: 'utf-8'}
+      )
+    );
+
+    // fs.writeFileSync(
```
em-c-rod, line 14: Would it make SonarQube happy to just have this once at the top of the file instead of leaving each one commented?

libs/hdf-converters/src/dbprotect-mapper.ts
charleshu-8: [Are you referring to this](https://github.com/mitre/heimdall2/issues/4523)? If so, it should be a simple fix on this mapper since it's just naming.

libs/hdf-converters/test/mappers/forward/dbprotect_mapper.spec.ts
Amndeep7: Yeah SQ complains about basically any commented out code since it (more or less rightfully) thinks that you can just remove it and if you want it back then grab it from source control.  However, we use it frequently enough that it's just easier to have on hand always so I'm fine with this being a WONTFIX.

Amndeep7 (approved):

PR #4632, opened 2023-06-21
Improving Splunk Interface

charleshu-8:
- Bug fixes to Splunk UI (reload button, uncaught error chains)
- Standardizing buttons to align w/ Tenable tab
- Require user to enter data into empty fields on Splunk login page
- Set index field on Splunk upload to an empty string default
- Add ability to search a new index w/in the file list view
- Help text changes


apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
```diff
@@ -84,39 +88,47 @@ export default class AuthStep extends Vue {
...
 
+  /** Form required field rules. Maybe eventually expand to other stuff */
+  reqRule = (v: string | null | undefined) =>
+    (v ?? '').trim().length > 0 || 'Field is Required';
+
   async login(): Promise<void> {
-    if (!/^https?:\/\//.test(this.hostname)) {
-      this.hostname = `https://${this.hostname}`;
-    }
+    if (this.index !== '') {
```
Amndeep7, line 96: do this as a guard clause, i.e. if this.index === '', throw that error.  then you can have the rest of your code be unindented

Amndeep7 (changes requested):

apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
charleshu-8: Maybe I'm missing something, but how would error handling work this high up in the frontend?

georgedias: What Amndeep is saying is use:
if (this.index !== '') {
    SnackbarModule.failure('Failed to login - A valid index is required');
} else {
   ....
}

Amndeep7: don't even need an explicit else clause if you just return early from the function

charleshu-8: Issue should be addressed now

apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
```diff
@@ -79,18 +83,27 @@ const localHostname = new LocalStorageVal<string>('splunk_hostname');
 export default class AuthStep extends Vue {
   @Prop({type: String, required: false}) indexToShow?: string;
   username = '';
-
   password = '';
   hostname = '';
   index = '';
 
+  // Form required field rule
+  reqRule = (v: string | null | undefined) =>
```
Amndeep7, line 91: this seems like a duplicated function from the aws/auth file.  is it possible to put it in a utility function for use in aws/splunk/tenable files?

apps/frontend/src/components/global/upload_tabs/splunk/AuthStep.vue
```diff
@@ -105,11 +118,7 @@ export default class AuthStep extends Vue {
       localUsername.set(this.username);
       localPassword.set(this.password);
       localHostname.set(this.hostname);
-      if (this.indexToShow === undefined) {
-        localSplunk2HDFIndex.set(this.index);
```
Amndeep7, line 109: i saw that you removed the default of 'hdf' as the index.  is there no need for the splunk2hdf thing anymore?

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportSplunkModal.vue
```diff
@@ -23,7 +23,7 @@
           <v-stepper-items>
             <v-stepper-content step="1">
               <AuthStep
-                index-to-show="hdf"
+                index-to-show=""
```
Amndeep7, line 26: add a note into the help area that the index needs to exist on the splunk instance before exporting to it

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportSplunkModal.vue
```diff
@@ -45,7 +45,10 @@
                   Are you sure that you have configured it properly?
                 </span>
                 <br />
-                <span>The 'index' field requires a valid index name from the specified Splunk instance.</span>
+                <span
+                  >The 'index' field requires a valid index name from the
```
Amndeep7, line 49: please don't format it like this.  keep the braces with their tags.  put just the content into its own line.

apps/frontend/src/components/global/ExportSplunkModal.vue
```diff
@@ -45,6 +45,11 @@
                   Are you sure that you have configured it properly?
                 </span>
                 <br />
+                <span
```
Amndeep7, line 48: image

now that we're putting actual content into there instead of a link out, we should probably make it left aligned instead of centered.  this is for both the splunk tab on the load modal as well as in the export.

apps/frontend/src/components/global/ExportSplunkModal.vue
```diff
@@ -45,6 +45,11 @@
...
+                  >The 'index' field requires a valid index name from the
+                  specified Splunk instance.</span
+                >
+                <br />
                 <span>
                   For installation instructions and further information, check
```
Amndeep7, line 54: image

that leads out to a separate plugin we had previously written, but i thought we were using the splunk mapper at this point?

Amndeep7: we might want to harvest the accreditations and material from the readme and put it into a page on the heimdall wiki if i'm correct about us using hdf converters now so that we don't lose track of like those screenshots and other explanatory material (so long as that material is accurate)

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportSplunkModal.vue
```diff
@@ -38,26 +38,40 @@
...
                 <span>
-                  For installation instructions and further information, check
-                  here:
+                  Accessing a Splunk instance from Heimdall requires the input
+                  of the following information:
+                  <br />
+                  username: A qualified username recognized by the referenced
+                  Splunk instance.
+                  <br />
+                  password: A qualified password recognized by the referenced
```
Amndeep7, line 55: Question: is there anyway to use an apikey/auth token or something with splunk as opposed to having to use hard credentials?

apps/frontend/src/components/global/ExportSplunkModal.vue
```diff
@@ -38,26 +38,40 @@
...
+                  Include port number if available.
+                  <br />
+                  index: A valid index name within the referenced Splunk
+                  instance.
+                  <br />
+                  For installation instructions and further information, see:
                 </span>
                 <v-btn
                   target="_blank"
-                  href="https://github.com/mitre/hdf-json-to-splunk/"
```
Amndeep7, line 54: can you validate that this plugin is basically just the splunk-hdf mapper but over in splunk instead?  doing a brief skim, it seems to have other functionality embedded in there as well like some sorta script to translate files from s3 into splunk.  we should ask in standup if we know if anyone is still using this plugin cause currently it has been unmaintained for 3-4 years, and i'd like to officially archive it if possible.  i didn't notice any dependencies aside from the sdk being like hard embedded in the repo for some reason so hopefully there's no security concerns if we are in fact forced to leave it up.  finally, if we are able to archive it, we'll probably be breaking some links since we have a custom dns entry up which we should probably have redirect to some landing page on the saf site saying that whatever you tried accessing has been discontinued.  the reason we have to do that is if we archive it without doing anything about the domain name existing, then folks can grab it and point it elsewhere which is kinda wild that that's how the internet works lol.

Amndeep7 (commented):
my questions aren't blockers.  i'll probably get this merged in tomorrow.

apps/frontend/src/components/global/ExportSplunkModal.vue
charleshu-8: Yes, I believe authentication tokens are supported as well.

Amndeep7 (approved):

PR #4655, opened 2023-06-27
No rounding in compliance chart

### LINKED ISSUE #4642: Improve rounding so we do not round up to 100%
Heimdall will report 100% compliance when it is between 99.5% and 99.9%

Rounding up to 100% should not occur and is misleading that there are 0 failures.

None

None

image

Amndeep7:
Resolves #4642

Instead of rounding, it does truncation at the end of the math process. Enforces a minimum resolution of four decimal points.

### DIFF
--- a/apps/frontend/src/components/cards/ComplianceChart.vue
+++ b/apps/frontend/src/components/cards/ComplianceChart.vue
@@ -20,7 +20,7 @@ import VueApexCharts from 'vue-apexcharts';
 import Component from 'vue-class-component';
 import {Prop} from 'vue-property-decorator';
 
-export const MAX_DECIMAL_PRECISION = 4;
+export const MAX_DECIMAL_PRECISION = 2;
 
 @Component({
   components: {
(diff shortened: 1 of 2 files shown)


ejaronne:
As long as it truncates, I think 2 decimal points should be fine.

ejaronne (dismissed):
As long as it truncates, I think 2 decimal points is fine. Looks a little silly at 4 decimal points.

em-c-rod (approved):

apps/frontend/src/components/cards/ComplianceChart.vue
```diff
@@ -45,7 +47,12 @@ export default class ComplianceChart extends Vue {
             show: true,
             value: {
               color: '#99a2ac',
-              fontSize: '2rem'
+              fontSize: '2rem',
+              formatter: (val: number) =>
+                `${(
+                  Math.trunc(Math.pow(10, MAX_DECIMAL_PRECISION) * val) /
```
Hookwitz, line 53: Mathematically speaking, both `Math.pow()` functions cancel each other out. So neither is really necessary and could be rewritten to just `Math.trucn(val).toFixed(MAX_DECIMAL_PRECISION)`, right?

Hookwitz (changes requested):

apps/frontend/src/components/cards/ComplianceChart.vue
Amndeep7: nope cause math.trunc cuts of all decimal values.  consequently i bring as much decimal precision as i need over to the left of the decimal point, chop it all off with trunc, and then shove it back over the dot again with the divide

PR #4659, opened 2023-06-28
Admin panel

Rlin232:


apps/frontend/src/components/global/TopbarDropdown.vue
```diff
@@ -29,29 +29,29 @@
               text="Admin Panel"
               icon="mdi-shield-account"
               link="/admin"
-              >Admin Panel</IconLinkItem
-            >
+              >Admin Panel
```
Amndeep7, line 32: i really dislike when content has a lingering brace on it

i.e.

```
<a
>inner stuff
</a>
```

should be

```
<a>
inner stuff
</a>
```

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -113,14 +135,20 @@ import {Prop} from 'vue-property-decorator';
...
+  @Prop({type: Boolean, default: false}) readonly adminPanel!: boolean;
 
   editedGroup: IGroup | null = null;
   selectedGroupUsers: ISlimUser[] | null = [];
   dialogDelete = false;
   dialogDisplayUsers = false;
   search = '';
-  allGroupsHeaders: Object[] = [
+
+  groupsHeaders: Object[] = [
```
Amndeep7, line 146: 'Object' is not a usable type, most things are 'Objects' in js-land.  make this a record, but we know what all the values are so you could also do a more literal type of object type where you either set up the type beforehand or just embed it in the array type

i..e

```
type GroupsHeader = {
  text: string,
  blah
}

groupsHeaders = GroupsHeader[]
```

or

```
groupsHeaders: {text: string, blah}[] =
```

apps/frontend/src/components/global/groups/GroupUsers.vue
```diff
@@ -40,5 +40,17 @@ import {Prop, VModel} from 'vue-property-decorator';
 export default class GroupUsers extends Vue {
   @VModel({type: Array, required: true}) selectedGroupUsers!: ISlimUser[];
   @Prop({type: Boolean, default: false}) readonly dialogDisplayUsers!: boolean;
+
+  get role() {
```
Amndeep7, line 44: what is going on here?

ideally when we have an area where whatever the string is isn't right, we just try using enums instead

Amndeep7 (changes requested):

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -35,31 +35,58 @@
...
+          class="ma-1"
+          @click="displayUsersDialog(item.owners)"
+        >
+          {{ `+${item.owners.length - 3} more` }}
         </v-chip>
       </template>
       <template #[`item.actions`]="{item}">
-        <div v-if="item.role == 'owner'">
-          <GroupModal id="editGroupModal" :create="false" :group="item">
+        <div v-if="item.role == 'owner' || adminPanel">
```
Amndeep7, line 83: triple eq

apps/frontend/src/components/global/groups/GroupUsers.vue
Amndeep7: why was this marked as resolved?  i don't remember us covering this in conversation, there's been no comment supplied here, and no change to the code.

Amndeep7 (changes requested):

apps/frontend/src/components/global/groups/GroupUsers.vue
Amndeep7: im dumb, we did talk about this.  this was spun out to a new issue.

Amndeep7 (approved):

PR #4695, opened 2023-07-07
Group Descriptions

Rlin232:


apps/backend/migrations/20210107170456-create_groups_table.js
```diff
@@ -18,6 +18,11 @@ module.exports = {
         allowNull: false,
         defaultValue: false
       },
+      desc: {
```
camdenmoors, line 21: You should probably use a new migration instead, updating an old migration won't apply to existing deployments

Amndeep7: thanks for the review @camdenmoors

apps/backend/test/constants/groups-test.constant.ts
```diff
@@ -2,15 +2,18 @@ import {CreateGroupDto} from '../../src/groups/dto/create-group.dto';
 
 export const GROUP_1 = {
   name: 'Heimdall Group',
-  public: true
+  public: true,
+  desc: ''
```
Amndeep7, line 6: one of these groups should not have 'desc' set cause we want to test the optional thing right?

apps/backend/test/constants/groups-test.constant.ts
```diff
@@ -2,15 +2,18 @@ import {CreateGroupDto} from '../../src/groups/dto/create-group.dto';
...
 };
 
 export const PRIVATE_GROUP = {
   name: 'Private Heimdall Group',
-  public: false
+  public: false,
+  desc: ''
```
Amndeep7, line 12: we should also have a group with an actual description.

apps/backend/test/constants/groups-test.constant.ts
```diff
@@ -2,15 +2,18 @@ import {CreateGroupDto} from '../../src/groups/dto/create-group.dto';
...
 };
 
 export const UPDATE_GROUP: CreateGroupDto = {
   name: 'Updated Group',
-  public: true
+  public: true,
+  desc: ''
```
Amndeep7, line 18: we should also have a test that explicitly validates changing a description

apps/frontend/src/components/global/groups/GroupManagement.vue
```diff
@@ -80,15 +80,44 @@
...
+          <template #activator="{on}">
+            <v-icon small title="Info" data-cy="info" class="mr-2" v-on="on">
+              mdi-information
+            </v-icon>
+          </template>
+          <v-card color="transparent" flat max-width="600px" overflow-y-auto>
+            <span>
+              <v-card-title>{{ item.name }}</v-card-title>
+              <v-card-text>
+                Visibility: {{ item.public ? 'Public' : 'Private' }} <br />
```
Amndeep7, line 93: don't use breaks.  let v-row/v-col handle this

apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -45,6 +45,13 @@
               </v-tooltip>
             </v-col>
           </v-row>
+          <v-textarea
+            v-model="groupInfo.desc"
+            data-cy="desc"
+            label="Description"
+            rows="1"
+            auto-grow
```
Amndeep7, line 53: what happens when you have a giant description for some reason?

Amndeep7 (changes requested):

apps/backend/migrations/20230712110759-add_desc_to_groups.js
```diff
@@ -0,0 +1,28 @@
...
+      ]).then(() => {
+        // Update all existing evaluations in the database to be public
+        // since we have no way of tracking who uploaded them.
+        // All evaluations going forward will be private.
+        queryInterface.bulkUpdate('Groups', { desc: '' })
+      })
+    })
+  },
+
+  down: (queryInterface, Sequelize) => {
```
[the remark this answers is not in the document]
Rlin232: Yes, this is what the other scripts do

apps/backend/src/groups/dto/group.dto.ts
```diff
@@ -9,6 +9,7 @@ export class GroupDto implements IGroup {
   readonly public: boolean;
   readonly role?: string;
   readonly users: SlimUserDto[];
+  readonly desc: string;
```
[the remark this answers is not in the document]
Rlin232: Yes. This follows the approach we took with the "public" variable. The GroupDto will grab the existing desc from the DB, which is defaulted to an empty string if there isn't already a value.

apps/backend/test/constants/groups-test.constant.ts
Rlin232: Sure, I'll add a description to PRIVATE_GROUP

apps/frontend/src/components/global/groups/GroupModal.vue
Rlin232: It'll just expand properly, hence the "auto-grow"

apps/backend/test/constants/groups-test.constant.ts
Rlin232: No - the createGroupDTO still needs a desc to be passed in even though the decorator says @IsOptional. It is similar to the "public" field, which should have a default passed in in the first place.

apps/backend/migrations/20230712110759-add_desc_to_groups.js
```diff
@@ -0,0 +1,28 @@
...
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.addColumn('Groups', 'desc', {
+          type: Sequelize.STRING,
+          allowNull: false,
+          defaultValue: ''
+        }, { transaction: t }),
+      ]).then(() => {
+        // Update all existing evaluations in the database to be public
```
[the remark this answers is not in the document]
Rlin232: The entire .then() block is not necessary.

apps/frontend/src/components/global/groups/GroupManagement.vue
Rlin232: Used divs instead - for some reason v-row doesn't adopt the padding of its container.

apps/backend/src/groups/dto/create-group.dto.ts
```diff
@@ -9,4 +9,8 @@ export class CreateGroupDto implements ICreateGroup {
   @IsOptional()
   @IsBoolean()
   readonly public!: boolean;
+
+  @IsOptional()
```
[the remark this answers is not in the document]
Rlin232: Input parameters* are allowed to be null, but the database values cannot be null due to the constraints listed in the model

Amndeep7 (approved):

PR #4699, opened 2023-07-10
resolves #4698

### LINKED ISSUE #4698: [BUG] HDF-converters not parsing Nessus scan results correctly
Loading Nessus scans into Heimdall results in the app treating all Nessus scan plugins as single controls, even if they have multiple ReportItem tags within one plugin tag.

**Expected behavior**
Each individual ReportItem in a `.nessus` file should be treated as a single control by Heimdall.

wdower:
The `fast-xml-parser` module was set to strip out the namespace prefix from XML tags (ex the `cm` in `<cm:compliance-check-name`) using the `removeNSPrefix` option for `parseXML`.

Problem was that the nessus mapper was looking for XML tags that had not had the namespace stripped out. So `parseXML` was (correctly) returning values like `compliance-check-name` as attributes, but the mapper was looking for `cm:compliance-check-name`.


Amndeep7 (approved):
If we can get an anonymized sample file that would test against this, it would be great.  Can you ask your sponsor to see if they would mind if we attempted to anonymize their data?  Doesn't need to be a blocker for this pr.

PR #4722, opened 2023-07-17
Unique Group Names

Rlin232:


apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -181,9 +181,13 @@ export default class GroupModal extends Vue {
     }
   }
 
-  async cancel(): Promise<void> {
-    this.dialog = false;
-    location.reload();
+  // Upon group name change, the component will detect whether the current state is acceptable
+  uniqueName = true;
+  checkUniqueName(name: string) {
+    this.uniqueName = !GroupsModule.allGroups.some(
```
Amndeep7, line 187: why are we doing the double negation?  feels like the variable name should just be changed

apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -181,9 +181,13 @@ export default class GroupModal extends Vue {
...
+      (group) => group.name === name && group.id !== this.groupInfo.id
+    );
+    if (!this.uniqueName) SnackbarModule.failure(`Group names must be unique`);
```
Amndeep7, line 190: one line if statements are bad, always use curly braces

test/integration/groups.cy.ts
Amndeep7, line 1: we should be testing the duplicate names restriction too

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
```diff
@@ -0,0 +1,77 @@
...
+
+        -- Update the controlling processing name variable
+          processing_group := rec_group.name;
+        END IF;
+      END LOOP;
+      -- close the cursor
+      CLOSE dup_names;
+    END; $$
+    `)
+    await queryInterface.sequelize.query(`CALL update_group_names_transaction()`)
```
Amndeep7, line 60: why are we writing the transaction and then calling it as opposed to using sequelize.transaction?

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
```diff
@@ -0,0 +1,77 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+    CREATE OR REPLACE PROCEDURE update_group_names_transaction()
+    LANGUAGE plpgsql
+    AS $$
+    DECLARE
+      check_name boolean := true;
+      offSet_cnt INTEGER :=1;
```
Amndeep7, line 11: seems like spaces are fine here (and elsewhere) for assignment

```suggestion
      offSet_cnt INTEGER := 1;
```

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(`
+    CREATE OR REPLACE PROCEDURE update_group_names_transaction()
+    LANGUAGE plpgsql
+    AS $$
+    DECLARE
+      check_name boolean := true;
```
Amndeep7, line 10: use consistent naming patterns.  i'm not sure what it is for sql, but a mix of stringcase and camelcase at teh same time probably isn't it.

also use consistent capitalization - why is boolean all lower case whereas all the others are all caps

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
```diff
@@ -0,0 +1,77 @@
...
+          (SELECT name FROM(SELECT id, name, ROW_NUMBER() OVER(PARTITION BY name) AS occurrences FROM "Groups") AS duplicates
+        WHERE duplicates.occurrences > 1) ORDER BY name, id;
+
+    BEGIN
+      -- Open the cursor
+      OPEN dup_names;
+      LOOP
+        -- retrieve a row of data from the result set (cursor) into the rec_group variable
+        FETCH dup_names INTO rec_group;
+      -- exit when no more records are available in the resul set
```
Amndeep7, line 25:
```suggestion
      -- exit when no more records are available in the result set
```

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
Amndeep7, line 1: does this handle that usecase that we discussed where

```
a
a
a-2
```

turns into 

```
a
a-2
a-2
```

(and therefore still has a dupe)

Amndeep7 (changes requested):

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
Rlin232: For an 
`a, a, a-2`
It will turn into `a-1, a-3, a-2` respectively.

apps/backend/src/groups/groups.service.ts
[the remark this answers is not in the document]
Rlin232: Made the return type void (as it's a function that does a modification)

apps/backend/src/groups/groups.service.ts
```diff
@@ -126,14 +126,28 @@ export class GroupsService {
...
 
   async update(groupToUpdate: Group, groupDto: CreateGroupDto): Promise<Group> {
+    if (
+      (await this.groupModel.findAll({where: {name: groupDto.name}})).length > 1
+    ) {
+      throw new ForbiddenException(
+        'Duplicate key detected. The names of groups must be unique.'
+      );
+    }
     groupToUpdate.update(groupDto);
```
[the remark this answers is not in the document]
Rlin232: Opted for the latter option, for the sake of consistency with updategroupuserrole

apps/backend/migrations/20230725092535-change-group-names-to-unique.js
Amndeep7: why would it be necessary to fix them again?  this is just to transition the group names the one time.  after that, our systems should be able to enforce the uniqueness requirement.  if the user goes in and manually mucks around with the group names within the db itself, that's on them to fix.

Amndeep7 (approved):

PR #4775, opened 2023-08-01
Integration Test Upgrade for Node 18

charleshu-8:
- Part of Node 18 upgrade piecemealing
- Attempt to solve Cypress test unreliability

### DIFF
--- a/test/support/index.ts
+++ b/test/support/index.ts
@@ -26,11 +26,6 @@ Cypress.on('uncaught:exception', (err) => {
   return !err.message.includes('getBoundingClientRect');
 });
 
-// Extends timeout counter to 30000 ms
-Cypress.config('defaultCommandTimeout', 30000);
-// Forces failed tests to retry up to 2 times
-Cypress.config('retries', {runMode: 2});
-
 Cypress.Commands.add('login', ({email, password}) => {
   cy.get('input[name=email]').clear();
   cy.get('input[name=email]').type(email);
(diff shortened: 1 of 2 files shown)


Amndeep7 (changes requested):
i understand that the best practice is to not use async await for the callback functions for the tests themselves, which is why eslint is yelling at us.  however, i think that's cause cypress is expecting us to handle asynchronous stuff through other paradigms.  try out some of the solutions in this stackoverflow question and see what works / feels good to write / looks aesthetic: https://stackoverflow.com/questions/49980311/cypress-io-how-to-handle-async-code.  after we can get it working in the individual tests themselves, let's see if we can move stuff back out into the beforeeach.

test/support/index.ts
```diff
@@ -26,6 +26,11 @@ Cypress.on('uncaught:exception', (err) => {
   return !err.message.includes('getBoundingClientRect');
 });
 
+// Extends timeout counter to 30000 ms
+Cypress.config('defaultCommandTimeout', 30000);
```
Amndeep7, line 30: shouldn't we change the config here? https://github.com/mitre/heimdall2/blob/master/cypress.config.ts

Amndeep7 (changes requested):

test/support/index.ts
charleshu-8: Changed as requested.

Amndeep7 (approved):

Amndeep7 (approved):
Let's get this merged in.  IIRC when testing with node18, it might be necessary to do the Henry/Charles's original approach where we have to put the beforeeach function's commands into each test, but I'd like to avoid that if possible.  This'll be good for now to at least help out with cypress' flakiness as it is right now.

PR #4785, opened 2023-08-03
CAAT converter

### LINKED ISSUE #4740: CAAT export should happen in both Heimdall and the SAF CLI
- [ ] move CAAT export from Heimdall frontend to HDF Converters
- [ ] re-enable it in the frontend
- [ ] add support in the cli

Amndeep7:
Resolves  #4740


apps/frontend/src/components/global/ExportCaat.vue
```diff
@@ -38,190 +33,24 @@ type CAATRow = {
...
-      }
-      // Add rows to sheet
-      const ws = XLSX.utils.json_to_sheet(rows, {header: this.header});
-      wb.Sheets[sheetName] = ws;
+      }).slice();
+      return {data, filename, controls};
     });
-
-    const wbout = XLSX.write(wb, {bookType: 'xlsx', type: 'binary'});
+    const caat = new FromHDFToCAATMapper(inputData).toCAAT(false);
```
em-c-rod, line 49: do you want to pass returnWorkbook=false for readability in .toCAAT?

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+        contextualizedEvaluation = contextualizedFile;
+      }
+      contextualizedEvaluation = ensureContextualizedEvaluation(
+        contextualizedEvaluation
+      );
+      // if provided a filename use that, otherwise the name of the first profile in the execjson (usually there's either only one or if there are multiple then the one at the top is the wrapper or overlay profile) otherwise if there are no profiles at all then a default of 'ExecJSON'
+      return {
+        data: contextualizedEvaluation,
+        filename:
+          datum.filename ??
```
em-c-rod, line 105:
```suggestion
          datum.filename ??        // If a filename is provided, use it.
```

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+          contextualizedEvaluation.data.profiles.at(0)?.name ??
```
em-c-rod, line 106:
```suggestion
          contextualizedEvaluation.data.profiles.at(0)?.name ??      // Otherwise, use the name of the first profile. This is typically the only profile or the wrapper or overlay profile if there are multiple.
```

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+          'ExecJSON',
```
em-c-rod, line 107:
```suggestion
          'ExecJSON',          // Otherwise, set the default.
```

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+        }\r\n`;
+        const message = result.message
+          ? `Message: ${result.message}\r\n\r\n`
+          : '\r\n';
+        return `${statusAndTest}${message}`;
+      }).join('') ?? '';
+    return `${controlStatus}${description}`;
+  }
+
+  createTestResult(hdf: HDFControl): string {
```
em-c-rod, line 142: Food for thought - I recently found this list of best practice coding verbs for PowerShell commands. It seems to be a good reference in general on the best verbs to use. I suggest using something besides create here, perhaps "set".
https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.3

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+    return this.fix(hdf.wraps.impact === 0 ? 'none' : controlSeverity);
+  }
+
+  getRow(control: ContextualizedControl, filename: string): CAATRow[] {
+    const hdf = control.hdf;
+    const allRows: CAATRow[] = _.compact(
+      hdf
+        .canonized_nist(FromHDFToCAATMapper.NistCanonizationConfig)
+        .map((formattedNistTag) => {
+          // I don't think the canonized_nist function would return something that wasn't a nist tag, but this check was there originally quite probably for a reason, but I've not been able to find a sample that triggers this case myself
```
em-c-rod, line 158:
```suggestion
          // I have not found a sample that triggers this case because the canonized_nist function seems to only return a nist tag
```

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+          row['Finding Title'] = `Test ${this.fix(hdf.wraps.id)} - ${this.fix(
+            hdf.wraps.title
+          )}`;
+          if (hdf.start_time) {
+            row['Date Identified'] = FromHDFToCAATMapper.formatDate(
+              new Date(hdf.start_time),
+              '/'
+            );
+          }
+          row['Finding ID'] = `${filename} - Test ${this.fix(hdf.wraps.id)}`;
```
em-c-rod, line 175: Do you need to add checks before referencing all of these nested fields, or is there a default that handles it? For example, what is hdf.wraps is null?

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+            hdf.descriptions.fix ?? hdf.wraps.tags.fix
+          );
+          row['Impact'] = this.createImpact(hdf);
+          return row;
+        })
+    );
+    return allRows;
+  }
+
+  // returnWorkBook: true -> raw workbook class
```
em-c-rod, line 198: Can you clarify why you would set this value one way or another?

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+      // Ensure sheet name uniqueness
+      let renameCount = 2;
+      let sheetName: string = `${
+        d.filename ?? d.data.data.profiles.at(0)?.name ?? 'ExecJSON'
+      }`.substring(0, 31);
+      while (takenSheetNames.includes(sheetName)) {
+        sheetName =
+          `${
+            d.filename ?? d.data.data.profiles.at(0)?.name ?? 'ExecJSON'
+          } `.substring(0, 26) + ` (${renameCount})`;
```
em-c-rod, line 218: Why is this getting the substring(0, 25)? Is that the number of digits allowed for a sheet name?

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  // returnWorkBook: true -> raw workbook class
+  // returnWorkBook: false | undefined -> binary string
+  toCAAT(returnWorkBook?: false) {
+    // Sheet names must be unique across the workbook
+    const takenSheetNames: string[] = [];
+
+    // Define our workbook
+    const wb = XLSX.utils.book_new();
```
em-c-rod, line 205: This is okay with me, but could also be workBook instead of wb. (Similarly with ws).

em-c-rod (commented):
I left some comments and questions.

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
Amndeep7: brutal editing

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
Amndeep7: https://github.com/mitre/heimdall2/blob/master/libs/inspecjs/src/compat_impl/compat_inspec_1_0.ts#L40

wraps can't be null.  if it were null, there would be no underlying data in the first place.  wraps is the name they chose for the attribute that references the original control more or less.  certain attributes like id are required for a control which is why you can access them without any checks.

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
Amndeep7: in our usecase, we've finished processing and just want to print the file out so we could simplify things; however, i'm taking the view that it's possible someone else is a consumer of the hdf-converters library and they might have some usecase where they'd like to further edit the workbook.

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
Amndeep7: sheet names can be at max 31 characters.  i should probs make that a constant.  the 26 is more or less arbitrary so as to have enough space to put in the rename count

apps/frontend/src/components/global/ExportCaat.vue
Amndeep7: I don't know if that's necessary since they can just use their ide to figure out the parameter name/type which clearly (hopefully) states what's going on

libs/hdf-converters/src/converters-from-hdf/caat/reverse-caat-mapper.ts
Amndeep7: sure.  the verb that most aligns with 'create' is 'new' tho according to their list of synonyms to avoid.  i would consider using 'set' if we were passing around the original object and building it up factory style but in this case we're spinning out the creation of the subobjects to other functions and then calling them from the primary function.

PR #4792, opened 2023-08-04
Avoid Hard Refresh When Saving Group

Rlin232:


apps/frontend/src/components/global/RegistrationModal.vue
```diff
@@ -216,6 +216,7 @@ export default class RegistrationModal extends Vue {
             SnackbarModule.notify(
               'You have successfully registered a new user'
             );
+            ServerModule.FetchAllUsers();
```
Amndeep7, line 219: fetch all users returns a promise, but it's not properly chained nor written in async/await notation.

Amndeep7, line 219: i'd appreciate you re-writing this function to use async/await notation as well instead of promise chaining

apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -209,13 +209,13 @@ export default class GroupModal extends Vue {
       this.groupInfo = newGroup();
     }
     this.dialog = false;
+    await GroupsModule.FetchGroupData();
     SnackbarModule.notify(`Group Successfully Saved`);
```
Amndeep7, line 213: isn't it possible for some of these commands esp the axios commands to potentially fail?  we should probably be wrapping a bunch of this in a try/catch instead of always sending out that 'group saved' notification.

apps/frontend/src/components/global/groups/GroupModal.vue
```diff
@@ -209,13 +209,13 @@ export default class GroupModal extends Vue {
       this.groupInfo = newGroup();
     }
     this.dialog = false;
+    await GroupsModule.FetchGroupData();
```
Amndeep7, line 212: should probably have a comment here explaining what's going on

apps/frontend/src/store/server.ts
```diff
@@ -253,9 +253,9 @@ class Server extends VuexModule implements IServerState {
           // then clear their token and refresh the page
           this.Logout()
         )
-        .then(() => {
-          this.FetchAllUsers();
-          GroupsModule.FetchGroupData();
+        .then(async () => {
```
Amndeep7, line 256: can you rewrite these functions using async await?

Amndeep7 (changes requested):

apps/frontend/src/components/global/groups/GroupModal.vue
Rlin232: If the group creation throws an error, that will happen first and go to the snackbar I believe. I'll just add an extra try/catch in case that isn't what actually happens.

Amndeep7 (approved):

Issue #4047, opened 2023-02-06
[BUG] Heimdall JSON to CKL conversion using incorrect tags


pbarrette:
**Describe the bug**
Importing an InSpec JSON results file and exporting to CKL format results in the incorrect population of certain tags

**To Reproduce**
1. Execute a STIG based InSpec scan against an asset.
2. Import the resulting JSON file into Heimdall at https://heimdall-lite.mitre.org/
3. Export the data as a STIG Viewer compatible CKL checklist.
4. Review the CKL in STIG Viewer.

**Expected behavior**
If the JSON contains a `gid` tag, it should be mapped to the `Vuln_Num` attribute in the CKL.

**Desktop (please complete the following information):**
 - OS: Windows Server 2019
 - Browser: Chrome
 - Version: 109.0.5414.75

**Additional context**
Additionally, the SAF CLI hdf2ckl conversion misses multiple tags as partially noted [here](https://github.com/mitre/saf/issues/1074).
Specifically:
The `gid` tag should be mapped to `Vuln_Num`
The `gtitle` tag should be mapped to `Group_Title`
The `rid` tag should get mapped to `Rule_ID`
The `stig_id` tag should get mapped to `Rule_Ver`

I may be missing other translation mappings, but these are the ones I'm seeing offhand.

I have attached a ZIP file containing 4 files: [example.zip](https://github.com/mitre/heimdall2/files/10668008/example.zip)
`inspec.json` - An InSpec JSON file containing scan results.
`heimdall.ckl` - The Heimdall CKL export of the InSpec results.
`saf-cli.ckl` - The SAF CLI converted CKL file.
`stigviewer.ckl` - A STIG Viewer checklist created directly from the official ESXi 6.7 STIG.

The InSpec JSON file is cut down to a single control and sanitized of system specific information.
The STIG Viewer created CKL file has been manually cut down to the same control.

Issue #3051, opened 2022-05-18
move export to ckl functionality to be within hdfconverters


Amndeep7:
while this functionality originally came from heimdall_tools, it makes sense to put it into hdfconverters now.  this also allows us to re-use the same code within the saf cli.

Issue #4604, opened 2023-06-15
[Checklist] Handle Rev5/6 Nist Tags In Rule Body Info


ChrisHinchey:
Currently only 800-53 rev 4 is supported for the rule body information.  Once rev 5/6 are brought into Heimdall, make sure to update what is being pulled.

PR #4841, opened 2023-08-18
Reverse hdf to ckl

Hookwitz:
Resolves: #3051 
Resolves: #4047 
Resolves: #5554 

- [x] Reverse Map
    - [x] Previously ckl to hdf files
    - [x] non-checklist origin hdf files
- [x] tests
    - [x] Remove omitCklUuid function as no longer needed
- [x] hook mapper to ExportCKLModal
- [x] include hdf specific data that may need to be preserved
- [x] datepicker (no default date, has a clear date button)
- [x] target marking textbox (default CUI)
- [x] target comments textbox
- [x] version/release semver parsing
- [x] vuln_discuss formatting (unsure if it matters if the og is all on one line and ours is formatted; the satisfies line)
- [x] header attributes (notice, uuid, filename, description, target key)
- [x] figure out what's going with the version from the sample (pulling out 6.7 while the thing says 1.2 - 6.7 is the version of the app not the ckl?)
- [x] figure out why the marking default value is being used despite an empty marking intentionally being provided
- [x] control's vuln_num needs to be changed to match issue
- [x] Performance optimizations - currently takes ~200ms to process each control - RESULTS: The _.get(object, attribute, default) was computing the default when given a function for that even if the attribute existed for the object. Changing this to only execute when the get is undefined saved performance time.
- [x] Looking nice

Will not be completed in this pr:
- [ ] SAF CLI integration - https://github.com/mitre/saf/issues/385
- [ ] Help/info text - the text provided in the schema is nigh useless, maybe there's stigviewer documentation that goes more in-depth? #5688 
- [ ] Input validation enforcement (currently you can override the validation and click export anyways) - probably needs to be a v-form instead of stuff on top of v-cards (https://github.com/mitre/heimdall2/issues/5634)
- [ ] You should be able to set the classification level of a results file from the export UI - need to figure out a solution for not just the ckl header level classification but also the individual control level #5689 
- [ ] Source should not be hardcoded to STIG.DOD.MIL - hard to determine when it does come from a STIG vs not #5690 
- [ ] CCI Schema exists - could use this schema to create a JSONIX object instead of using the fast-xml-parser
- [ ] change 'Heimdall Version' to 'MITRE SAF OHDF Converters Version' (https://github.com/mitre/heimdall2/issues/4523)
- [ ] New CKLB is JSON now (https://github.com/mitre/heimdall2/issues/5603)
- [ ] Explore the nist mapper. It doesn't seem to work properly #4604


libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
```diff
@@ -112,63 +150,127 @@ function getStatus(input: string): ExecJSON.ControlResultStatus {
...
+
+function getCopyright(input: string): string {
+  if (!input) {
+    return '';
+  } else {
+    return JSON.parse(input).hdfSpecificData?.copyright;
+  }
+}
+
+function getCopyrightEmail(input: string): string {
```
Hookwitz, line 258: Not sure if these three functions can be or need to be simplified/generic.

Amndeep7:
pushing up my changes @em-c-rod with debugging left in - i'll remove it later

Amndeep7:
there are several disparities between the hdf sample and the stigviewer sample provided in #4047.  one thing that was discussed was the vuln_discuss attr for the single control in there.

```
# stigviewer.ckl
					<ATTRIBUTE_DATA>Enabling Lockdown Mode disables direct access to an ESXi host, requiring the host to be managed remotely from vCenter Server. This is done to ensure the roles and access controls implemented in vCenter are always enforced and users cannot bypass them by logging on to a host directly. By forcing all interaction to occur through vCenter Server, the risk of someone inadvertently attaining elevated privileges or performing tasks that are not properly audited is greatly reduced.

Satisfies: SRG-OS-000027-VMM-000080, SRG-OS-000123-VMM-000620</ATTRIBUTE_DATA>
```

```
# inspec.json
                "desc": "Enabling Lockdown Mode disables direct access to an ESXi host,\nrequiring the host to be managed remotely from vCenter Server. This is done to\nensure the roles and access controls implemented in vCenter are always enforced\nand users cannot bypass them by logging on to a host directly. By forcing all\ninteraction to occur through vCenter Server, the risk of someone inadvertently\nattaining elevated privileges or performing tasks that are not properly audited\nis greatly reduced.",
```

Loading the samples in heimdall and then running it through the reverse mapper results in exactly the same input being returned.

Amndeep7:
I remember us having an extended discussion amongst ourselves and the vmware folks about the control id vs gid issue.  I believe our conclusion was to support both and let the user decide what they wanna do, which is how I've implemented it - thereby completing 4047.

Amndeep7:
image

Latest/greatest UI

Would appreciate letting it be clunky so long as the functionality is there for everything that we need.

@em-c-rod and @ejaronne - please go ham with the testing, just keep in mind that unless its a showstopper, it'll probably be done in a phase 2 after I get back unless someone else wants to tackle it.

if this is good for merge, then i'll do a pass and get rid of all the console log stuff.  I'm leaving them in for now just to make debugging easier if yall do find any significant problems.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -13,28 +13,191 @@
...
+                            label="Release Number"
+                            type="number"
+                            :placeholder="profile.releasenumberplaceholder"
+                          />
+                          <label for="release-date-datepicker">
+                            Release Date
+                          </label>
+                          <v-date-picker
+                            id="release-date-datepicker"
+                            v-model="profile.releasedate"
```
em-c-rod, line 183: Is there a way to have the year attached to the calendar and not populated unless picked on the left side? Because it only updates the year when selecting a date if I am scrolling through the months. Can be in v2 if that will be time consuming.

Amndeep7: I'm not 100% sure what you mean.  It says the current year there since the calendar needs a date to be centered around on initialization.  If you export the ckl file, it doesn't actually populate any values so the releaseinfo string is still just the release number.
image
^ this is from exporting the rhel7 baseline in samples with no modifications in the export modal

Amndeep7:
@em-c-rod your last commit there popped up to the top of my inbox and caught my eye so wanted to comment real fast.  Don't believe everything sonarqube says.  It will always force to go onto ?? over || even when it's not actually correct.  In this case, the default value coming from the user supplied metadata section is 0.  This will consequently always override the value coming from a previously ckl file due to how ?? works.  This also won't be caught by any of the currently existing tests so you might wanna add that test case.

PR #4847, opened 2023-08-20
Release issues aug23

Amndeep7:


Amndeep7:
long term solution is to move to consistently using a datetime library whenever interacting with dates.  we have a mix of plain js and momentjs.  momentjs is no longer being actively maintained so we should find a replacement library.  their official recommendation is luxon.  luxon is able to (eventually) find the correct format type.

image

PR #4892, opened 2023-08-29
Conveyor mapper updates

rbogren-brock:
Updates to several fields for the conveyor-HDF mapper. Changes made to meet the SwAP use case.


libs/hdf-converters/package.json
```diff
@@ -63,7 +63,7 @@
     "quicktype": "^15.0.260",
     "ts-jest": "^29.1.0",
     "ts-node": "^10.0.0",
-    "typedoc": "^0.25.0"
+    "typedoc": "^0.24.1"
```
Amndeep7, line 66: why the regression in version number?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -10,16 +10,23 @@ const CONVEYOR_MAX_SCORE = 1000;
 enum scannerType {
   Moldy = 'Moldy',
   Stigma = 'Stigma',
-  CodeQuality = 'CodeQuality'
+  CodeQuality = 'CodeQuality',
+  Clamav = 'Clamav'
```
Amndeep7, line 14: would prefer to keep the capitalization of the original application: ClamAV

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -10,16 +10,23 @@ const CONVEYOR_MAX_SCORE = 1000;
 enum scannerType {
```
Amndeep7, line 10: might be worthwhile to order these alphabetically at this point

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -67,7 +77,9 @@ function mapSha2Filename(
   }
   return shaMappings;
 }
-
+/*
+Builds the code_desc field using values from data. Varies based on service
```
Amndeep7, line 81: i appreciate the comments

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -142,18 +155,21 @@ function preprocessObject(
...
+*/ 
 function controlMappingConveyor(): MappedTransform<
   ExecJSON.Control & ILookupPath,
   ILookupPath
 > {
   return {
     id: {path: 'sha256'},
-    title: {path: 'filename'},
-    desc: '',
+    title: {path: 'filename'},  // Should be heur_id from conveyor
```
Amndeep7, line 167: there seems to be a discrepancy between the newly added comment and what attribute we've been using here.  which is correct?

libs/hdf-converters/src/conveyor-mapper.ts
```diff
@@ -245,10 +264,12 @@ export class ConveyorResults {
       Object.entries(
         _.get(this.data, 'api_response.results') as Record<string, unknown>
       ) as [string, Record<string, unknown>][]
-    ).map(([scannerName, scannerData]) => [
-      scannerName,
-      new ConveyorMapper(scannerData, this.data, scannerName).toHdf()
-    ]);
+    ).map(([scannerName, scannerData]) => {
```
Amndeep7, line 267: why use a function body when we immediately return a single expression?

Amndeep7 (changes requested):

Amndeep7 (changes requested):
also now that I've enabled the workflows, the linter is bad at you.  you should be able to just run `yarn run lint` and have it autofixed.

rbogren-brock:
Ran linter and addressed comments. No disagreements with any of the suggestions.

libs/inspecjs/package.json
```diff
@@ -15,7 +15,7 @@
     "jest": "^27.0.6",
     "quicktype": "^15.0.260",
     "ts-jest": "^29.1.0",
-    "typedoc": "^0.25.0"
+    "typedoc": "^0.24.1"
```
Amndeep7, line 18: made the change in the package.json for hdf-converters but not here

yarn.lock
Amndeep7, line 1: if we're intentionally not making any changes to the package.json, (i.e. leaving typedoc at 0.25.x) then let's not make any changes to the yarn lock file either.

libs/hdf-converters/src/conveyor-mapper.ts
Amndeep7, line 1: would be great if you could add another test or modify the samples to also include clamav results

Amndeep7 (changes requested):
would appreciate it if you could give this pr a once-over as well so we can merge it in too: https://github.com/mitre/saf/pull/1488/files

Amndeep7 (approved):

PR #4971, opened 2023-09-20
Improving handling of XCCDF fields that may contain HTML

candrews:
Use the stopNodes feature of fast-xml-parser to avoid parsing the HTML in these fields so they can be treated as text. The HTML tags are then stripped from the text as tools expect only text (and not HTML) in these fields.

This PR implements the idea first proposed at https://github.com/mitre/heimdall2/pull/4255#issuecomment-1600856576 which is now possible because the `stopNodes` feature of fast-xml-parser was fixed in version 4.3.0 by https://github.com/NaturalIntelligence/fast-xml-parser/pull/608


Amndeep7:
image

fantastic - that looks a lot better now!  i'll merge when all the tests pass.  could you also make a corresponding PR in the SAF CLI to update the xccdf sample files over there?

Amndeep7 (approved):

Issue #2778, opened 2022-03-09
Request for converter from Grype to HDF


woodsonmiles:
[Grype](https://github.com/anchore/grype) is an open source vulnerability scanner from [Anchore](https://anchore.com/opensource/) for container images similar to Trivy. Grype produces a JSON output.

I've attached an example of a Grype json output created with the command
`Docker run anchore/grype:latest postgres -o json`
[postgres.zip](https://github.com/mitre/saf/files/8217754/postgres.zip)

woodsonmiles:
Reposting notes from our meeting:
https://github.com/mitre/saf/wiki/How-to-recommend-development-of-a-mapper - overarching steps for a mapper

https://saf.mitre.org/#/normalize - SAF page on the naturalization process, has the schema

https://heimdall-lite.mitre.org/ - our deployment of the Heimdall frontend (you can test by uploading here until you get into good enough shape to have a deploy preview running and/or if you wanna test against the latest Heimdall deployment)

https://github.com/mitre/heimdall2/blob/master/libs/inspecjs/src/generated_parsers/v_1_0/exec-json.ts - commented typescript types for the schema

https://github.com/mitre/saf/tree/main/src/commands/convert - the SAF CLI location where you’ll write the integration – please make sure to write a test as well

https://github.com/mitre/heimdall2/blob/master/apps/frontend/src/store/report_intake.ts - the location within Heimdall where you’ll write the fingerprinting

https://github.com/mitre/heimdall2/tree/master/libs/hdf-converters/src - HDF Converters which is where you’ll write the mapper itself – please make sure to write a test as well; you’ll be able to find plenty of examples if you look at the other mappers, the tests, and the sample jsons

https://github.com/mitre/heimdall2/pull/2551 - an example to showcase how the test automation works for linting, etc. and the Netlify deploy preview which will spin up an instance of the Heimdall frontend that will include the changes to HDF Converters as well as the fingerprinting

PR #5025, opened 2023-10-01
Draft: Add converter for grype reports

N1KN1M:
Resolves https://github.com/mitre/heimdall2/issues/2778

Related: https://github.com/anchore/grype/issues/701

Initial changes to standardize grype reports with the heimdall format.

No tests have been written yet.

The mappings and fingerprinting have to be agreed upon first.

### DIFF
--- a/apps/frontend/src/store/report_intake.ts
+++ b/apps/frontend/src/store/report_intake.ts
@@ -15,6 +15,7 @@ import {
   fingerprint,
   FortifyMapper,
   GoSecMapper,
+  GrypeMapper,
   INPUT_TYPES,
   IonChannelMapper,
   JfrogXrayMapper,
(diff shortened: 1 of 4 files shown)


Amndeep7:
Hi @N1KN1M!  Thank you for this contribution!  It looks like you're asking for some guidance on how to move forward with developing this mapper.  Please send me an email at amann@mitre.org, and we can set up a quick call to discuss our typical process and answer any questions of yours.

N1KN1M:
Hi @Amndeep7 , I'm trying to add the grype converter following the guidelines on https://github.com/mitre/heimdall2/issues/2778#issuecomment-1083647384

Since I'm not a core contributor to the grype project, I'd want someone from Grype to agree with the mappings before we proceed on this. 

From what I understand, the tests and the SAF cli integration are the only pieces pending once the fingerprinting and the mappings are agreed upon. I'm thinking of attaching a text containing simple dictionary like mapping to make it easier to read.

Amndeep7:
Hi @N1KN1M, thanks for the work that you've done on this mapper.  I'd really appreciate it if you could write up a quick test and provide some sample files so that I can get a better understanding of the Grype format and ensure that its contents are properly being normalized into OHDF.  If you'd like to reach out to the Grype folks, that'd be great.  I'd first ask if they have a formal schema for their output format, which we could then use here.  After I'm comfortable with the mapper, we can reach out again to confirm that the mapper won't lose anything in the normalization process.

N1KN1M:
@Amndeep7  I'd seen this detailed explanation of each required field in the mapper (https://saf.mitre.org/docs/mapper-background) after creating the initial changes, will ensure the values fit the bill as well as the grype schema before adding tests and taking this further

PR #5128, opened 2023-10-23
Add Threat Column

ejaronne:
DO NOT MERGE AS IS - REMOVES THE RUNTIME COLUMN

Add threat column to control rows (initially just replace Run Time with Related Threats)

Before merging:

- [ ] Re-add runtime column
- [ ] Figure out solution for adding/removing desired columns
- [ ] URLs include user provided values so make sure to sanitize those inputs
- [ ] Come up with a better algorithm for sorting

----

W/r to the control summary

- [ ] Confirm names
- [ ] Remove console logs
- [ ] Flesh out documentation
- [ ] Write saf cli integration
- [ ] Sonarqube doesn't understand when `||` is correct behavior as opposed to `??` so go and correct the warnings in the branch and then again on merge


Amndeep7:
image
got sorting working for threats - currently 'no threats' are considered 'less than' threats so ascending sort has nothing at the top and descending sort goes from greatest to least threat by alphanumeric order / standard string comparison

PR #5235, opened 2023-11-11
Possible Copy Workaround For HTTP

### LINKED ISSUE #5038: [BUG] copy button doesn't work in non-https contexts
@rathbunr reported this

https://developer.mozilla.org/en-US/docs/Web/API/Clipboard

clipboard api is only available in https contexts.  need to add the workaround using the old api when this one doesn't work.

ChrisHinchey:
- [x] Fixes #5038


Amndeep7 (approved):

Amndeep7:
@ChrisHinchey I've approved this cause I believe it should work.  Can you properly test it tho to confirm?  Pretty sure the easiest way to do so would be to spin up a heimdall demo locally (with no changes) or a docker instance with the nginx config / docker compose potentially needing adjustment to expose :80

Amndeep7:
do the research but i'm pretty sure the sonarqube warning can be ignored cause there's no actual replacement for that deprecation

ChrisHinchey:
> do the research but i'm pretty sure the sonarqube warning can be ignored cause there's no actual replacement for that deprecation

Yeah, it seems like the new API just isn't fully implemented so there isn't a way around it at the moment

ChrisHinchey:
> @ChrisHinchey I've approved this cause I believe it should work. Can you properly test it tho to confirm? Pretty sure the easiest way to do so would be to spin up a heimdall demo locally (with no changes) or a docker instance with the nginx config / docker compose potentially needing adjustment to expose :80

I did spin up a heimdall instance locally, so it should be using the unsecured copy, but it seems like it is still using the secure copy.  Not sure if it is a browser thing (tried Firefox and Chrome), but I can't seem to make it use the unsecured version, which is the reason for the note at the top of the PR

Amndeep7:
This functionality is still wrong:

heimdall: clipboard is there but empty 
image

neverssl: clipboard is not there 
image

github: clipboard is there and object is populated with functions 
image

the if statement should probably check both if the clipboard object exists and also if the function we need exists within it

Amndeep7 (changes requested):
@ChrisHinchey please see the above comment

apps/frontend/src/components/generic/CopyButton.vue
```diff
@@ -16,11 +16,34 @@ export default class CopyButton extends Vue {
...
+    // Set the style of the textarea to make it invisible
+    tempTextArea.style.opacity = '0';
+    tempTextArea.style.pointerEvents = 'none';
+    tempTextArea.style.position = 'absolute';
+    tempTextArea.style.left = '-9999px';
+
+    document.body.appendChild(tempTextArea);
+    tempTextArea.focus();
+    tempTextArea.select();
+    document.execCommand('copy');
```
Amndeep7, line 45: need to capture the return value from here and manually throw an error if it's false

"A boolean value that is false if the command is unsupported or disabled."

https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

Amndeep7 (approved):

PR #5236, opened 2023-11-11
Fix Group Empty Array Bug

### LINKED ISSUE #4880: [BUG] empty array creating undefineds in group management page
image

ChrisHinchey:
- [ ] Fixes #4880


Amndeep7 (approved):
Feels like this is symptom resolution as opposed to problem resolution.  The problem is that this modal is nominally created even for groups that don't have any members.  But for those groups, you can't even click on the thing to make this modal pop up in the first place.  So a) why is this modal being created for those group subsections where they shouldn't apply, and b) why are we creating many of these modals at all?  Feels like we should have *the one* modal sitting there, that will get populated with data when a user selects the given group subsection.

image
Empty 'members' userlist within that group which is what's causing this issue I think.

I'm approving it cause at least it removes a superfluous error showing up in the logs, but really I'd like to address it at the deeper level.  So if you decide to merge (which you can), then please make a new issue or write some comments in the original issue (which should be kept open) explaining the above.

@ChrisHinchey

PR #5237, opened 2023-11-13
Quick Fixes For Heimdall Issues

### LINKED ISSUE #4177: Docker compose tries to spin up heimdall even when postgres isn't ready
https://stackoverflow.com/questions/35069027/docker-wait-for-postgresql-to-be-running

Docker compose v2 recently added support for a 'long form' depends_on attribute that lets us wait for postgres to be healthy and not just started before trying to create the server.

Also we can get rid of the version attribute at the top of the file.

### LINKED ISSUE #4367: [BUG] Number of controls "viewed" seems wrong when controls are filtered
image

Mark several controls as "viewed" by using the checkmark on the side.  Apply filters such that at least one of those controls is filtered out as well as all other "unviewed" controls.  You will then get something like the above where the number of controls "viewed" is greater than the total number of controls displayed on the screen.

IMO correct behavior is to make sure that the numerator is not greater than the denominator there, but also mark in certain key spots the number of controls that are hidden by filtering or by selecting the "show only unviewed" button.

### LINKED ISSUE #4466: [BUG] login screen should have a 'display characters' option
similar to the registration page, there should be a button that you can select on the login page that lets you see the characters in your password

### LINKED ISSUE #4962: [BUG] private groups should not be displayed on public files in the db view unless you're already in the group
1. make a file public
2. make a group private
3. log into an account that isn't in that group
4. see all the files and notice how it shows that the private group is listed for that file even tho the user shouldn't know about the existence of that group

ChrisHinchey:
- [x] Fixes #4466 
- [x] Fixes #4367 
- [x] Fixes #4962 
- [x] Fixes #4177 

Checkmark = @Amndeep7's signoff approval on functionality (there might still be comments on improving the code)


apps/frontend/src/components/cards/controltable/ControlTable.vue
```diff
@@ -108,7 +108,11 @@
             text="Controls Viewed"
             sort="disabled"
             :viewed-header="true"
-            :number-of-viewed-controls="viewedControlIds.length"
```
Amndeep7, line 111: image

seems to have broken the counting entirely within the deploy preview where things aren't being hidden after being marked as viewed nor as is the count being adjusted

apps/frontend/src/components/global/login/LocalLogin.vue
```diff
@@ -25,13 +25,18 @@
...
           :error-messages="requiredFieldError($v.password, 'Password')"
-          type="password"
+          :type="showPassword ? 'text' : 'password'"
           name="password"
           label="Password"
           prepend-icon="mdi-lock"
           tabindex="2"
           @blur="$v.password.$touch()"
-        />
+          ><template #append>
```
Amndeep7, line 34: formatting please - the automated stuff looks terrible

apps/frontend/src/mixins/EvaluationMixin.ts
```diff
@@ -1,14 +1,20 @@
+import {ServerModule} from '@/store/server';
 import {IGroup} from '@heimdall/interfaces';
 import {Component, Vue} from 'vue-property-decorator';
 import {IVuetifyItems} from '../utilities/helper_util';
 
 @Component({})
 export default class EvaluationMixin extends Vue {
   convertGroupsToIVuetifyItems(groups: IGroup[] | undefined): IVuetifyItems[] {
+    const currentUserEmail = ServerModule.userInfo.email;
```
Amndeep7, line 9: is identifying the user via email the correct way to uniquely identify the user?  feels like we should be able to easily just grab their id and use that to filter instead.

Amndeep7 (changes requested):

apps/frontend/src/components/global/login/LocalLogin.vue
ChrisHinchey: Good catch.  I don't know why the linter isn't picking those up.  I know we had similar issues with the checklist PR stuff.  It seems like sometimes it catches it and other times it doesn't

apps/frontend/src/mixins/EvaluationMixin.ts
ChrisHinchey: I 100% agree.  Changed accordingly

apps/frontend/src/components/cards/controltable/ControlTable.vue
ChrisHinchey: It was a reactivity issue with Sets, I moved backed to array and implemented a fix.  It should be working now

apps/frontend/src/components/cards/controltable/ControlTable.vue
```diff
@@ -129,7 +129,7 @@
           :control="item.control"
           :expanded="expanded.includes(item.key)"
           :show-impact="showImpact"
-          :viewed-controls="viewedControlIds"
+          :viewed-controls="[...viewedControlIds]"
```
Amndeep7, line 132: does this spread operation into an array still need to be done when we're not doing the set thing anymore?

Amndeep7 (changes requested):

apps/frontend/src/components/cards/controltable/ControlTable.vue
Amndeep7: It's working

Amndeep7:
@ChrisHinchey can you ask @ejaronne about the private groups issue?  I've tested the functionality and it seems like it works, but now there's no way for an admin to know that a private group is attached to a given file.  We should probably add a new tab in the admin panel to manage files.  This should be spun off as a new task and shoved underneath the groups epic imo.

apps/frontend/src/components/global/login/LocalLogin.vue
```diff
@@ -25,13 +25,19 @@
...
           tabindex="2"
           @blur="$v.password.$touch()"
-        />
+        >
+          <template #append>
+            <v-icon @click="showPassword = !showPassword">{{
+              showPassword ? 'mdi-eye' : 'mdi-eye-off'
+            }}</v-icon>
+          </template></v-text-field
+        >
```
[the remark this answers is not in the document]
ChrisHinchey: I still wonder why the linter isn't fixing these

apps/frontend/src/components/cards/controltable/ControlTable.vue
ChrisHinchey: I removed it.  It was left over from previous changes.

apps/frontend/src/mixins/EvaluationMixin.ts
```diff
@@ -1,14 +1,20 @@
...
 
 @Component({})
 export default class EvaluationMixin extends Vue {
   convertGroupsToIVuetifyItems(groups: IGroup[] | undefined): IVuetifyItems[] {
+    const currentUserId = ServerModule.userInfo.id;
     if (!groups) {
       return [];
     }
-    return groups.map((group) => {
+    const validatedGroups = groups.filter((group) =>
```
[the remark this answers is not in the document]
ChrisHinchey: I wasn't aware we were already tracking the groups that the user was in.  Unless I am thinking about something wrong, we wouldn't need the intersection since my logic was just to determine the groups the user was in anyways.

docker-compose.yml
```diff
@@ -23,7 +21,8 @@ services:
     ports:
       - "3000"
     depends_on:
-      - "database"
+      database:
+        condition: service_healthy
```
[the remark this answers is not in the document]
ChrisHinchey: I still need to test to make sure it works as intended, but I added a health check for the database.

ChrisHinchey:
@Amndeep7 To address the admin file management tab I created #5259

docker-compose.yml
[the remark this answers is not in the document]
Amndeep7: @ChrisHinchey reminder to check this please.

apps/frontend/src/mixins/EvaluationMixin.ts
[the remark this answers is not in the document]
Amndeep7: For one of them we just want the groups the user is in.  But for the other we only want the groups that are applied to the evaluation that the user is also in.  Having the function return the set of all groups that the user is in would be incorrect in that case. 

https://github.com/search?q=repo%3Amitre%2Fheimdall2+convertgroupstoivuetifyitems&type=code

Amndeep7:
@ChrisHinchey please see new comments

apps/frontend/src/mixins/EvaluationMixin.ts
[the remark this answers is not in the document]
ChrisHinchey: Ah, I get what you mean. I just pushed up the fix now.

docker-compose.yml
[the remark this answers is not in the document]
ChrisHinchey: Just sent you a message showing that it looks like everything is working properly

Amndeep7 (approved):

PR #5260, opened 2023-11-19
Add Missing HTML Export Static Files

ChrisHinchey:


Amndeep7 (approved):
This will probably work.  Please double check by following a similar process as to the one described here (https://github.com/mitre/saf/wiki/How-to-use-an-unreleased-version-(example:-from-a-branch)-of-HDF-Converters-in-the-SAF-CLI) to validate if it does work in the saf cli before we merge.  Please make a pr and update the wiki to support an unreleased heimdall frontend as well as hdf-converters in teh saf cli.

PR #5298, opened 2023-11-29
Add Better Error Display

### LINKED ISSUE #5226: The error and backtrace should be in the test results and control details
[shell_stig-1.json](https://github.com/mitre/redhat-enterprise-linux-8-stig-baseline/files/13309460/shell_stig-1.json)

Bug in heimdall. Should show the error and backtrace

Screenshot 2023-11-09 at 09 44 27

ChrisHinchey:
- [x] Fixes #5226


apps/frontend/src/components/cards/controltable/ControlRowCol.vue
```diff
@@ -40,6 +40,22 @@
       />
       <!-- eslint-enable vue/no-v-html -->
     </v-col>
+    <v-col v-if="'backtrace' in result" cols="12" class="pa-2">
```
Amndeep7, line 43: image

backtrace expansion thing shouldn't be there if there is no backtrace provided.  i get that you're trying to do that check in the v-if here but obv it's not working properly.

you should just do object.prototype.hasownproperty: https://stackoverflow.com/a/455340/645647

apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
```diff
@@ -10,11 +10,33 @@
...
           <v-tab-item value="tab-test">
             <div class="pa-4">
-              <div v-if="caveat || justification || rationale || comments">
+              <div
+                v-if="
+                  caveat ||
+                  justification ||
+                  rationale ||
+                  comments ||
+                  errorMessage !== ''
```
Amndeep7, line 19: similar to the other ones blocks, unless we're explicitly only checking for empty string, we should probably do the 'string as boolean' thing so that it handles null/undefined as well

apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
```diff
@@ -166,6 +188,12 @@ export default class ControlRowDetails extends mixins(HtmlSanitizeMixin) {
     return this.control.hdf.descriptions.comments;
   }
 
+  get errorMessage(): string {
+    return this.control.hdf.segments?.length == 0
+      ? "The control didn't return any reults.  Check with the author of the profile to ensure the code is correct."
```
Amndeep7, line 193:
```suggestion
      ? "The control didn't return any results.  Check with the author of the profile to ensure the code is correct."
```

Amndeep7 (changes requested):

apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
ChrisHinchey: Since we aren't checking for a value from the `control.hdf` object and rather a custom string based on a condition, wouldn't we want to still check for an empty string unless we make the fail condition a null/undefined

aaronlippold:
This has been very useful over the last few days. Hopefully we can get it merged during this release.

aaronlippold:
Related https://github.com/mitre/saf/issues/2074

aaronlippold:
Related https://github.com/inspec/inspec/issues/6900

apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
Amndeep7: Re-reading the code, I see what you mean w/r to that function basically returning an enum of [warning | empty-string].  However, in that case, it makes even more sense to just do v-if="stuff || comments || errorMessage" cause "errorMessage !== ''" is semantically equivalent to "errorMessage (as a boolean expression)" due to the empty string being falsy.

Amndeep7 (changes requested):

Amndeep7 (approved):
I'd appreciate it if you made that stylistic change but it's not the most important thing so I'mma approve this pr.  Merge it in after you make your choice.  If you choose not to make the change, you're gonna need to re-run the e2e test until it decides to pass before applying the ready-to-merge label.

ChrisHinchey:
I don't mind making the stylistic change, I just wasn't sure if we wanted to make it explicit since it is handled differently from the rest. Just pushed the change

PR #5437, opened 2024-01-13
Fix error message incorrectly showing up for overlay profile results

### LINKED ISSUE #5418: ERROR message incorrectly shows for overlay profile results
Probably introduced during PR https://github.com/mitre/heimdall2/pull/5298 

It is incorrectly casting the ERROR message for items in overlays. It is probably seeing the "results": [] that appears in the additional profile objects that show up in an overlay profile result set. However, all actual results are consolidated in baseline profile object! Can be seen in built in samples:

image

image

image

image

image

Amndeep7:
Resolves #5418 

Still showing up in actual profile error situations

Sample file from original issue: https://github.com/mitre/heimdall2/issues/5226

image

No longer showing up in overlays

Sample file from new issue: https://github.com/mitre/heimdall2/issues/5418

image

### DIFF
--- a/apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
+++ b/apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
@@ -189,7 +189,7 @@ export default class ControlRowDetails extends mixins(HtmlSanitizeMixin) {
   }
 
   get errorMessage(): string {
-    return this.control.hdf.segments?.length == 0
+    return this.control.root.hdf.segments?.length == 0
       ? "The control didn't return any results.  Check with the author of the profile to ensure the code is correct."
       : '';
   }


apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
```diff
@@ -189,7 +189,7 @@ export default class ControlRowDetails extends mixins(HtmlSanitizeMixin) {
   }
 
   get errorMessage(): string {
-    return this.control.hdf.segments?.length == 0
+    return this.control.root.hdf.segments?.length == 0
```
aaronlippold, line 192: The old guy still has good instincts :)

wdower (approved):

apps/frontend/src/components/cards/controltable/ControlRowDetails.vue
ejaronne: Just curious: how does it determine what is a "root" profile object in the results json? Is it based on any that do not contain a dependency?

PR #5574, opened 2024-02-26
Node1819plus

Amndeep7:


.github/workflows/draft-release.yml
```diff
@@ -24,18 +24,18 @@ jobs:
       # write permission is required for autolabeler
       # otherwise, read permission is required at least
       pull-requests: write
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
georgedias, line 27: Is there an issue if we use latest! or do we want to force to use ubuntu 22.04 for the draft release?

georgedias (commented):
Just one question on the draft-release.yml

.github/workflows/draft-release.yml
Amndeep7: We're using ubuntu 22.04 for all the rest of the actions.

Amndeep7:
The reason why we are using Ubuntu 22.04 vs ubuntu-latest is that we ran into versioning issues in the past with the tools that were provided in the image.  If we hardlock on a particular OS version then we can test each updated image for the runner as they show up instead of eating it when we try to do a release or something.

@em-c-rod @georgedias

PR #5588, opened 2024-02-28
modify default case to conform to format defined in CLI

DMedina6:
Original implementation uses the ms() method to convert frequency if one of the other cases isn't applicable. Re-implemented to conform to standard specified in SAF CLI (<x>d/wk/m/yr) using regular expressions

Closes: https://github.com/mitre/saf/issues/2067


Amndeep7 (changes requested):

libs/hdf-converters/src/utils/attestations.ts
```diff
@@ -50,11 +50,26 @@ export function advanceDate(
       date.add(1, 'day');
       break;
     default:
-      const msTime = ms(frequency);
-      if (!msTime) {
-        throw new Error('Unknown date format: ' + frequency);
-      } else {
-        date.add(msTime, 'milliseconds');
+      const match = frequency.match(/(\d+(?:\.\d+)?)([a-z]+)/);
```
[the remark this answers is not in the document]
Amndeep7: To be clear, I'm fine with changing the SAF CLI prompt to better explain the syntax we desire (so both the phrases defined above as well as whatever syntax the external library uses).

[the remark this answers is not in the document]
Amndeep7: This review is no longer applicable.  Please see our chat for the new direction.

libs/hdf-converters/src/utils/attestations.ts
```diff
@@ -50,11 +50,29 @@ export function advanceDate(
...
+      
+      // throw error if input doesn't match specified format of <x>d/wk/m/yr
+      if (!match){ throw new Error('Unknown date format: '+frequency); };
+
+      const freq_num = match[1];
+      const unit = match[2];
+
+      switch (match[2]) {
```
Amndeep7, line 61: this switch is unnecessary and in fact will lead to us throwing an error even when the user puts in something like the word 'days' themselves.  imo let's just feed whatever we receive as input from the user into the library and catch -> throw errors from that library but with a nicer error statement probably.

also now that we're throwing this error, we should probably have good error handling wherever this function is called.

Amndeep7 (changes requested):

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -110,4 +110,28 @@ describe('attestations', () => {
       advanceDate(moment.utc(1662758942000), 'fortnightly').toISOString(true)
     ).toEqual('2022-09-23T21:29:02.000+00:00');
   });
+
+  it('Should parse input for custom number of days correctly', () => {
```
em-c-rod, line 114: Can you add clauses in the attestation array that makes it clear the user input will be understood as expected?

libs/hdf-converters/sample_jsons/attestations/rhel7-json-attestedjson.json
```diff
@@ -7999,8 +7999,22 @@
...
+              "start_time": "2024-03-12T15:51:51.580Z"
             }
-          ]
+          ],
+          "attestation_data": {
+            "control_id": "V-72083",
+            "explanation": "The operating system off-loads audit records onto a different system from the system being audited.",
+            "frequency": "7w",
+            "status": "passed",
+            "updated": "2024-03-12",
```
em-c-rod, line 8015: How do we want to handle these tests - would they fail in 8 weeks, for example? Because of the date?

libs/hdf-converters/src/utils/attestations.ts
```diff
@@ -49,14 +48,38 @@ export function advanceDate(
...
+      if (!match) {
+        throw new Error(
+          'Unknown date format: ' +
+            frequency +
+            '. Please use a number followed by d/w/m/y to indicate days, weeks, months, or years, e.g. 1d/2w/3m/1y/custom.'
+        );
+      }
+
+      const number = match[1];
+      const unit = match[3];
```
em-c-rod, line 65: How does the 1 and 3 index here work? I want to make sure we are not making any assumptions about number of days or spaces or something, which may be taken care of by the regex, but I'm not sure. For example, '120 d' or '15       w'

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -27,6 +27,31 @@ const attestationArray: Attestation[] = [
     status: ExecJSON.ControlResultStatus.Passed,
     updated: '2022-01-02',
     updated_by: 'Alec Hardison, Security'
+  },
+  {
+    control_id: 'V-72083',
+    explanation:
+      'The operating system off-loads audit records onto a different system from the system being audited.',
+    frequency: '7w',
```
em-c-rod, line 35: Can you make some of these test inputs double or triple digit numbers?

libs/hdf-converters/src/utils/attestations.ts
DMedina6: So the regular expression allows for 0-10 number of spaces between the number and the unit, and the space (or lack of spaces) would be match[2], and match [1] is the number and match [3] is the unit which also have character limits because the pipeline kept saying it was a security concern as somebody could enter a ton of zeros and DDoS the system

libs/hdf-converters/sample_jsons/attestations/rhel7-json-attestedjson.json
DMedina6: The ones that aren't meant to expire put the current date in the updated field:     
updated: moment().format('YYYY[-]MM[-]DD')

But now that I think about it, the files the tests write to with writefilesync (those commented out lines) won't have the updated dates unless they are uncommented and made to run again. Realistically the answer to that would be to uncomment the lines and have them run every time (which wouldn't affect anything so i'm not sure why they're commented out anyhow)

libs/hdf-converters/src/utils/attestations.ts
```diff
@@ -181,9 +229,15 @@ function getFirstPath(
...
-    return _.get(object, paths[index]) as string;
   }
+  const stringOrDate = _.get(object, paths[index]);
+  if (_.isString(stringOrDate)) {
+    return stringOrDate;
+  }
+  //return _.get(object, paths[index]);
+  // Do we care about time?
+  console.log('stringorDate: ' + stringOrDate);
+  return `${stringOrDate.getFullYear()}-${String(stringOrDate.getMonth() + 1).padStart(2, '0')}-${String(stringOrDate.getDate()).padStart(2, '0')}`;
```
em-c-rod, line 240: This change is not clear to me. Can you explain more about this section?

libs/hdf-converters/test/attestations/attestations-old.spec.ts
```diff
@@ -0,0 +1,168 @@
+import fs from 'fs';
```
em-c-rod, line 1: You can delete this file in favor of your new tests

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
+    control_id: 'V-73165',
+    explanation: 'This is a failing attestation that is expired',
+    frequency: '1d',
+    status: ExecJSON.ControlResultStatus.Failed,
+    updated: '1999-01-01',
+    updated_by: 'Jane Doe'
+  }
+];
+
+const aPass_cSkip_unexp: Attestation[] = [
```
em-c-rod, line 51:
```suggestion
const validPassingAttestationForSkippedControl: Attestation[] = [
```

I recommend a more spelled out name. I'm not worried about short variable names here. We have the real estate and it's easier to understand something more spelled out. The one I listed is only a suggestion. Feel free to modify as you choose. (In this example I used valid for not expired, and would have put expired for expired, but you can brainstorm that, too.)

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
+  {
+    control_id: 'SV-230223',
+    explanation:
+      'This is a test unexpired passing attestation for a skipped control',
```
em-c-rod, line 55:
```suggestion
      'This control passes according to this attestation',
```

Again, take my suggestion as only one way to do it, but consider what exactly this explanation represents and write it to that.

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
+    explanation:
+      'This is a test expired passing attestation for a skipped control',
+    frequency: '1d',
+    status: ExecJSON.ControlResultStatus.Passed,
+    updated: '1999-01-01',
+    updated_by: 'Attestation Tester'
+  }
+];
+
+// Naming convention: a<Pass/Fail>_<c<Pass/Fail/Skip>>_<unexp/exp>, so attestation<status>_control<status>_<unexpired/expired>
```
em-c-rod, line 74: No need for comments if the names are already self explanatory :) Make the reader not have to search for this decoding to understand

libs/hdf-converters/src/utils/attestations.ts
DMedina6: Excel stores dates as Date objects and they don't format nicely when converting to a string using sheetjs. This returns a string version of the 'updated' field with the padding needing to have it resemble what the data would otherwise look like coming from normal attestation files.

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
+
+  it('Should return correct date when given custom number of years to advance by', () => {
+    expect(
+      advanceDate(moment.utc(1662758942000), '5y').toISOString(true)
+    ).toEqual('2027-09-09T21:29:02.000+00:00');
+  });
+});
+
+describe('CreateAttestationMessage', () => {
+  it('Should create a message for an unexpired attestation', () => {
```
em-c-rod, line 163: What do you mean by message? Is this a field that is added to the data? Where does it show up? What does it represent? I would add a little more clarity here.

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
+  parseXLSXAttestations,
+  convertAttestationToSegment
 } from '../../src/utils/attestations';
 import {omitHDFTimes, omitVersions} from '../utils';
-
-const attestationArray: Attestation[] = [
+import _ from 'lodash';
+// test a new attestation array for each frequency, use arbitrary far date (yr 3000, 1900)
+// try applying to attestations w diff statuses
+const attestationMessageArray: Attestation[] = [
```
em-c-rod, line 16: Is it easier to also break these down into individual attestation arrays of one item with more meaningful names? When you call `attestationMessageArray[0]` in the tests, it doesn't tell me so much.

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
+    console.error = jest.fn();
+  });
+
+  afterEach(() => {
+    console.error = consoleOriginal;
+  });
 
-    //fs.writeFileSync('sample_jsons/attestations/rhel7-json-attestedjson.json', JSON.stringify(output, null, 2))
+  /**
+     * Adding an attestation results in HDF object having an additional result object in controls.results array
```
em-c-rod, line 258:
```suggestion
     * Adding an attestation creates an additional result object in controls.results array in the HDF object 
```

Using a word like `results` here as a verb gets confusing because we are also using it so much as a noun, so I would just restate to something like this.

libs/hdf-converters/test/attestations/attestations.spec.ts
```diff
@@ -6,108 +6,371 @@ import {
...
 
-    expect(omitHDFTimes(omitVersions(output))).toEqual(
-      omitHDFTimes(omitVersions(expected))
+    // Check that the results array has one additional entry
+    expect(output.profiles[0].controls[2].results.length).toEqual(2);
+    // Check that the status of the new result is passing
+    expect(output.profiles[0].controls[2].results[1].status).toEqual(
+      ExecJSON.ControlResultStatus.Passed
+    );
+    // Check that the attestation data added to the control is the equal to the attestation passed into the function
```
em-c-rod, line 300:
```suggestion
    // Check that the attestation data added to the control is the attestation passed into the function
```

libs/hdf-converters/sample_jsons/attestations/attestations-2.xlsx
em-c-rod, line 1: Can you give this a more descriptive name? And just to make sure, are you using it?

libs/hdf-converters/sample_jsons/attestations/sample_input_report/rhel8_sample_more.json
em-c-rod, line 1: Can you give this a more descriptive name?

libs/hdf-converters/src/utils/attestations.ts
```diff
@@ -2,19 +2,23 @@ import * as XLSX from '@e965/xlsx';
...
+
+// export type Attestation = {
+//   control_id: string;
+//   explanation: string;
+//   frequency: string;
+//   status: 'passed' | 'failed';
+//   updated: string;
+//   updated_by: string;
+// };
+// Convert from using enum type to enum values
```
em-c-rod, line 19: I don't understand this change. Can you explain?

libs/hdf-converters/test/attestations/attestations.spec.ts
DMedina6: Sure. It does seem clearer to use valid/expired vs unexpired/expired since that's easier to distinguish while reading. I went with validPassingAttestation_passingControl with an underscore between the attestation and control descriptors

libs/hdf-converters/test/attestations/attestations.spec.ts
DMedina6: That makes sense. It should at least be written like one would write an actual attestation, as opposed to just putting whatever in there, in order to make the test clear and understandable to a user or developer.

libs/hdf-converters/test/attestations/attestations.spec.ts
DMedina6: Removed the comment and changed the naming convention to not need an explanation.

libs/hdf-converters/test/attestations/attestations.spec.ts
DMedina6: Would it be be more clear if I re-arrange the order of the tests being conducted @em-c-rod? Seeing what addAttestationToHDF() does and the comment I add that describes the HDF json structure for controls and attestations might make the rest more clear if that's read first

libs/hdf-converters/test/attestations/attestations.spec.ts
DMedina6: Switching to not use this array of attestations entirely in favor of the more clearly named individual attestations below

libs/hdf-converters/src/utils/attestations.ts
DMedina6: The AttestationData interface is what this file should've been using, but the original creator wanted to use strings instead of ControlAttestationStatus enums, so they made their own class for it with a slightly different name. This changes it so it actually uses that original AttestationData while still replacing the status with a string as defined by the ControlAttestationStatus enum:

export enum ControlAttestationStatus {
    Failed = "failed",
    Passed = "passed",
}

I may be misunderstanding this feature though. Credit to @Amndeep7 for the suggestion and change honestly

libs/hdf-converters/sample_jsons/attestations/attestations-2.xlsx
DMedina6: I was using it. I went with attestations_xlsxInputFormat and removed any files I wasn't using.

libs/hdf-converters/sample_jsons/attestations/sample_input_report/rhel8_sample_more.json
DMedina6: I used rhel8_sample_oneOfEachControlStatus.json if that's more clear

em-c-rod (approved):
The fixed functionality and added tests will make this much more robust. Thanks for the work.

Issue #5541, opened 2024-02-14
Create HTML reports based on filtered results in Heimdall


em-c-rod:
**Is your feature request related to a problem? Please describe.**
When creating HTML reports from results in Heimdall, the exported information shows either all or no results.

**Describe the solution you'd like**
I want to export reports of only the failures in a set of results, for example, or whatever I have filtered in the results view.

PR #5595, opened 2024-03-01
HTML Reports based on filter data

georgedias:
Fixes issue #5541


libs/hdf-converters/package.json
```diff
@@ -6,7 +6,7 @@
   "files": [
     "lib"
   ],
-  "main": "src/index.ts",
```
Amndeep7, line 9: This should not be changed.

certs/ssl_certificate.crt
Amndeep7, line 1: delete these certs.  you will probably want to regenerate them for whatever you were using them for.

Amndeep7 (changes requested):

libs/hdf-converters/package.json
em-c-rod: updated :+1:

certs/ssl_certificate.crt
em-c-rod: fixed

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -119,22 +119,60 @@ export default class ExportHTMLModal extends Vue {
...
+         *         return filteredControls.includes(control.id);
+         *       }
+         *     });
+         * )
+         * the contextualize object does not get updated and the results
+         * in data_store get out of sync, there is when utilizing the
+         * ".contains" it returns ann results where the file.evaluations
+         * returns the filtered controls.
+         */
+        let filteredControls: string[] = [];
```
Amndeep7, line 149: the way that this is implemented will not satisfy the issue where it was stated that we should be able to handle any of the filters that the user specifies.  this will only filter on pass/fail/whatever.

https://github.com/mitre/heimdall2/blob/dc5cddd7ae0f0c396044e91350044ff51ab00db0/apps/frontend/src/components/global/ExportCaat.vue#L43

use this example for how to get all the filter-passing controls.  it should also mean that we don't need basically all of what're currently lines 122-123 and 150-164 since it'll just be a simple assignment directly onto 'filteredControls' (which can also then be made const).

Amndeep7 (changes requested):

libs/hdf-converters/src/converters-from-hdf/html/reverse-html-mapper.ts
```diff
@@ -29,12 +29,14 @@ type InputData = {
   data: ContextualizedEvaluation | string;
   fileName: string;
   fileID: string;
+  filteredControls: string[];
```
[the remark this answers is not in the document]
georgedias: I have no problems making this optional, but I don't think this is used by the SAF CLI! Well at least, I was not under that impression.

libs/hdf-converters/src/converters-from-hdf/html/reverse-html-mapper.ts
```diff
@@ -324,7 +335,9 @@ export class FromHDFToHTMLMapper {
           100
       );
       // Set compliance level
-      this.outputData.compliance.level = complianceLevel;
+      this.outputData.compliance.level = complianceLevel.includes('NaN')
```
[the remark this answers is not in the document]
georgedias: Look at the logic on how the percentages are being calculated, then image that you only have 1 say Nor Review, or failed, you will get a divide by zero.

libs/hdf-converters/src/converters-from-hdf/html/reverse-html-mapper.ts
```diff
@@ -201,7 +208,11 @@ export class FromHDFToHTMLMapper {
     const allResultLevels: ContextualizedControl[] = [];
     file.data.contains.map((profile) => {
       profile.contains.map((result) => {
-        allResultLevels.push(result);
+        file.filteredControls.forEach((element) => {
```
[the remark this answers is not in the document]
georgedias: Sure I can do that, however that is a personal preferences (map to flatMap), aside for the for...of which is an ES6, vs the forEach which is an ES5.

apps/frontend/src/components/global/ExportHTMLModal.vue
georgedias: Absolutely unequivocally disagree that the implementations does not satisfy the requirements. It works on all possible filtered permutations, did you even tried, apparently not if you're making this statement. 
_"the way that this is implemented will not satisfy the issue where it was stated that we should be able to handle any of the filters that the user specifies. this will only filter on pass/fail/whatever."_

Regarding the second opinion, the code that is being referenced (from the ExportCaat.vue) will not work. If you take a look at the reverse-html-mapper.ts it uses the file.data, which is the file.evaluation being set in ExportHTMLModule.vue and you are comparing to this code in the ExportCaat,vue `const data = file?.evaluation ?? '';`  now how does that filters the proper evaluations? it doesn't. The code on both are accomplishing different tasks.

georgedias (commented):
To be fair, the UC was to show report based on filtering data aka, Passed, Failure, etc. not filtered on NIST controls, there is, we’re just filtering on STIG control status.
For future implementation for the HTML Export process I proposed the format provided in issue #5596

Amndeep7:
"I want to export reports of only the failures in a set of results, for example, or whatever I have filtered in the results view."

I read this as supporting any/all filtering, but I could be wrong.

@em-c-rod can you please provide input?  You saying that you want to match the behavior of the other exports is unfortunately not usable input since there is unfortunately a wide variety of behaviors that occur.

Amndeep7:
I'm not gonna make it a blocker, but it would be nice if we could actually do the filtering process for an execjson within the store so that we don't need to pass around a set of controls at all.  This'll make both the mappers easier to implement due to them not needing to understand any filtering concepts at all and heimdall integrations with these mappers easier to write and overall less duplicative due to just needing to call the one store function instead of each mapper needing to implement filtering in its own special way.

libs/hdf-converters/src/converters-from-hdf/html/reverse-html-mapper.ts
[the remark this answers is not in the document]
Amndeep7: > Formula: compliance = Passed/(Passed + Failed + Not Reviewed + Profile Error) * 100

If we only have 1 reviewed or failed, there shouldn't be a divide by 0.  I could see division by 0 occurring if you managed to filter out all results though so I guess this validation check is alright.

libs/hdf-converters/src/converters-from-hdf/html/reverse-html-mapper.ts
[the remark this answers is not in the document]
Amndeep7: map to flatMap is more than just a syntactic difference.

in any case, it's just a consistency in coding paradigm fix - we're doing functional style "loops" with the maps, but then all the way inside we do a for loop around a procedural line where we are populating an array defined outside of all of these loops.

imo we should either go full procedural and use actual loops for everything or go full functional instead of this mishmash.  my simplification recommendation was to do the functional solution.

apps/frontend/src/components/global/ExportHTMLModal.vue
Amndeep7: image

image

Here I've filtered it down to a single control using the nist filters.  The html export does not only include that single control.  Consequently, it does not meet the requirement of the original ticket according to my reading of it.  Like I said elsewhere, we should get @em-c-rod or @ejaronne's opinions on which of our interpretations is correct.

Amndeep7: w/r to your second paragraph, I'm only talking about extracting out the set of controls that pass the filtering.  You can use the store to get that set I'm hoping.

em-c-rod: This may be a use case, but the ask for this is just to be able to show only the failed controls (aka filter on status). If this becomes an additional ask, we can make another PR.

em-c-rod: I do think it is relevant to consider severity filters here. Looking at severity filters as well was not explicitly the ask, but seems like it is paired in the same use case.

aaronlippold:
Adding methods or computed properties into the store - from the docs - seems to be the intended implemntation in the Vuex world. https://v3-migration.vuejs.org/breaking-changes/filters.html 

Given that we will eventually move to https://pinia.vuejs.org/core-concepts/getters.html moving the processing to the backend / middleware seems a good way to be able to reuse it with other parts of the GUI.

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -33,6 +33,16 @@
             <pre class="pt-5" v-text="description" />
           </v-col>
         </v-row>
+        <v-row>
+          <v-col cols="1">
+            <v-icon color="primary">mdi-information-variant-circle</v-icon>
+          </v-col>
+          <v-col>
+            Both Manager and Administrator Reports can be very data intensive,
```
em-c-rod, line 41:
```suggestion
            Both Manager and Administrator Reports can be data intensive,
```

apps/frontend/src/components/global/ExportHTMLModal.vue
```diff
@@ -33,6 +33,16 @@
...
           </v-col>
         </v-row>
+        <v-row>
+          <v-col cols="1">
+            <v-icon color="primary">mdi-information-variant-circle</v-icon>
+          </v-col>
+          <v-col>
+            Both Manager and Administrator Reports can be data intensive,
+            does consuming computer memory resources, which can take longer for
+            the browser to render the page.
```
em-c-rod, line 43:
```suggestion
```

em-c-rod (approved):

PR #5623, opened 2024-03-12
Update README.md

em-c-rod:


Amndeep7 (approved):
A similar warning will probably be required for the HDF Converters tests

PR #5686, opened 2024-03-29
Add Dependency-Track FPF Mapper

jbarayuga:
Creating a new mapper to be able to convert Dependency-Track's [Finding Packaging Format](https://docs.dependencytrack.org/integrations/file-formats/) (FPF) files to HDF


jbarayuga:
Some commits added without `--signoff` option. Commit history is short so I will close this and redo in a new PR

Issue #5719, opened 2024-04-11
Okta Integration Bug


em-c-rod:
**Describe the bug**
Broken Okta integration

**To Reproduce**
1. Deploy 2.10.4 Heimdall with Okta login
2. Load the login page
3. Click "login with Okta"
4. Try to login
5. Error: "Sorry, an problem occurred while signing you in. The reason given was: Cannot destructure property 'given_name' of 'userData' as it is undefined."

**Expected behavior**
Login successfully with no errors.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 2.10.4

**Additional context**
Related to these dependency updates but the okta.strategy.ts file was not updated along with the oidc.strategy.ts file.
- https://github.com/mitre/heimdall2/pull/5512
- https://github.com/mitre/heimdall2/pull/5548

PR #5725, opened 2024-04-12
Update okta.strategy.ts to work with passport-openidconnect 0.1.1

em-c-rod:
Closes: #5719 

To test:
1. Setup Okta development server
2. Integrated local heimdall build with Okta dev server via App integrations
3. Stored the client id, client secret, external url, and okta domain as env variables for heimdall
4. Built Heimdall using the docker build to allow for https (http does not work with the okta integration)
5. Created users in okta and assigned them to Heimdall, had to assign them a username for it to work correctly)
6. Also had to disable GP to test the integration successfully

Added documentation here for testing Okta integration - https://github.com/mitre/heimdall2/wiki/Heimdall-Authentication-Methods#testing

### DIFF
--- a/apps/backend/src/authn/okta.strategy.ts
+++ b/apps/backend/src/authn/okta.strategy.ts
@@ -44,6 +44,7 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
         scope: 'openid email profile'
       },
       async function (
+        // Like in oidc.strategy.ts, we changed the arity of the function to 9 to access the data we need from 'uiProfile' due to updates in the passport-openidconnect library which otherwise caused failures in the authentication process
         _issuer: string,
         uiProfile: OktaProfile,
         _idProfile: object,
--- a/apps/backend/src/authn/okta.strategy.ts
+++ b/apps/backend/src/authn/okta.strategy.ts
@@ -41,7 +41,7 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
         clientID: configService.get('OKTA_CLIENTID') || 'disabled',
         clientSecret: configService.get('OKTA_CLIENTSECRET') || 'disabled',
         callbackURL: `${configService.get('EXTERNAL_URL')}/authn/okta/callback`,
-        scope: 'openid email profile',
+        scope: 'openid email profile'
       },
       async function (
         _issuer: string,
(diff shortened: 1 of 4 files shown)


georgedias (commented):
Emily, during our testing Amndeep provided instructions to update the docker builds where we used this block of code.
`RUN if curl -o install_certs.sh https://gitlab.mitre.org/mitre-scripts/mitre-pki/raw/master/os_scripts/install_certs.sh; then sh install_certs.sh centos && rm -f install_certs.sh; else echo "Not on the MII, skipping certificate install"; fi`

Do we need to add there instructions, or will the docker file be modified to contain this code?

apps/backend/src/authn/okta.strategy.ts
```diff
@@ -35,24 +40,26 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
...
+        scope: 'openid email profile'
       },
-      (
-        _req: Request,
-        _token: string,
-        _tokenSecret: string,
-        profile: OktaProfile,
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      async function (
+        // Like in oidc.strategy.ts, we changed the arity of the function to 9 to access the data we need from 'uiProfile' due to updates in the passport-openidconnect library which otherwise caused failures in the authentication process
```
georgedias, line 47: Can we please state the meaning of the underscore (_) variables, thanks.

georgedias (changes requested):

DMedina6:
> Emily, during our testing Amndeep provided instructions to update the docker builds where we used this block of code. `RUN if curl -o install_certs.sh https://gitlab.mitre.org/mitre-scripts/mitre-pki/raw/master/os_scripts/install_certs.sh; then sh install_certs.sh centos && rm -f install_certs.sh; else echo "Not on the MII, skipping certificate install"; fi`
> 
> Do we need to add there instructions, or will the docker file be modified to contain this code?

I was wondering the same thing, but I thought these changes were supposed to allow us to test behind the MITRE proxies, which isn't relevant to the customers / sponsors deploying this application.

apps/backend/src/authn/okta.strategy.ts
em-c-rod: Good call! Added it.

em-c-rod:
> > Emily, during our testing Amndeep provided instructions to update the docker builds where we used this block of code. `RUN if curl -o install_certs.sh https://gitlab.mitre.org/mitre-scripts/mitre-pki/raw/master/os_scripts/install_certs.sh; then sh install_certs.sh centos && rm -f install_certs.sh; else echo "Not on the MII, skipping certificate install"; fi`
> > Do we need to add there instructions, or will the docker file be modified to contain this code?
> 
> I was wondering the same thing, but I thought these changes were supposed to allow us to test behind the MITRE proxy, which isn't relevant to the customers / sponsors deploying this application.

Correct! We don't need to add it here.

georgedias (approved):

Issue #2375, opened 2022-01-07
Change text for some export options


Amndeep7:
'Export as JSON' should be 'Export as HDF JSON'

'Export as Checklist' should be 'Export as DISA Checklist (CKL)'

S0han:
assign this to me, I can take a look

camdenmoors:
> assign this to me, I can take a look

Thanks for the interest in the project, please fork Heimdall2 and create a pull request. All PRs are welcome.

Amndeep7:
Completed at some point

PR #5852, opened 2024-05-23
Add option to normalize export ckl profile name

georgedias:
Provide a checkbox that normalizes (proper formats) the profile title (name). Also fixes issue # #2375

Example, if the name is: `redhat-enterprise-linux-8-stig-baseline`
It normalizes to:  'Redhat Enterprise Linux 8 Security Technical Implementation Guide`

### Logic ###
Process exchanges profile titles that end with 'stig-baseline', 'cis-baseline', 'srg-baseline' to Security Technical Implementation Guide and capitalizes all preceding words removing the dashes.

Without applying title formatting:
image

After applying title formatting:
image

### DIFF
--- a/apps/frontend/src/components/global/ExportCKLModal.vue
+++ b/apps/frontend/src/components/global/ExportCKLModal.vue
@@ -258,7 +258,6 @@
         </v-btn>
       </v-card-actions>
     </v-card>
-    <!-- </v-container> -->
   </v-dialog>
 </template>
 
@@ -313,7 +312,7 @@ export default class ExportCKLModal extends Vue {
   showingModal = false;
   formatProfileTitle = false;
   updatingProfileTitle = false;
-  originalProfileTitle = new Map<Number, string>();
+  originalProfileTitle = new Map<number, string>();
   roles = Object.values(Role);
   types = Object.values(Assettype);
   techareas = Object.values(Techarea);
--- a/apps/frontend/src/components/global/ExportCKLModal.vue
+++ b/apps/frontend/src/components/global/ExportCKLModal.vue
@@ -19,6 +19,8 @@
         <v-col cols="2">
           <v-checkbox
             v-model="formatProfileTitle"
+            v-b-tooltip.hover
+            title="Attempts to format the profile title into a proper CKL title name"
             class="mx-2"
             label="Format Profile Title"
           />
@@ -559,6 +561,7 @@ export default class ExportCKLModal extends Vue {
 
     // Find if we need to format the name
     let index = 0;
+    // Only format for UCs where the name ends with values contained in the baselineArray
     const baselineArray = ['stig-baseline', 'cis-baseline', 'srg-baseline'];
     for (const baseline of baselineArray) {
       if (name.indexOf(baseline) > 0) {
--- a/apps/frontend/src/components/global/ExportCKLModal.vue
+++ b/apps/frontend/src/components/global/ExportCKLModal.vue
@@ -12,7 +12,9 @@
     <v-card>
       <v-row no-gutters>
         <v-col cols="10">
-          <v-card-title class="headline"> Export as DISA Checklist </v-card-title>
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -526,6 +555,52 @@ export default class ExportCKLModal extends Vue {
...
   }
 
+  setProperName(name: string, fileIndex: number, profileIndex: number): string {
+    let newName = name;
+
+    // Find if we need to format the name
+    let index = 0;
+    const baselineArray = ['stig-baseline', 'cis-baseline', 'srg-baseline'];
+    for (const baseline of baselineArray) {
+      if (name.indexOf(baseline) > 0) {
```
DMedina6, line 565: I know this works for every relevant use case, but is this to test whether a string in the baselineArray is present in the name then should it not be '> -1'? Otherwise this excludes names where a baselineArray value is at the beginning of the string.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -284,6 +310,9 @@ export default class ExportCKLModal extends Vue {
   @Prop({type: Object, required: true}) readonly filter!: Filter;
 
   showingModal = false;
+  formatProfileTitle = false;
+  updatingProfileTitle = false;
```
DMedina6, line 314: I'm not sure what this does. I see it gets set to true when the name needs to be reformatted, but it doesn't get used at any point. Does this get used somewhere outside of these files?

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -526,6 +555,52 @@ export default class ExportCKLModal extends Vue {
...
+        index = name.indexOf(baseline);
+        break;
+      }
+    }
+
+    // We need to format the name
+    if (index > 0) {
+      this.updatingProfileTitle = true;
+      const originalTitleIndex = fileIndex + profileIndex;
+      // Preserver the old name
```
DMedina6, line 575: Typo

apps/frontend/src/components/global/ExportCKLModal.vue
DMedina6, line 565: I know this works for every relevant use case, but is this to test whether a string in the baselineArray is present in the name? Then should it not be '> -1'? Otherwise this excludes names where a baselineArray value is at the beginning of the string.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -526,6 +555,52 @@ export default class ExportCKLModal extends Vue {
...
+      if (!this.originalProfileTitle.has(originalTitleIndex)) {
+        this.originalProfileTitle.set(originalTitleIndex, name);
+      }
+      // Get the name value up to the index, replace dashes with spaces
+      newName = name.substring(0, index).split('-').join(' ');
+      // Convert the first letter of each word into uppercase
+      newName = newName.replace(/(?:^\w|[A-Z]|\b\w)/g, function (word) {
```
DMedina6, line 582: Why replace an uppercase letter in this regular expression (|A-Z|)?

DMedina6 (approved):
Code works as expected after testing, and tests all pass.  I'm going to approve this even though I left other comments for review.

apps/frontend/src/components/global/ExportCKLModal.vue
georgedias: You're correct it will exclude if the values are fount at the beginning of the name, which we should ignore, because a valid name was not provided.

apps/frontend/src/components/global/ExportCKLModal.vue
georgedias: The flag updatingProfileTitle is no longer needed, it was used on a previous logic, will remove.

PR #5889, opened 2024-06-05
Add Unicode Support for Exports

### LINKED ISSUE #5872: [BUG]
**Describe the bug**
Hi We love this report system, the only seggestion is exportHTML & exportCSV have not  correct language support,  could you handle this?  or had better a choice language option.

**Expected behavior**
I'm so happy that the CAAT spreadsheet can export other language correctly.

exportHTML_ERROR

charleshu-8:
- Add Unicode  support to HTML export
- Add Unicode  support to CSV export
- Check other exports for Unicode  support


Amndeep7 (changes requested):
Let's add a test for these exports and utf-8 characters

charleshu-8:
Probably want to wait on #5910 and move the injected BOM to OHDF Converters side of things.

Issue #5634, opened 2024-03-14
CKL metadata should be have input validation and consequently disable functionality or throw errors


Amndeep7:
Definition of done:

HDF Converters section:
- [x] The underlying mapper should validate the checklist metadata that it receives and throw an error if it fails there.  If you write the validation function well enough there, you might be able to import it and use it in Heimdall as well.

Heimdall section:
There is no validation for the checklist export modal despite us knowing the types and format of the data that should be in each field.  We should do user input validation before allowing the user to export a checklist file that might not be valid and therefore fail when someone attempts to consume it.
- [x] Each field has validation on it to at minimum require that it is of the correct typescript type (ex. number fields should be numbers)
- [x] Each field has validation on it to ensure that it matches the expected format of that field, ex. Stigviewer should be able to open each file, and it should also make sense that something looks formatted like that.
- [x] The export button should not be clickable unless all of the validation passes.  This will probably require doing some rearchitecting of the modal to make it possible.  Maybe check out Vuetify's form validation: https://v2.vuetifyjs.com/en/components/forms/

![image](https://github.com/mitre/heimdall2/assets/2015821/2f82e308-a776-4a7a-a33d-da13f8cec331)

SAF CLI section:
- [x] The SAF CLI generates CKL metadata - ensure that what the user provides meets the validation
- [x] The SAF CLI ckl mapper integration takes in CKL metadata - ensure that the error that will propagate up from the mapper is outputted in a way that will make sense to the end user

PR #5902, opened 2024-06-07
checklist metadata input validation on export

kemley76:
Fix to #5634. Updated the checklist export modal to validate IP address, MAC address, and FQDN before enabling export button.


libs/hdf-converters/package.json
```diff
@@ -62,6 +64,7 @@
     "@types/jest": "^27.0.0",
     "@types/lodash": "^4.14.161",
     "@types/node": "^20.1.0",
+    "@types/validator": "^13.11.10",
```
Amndeep7, line 67: After discussion, it seemed like these needed to be in dependencies, so let's get rid of the double import that's happening here.

libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
```diff
@@ -230,6 +231,31 @@ export function getChecklistObjectFromHdf(
   return _.get(hdf, 'passthrough.checklist', EmptyChecklistObject);
 }
 
+export function validateChecklistMetadata(
+  metadata: Pick<Asset, 'hostip' | 'hostfqdn' | 'hostmac'>
+): {isError: boolean; message: string; invalid: string[]} {
```
Amndeep7, line 236: Let's explore using an actual Results type monoid as opposed to this unless using that type turns out to be a pain for some reason

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -540,20 +584,15 @@ export default class ExportCKLModal extends Vue {
...
+  validateFormat(field: typeof ValidationProperties, name: string): string[] {
+    const errors: string[] = [];
+    const dirty = _.get(field, '$dirty');
+    const invalid = _.get(field, '$invalid');
+    if (!dirty) {
+      return [];
     }
-    const macPattern = /^([0-9A-Fa-f]{2}[:-]){5}[0-9A-Fa-f]{2}$/;
-    return macPattern.test(value) || 'Invalid MAC Address Format';
+    invalid && errors.push(`Invalid ${name} Format`);
```
Amndeep7, line 594: While this works, javascript is a language that's mostly built on the concept of statements, not expressions that do side effects like this.

I would rewrite this into the more typical if statement structure (`if (invalid) { errors.push }`) or make the return statement a ternary (`return invalid ? errors + ['invalid format'] : errors`).

Amndeep7 (changes requested):

libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
kemley76: I created a simple Result<T, E> type for this purpose.

apps/frontend/src/components/global/ExportCKLModal.vue
kemley76: This has been refactored away. No longer an issue

libs/hdf-converters/package.json
kemley76: Done.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -303,9 +327,29 @@ type FileData = {
...
+  return (_, file: ExtendedEvaluationFile) =>
+    !validateChecklistMetadata(file).invalid.includes(prop);
+}
+
 @Component({
-  components: {
-    LinkItem
+  mixins: [validationMixin],
+  components: {LinkItem},
+  validations: {
```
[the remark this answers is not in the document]
kemley76: Moved to separate issue

libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
```diff
@@ -230,6 +231,31 @@ export function getChecklistObjectFromHdf(
   return _.get(hdf, 'passthrough.checklist', EmptyChecklistObject);
 }
 
+export function validateChecklistMetadata(
```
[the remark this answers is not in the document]
kemley76: Done. All enum fields are now validated along with each profile's version, release number, and date.

libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
```diff
@@ -278,15 +304,23 @@ export class ChecklistResults extends ChecklistJsonixConverter {
   constructor(data: string | ExecJSON.Execution, withRaw = false) {
     super(jsonixMapping);
     this.data = data;
+    const throwIfInvalid = (metadata: Asset) => {
```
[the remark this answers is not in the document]
kemley76: Done.

libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
```diff
@@ -231,29 +234,88 @@ export function getChecklistObjectFromHdf(
...
+    releasedate: {
+      type: 'string',
+      conform: (date: string) => !date || isDate(date),
+      message: 'Release date must be a valid date'
+    }
   }
-  return {isError: errors.length > 0, message: errors.join('\n'), invalid};
+};
+
+export function validateChecklistMetadata(
```
Amndeep7, line 297: after discussion, we are going to move forward with having a helper function exposed in the api to do the checklist metadata validation.  this function and the contructors will utilize our internal functions for validating the asset information and the profile information.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -585,14 +589,12 @@ export default class ExportCKLModal extends Vue {
   }
 
   validateFormat(field: typeof ValidationProperties, name: string): string[] {
-    const errors: string[] = [];
     const dirty = _.get(field, '$dirty');
     const invalid = _.get(field, '$invalid');
-    if (!dirty) {
+    if (!dirty || !invalid) {
```
Amndeep7, line 595: After discussion, we have no idea what dirty means.  Please research.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -718,11 +720,34 @@ export default class ExportCKLModal extends Vue {
...
+          const result = validateProfileMetadata(profile);
+          if (!result.ok) {
+            isInvalid = true;
+            errorMessages.push(result.error.message);
+          }
+        }
+
+        // display error message upon any invalid inputs
+        if (isInvalid) {
+          SnackbarModule.failure(errorMessages.join('\n'));
```
Amndeep7: Can we do some more research to see if it's possible to restrict it to just integers natively?  Otherwise, I guess this is fine.

libs/hdf-converters/src/utils/result.ts
```diff
@@ -0,0 +1 @@
+export type Result<T, E = Error> = {ok: true; value: T} | {ok: false; error: E};
```
Amndeep7, line 1: Sweet and simple.  I like it.

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportCKLModal.vue
kemley76: The use of `dirty` is unclear and unnecessary. It will be removed. [https://vuelidate.js.org/#sub-v-values](https://vuelidate.js.org/#sub-v-values)

libs/hdf-converters/src/ckl-mapper/checklist-mapper.ts
kemley76: Done. Validation functions are both still there but can be called both at once from a new single checklist metadata validation method. This makes the logic on the frontend much simpler.

apps/frontend/src/components/global/ExportCKLModal.vue
kemley76: Done.

apps/frontend/src/components/global/ExportCKLModal.vue
kemley76: it is now restricted by cancelling keyboard events if it's not a number character. It is still possible to get invalid keys in there, but the user would have to really try for that to happen. It would still fail upon export and give an appropriate snackbar error message if they did manage it.

apps/frontend/src/components/global/ExportCKLModal.vue
Amndeep7, line 1: When the error message shows up, it hides the expected format.  Either the hint should stay up or the error message should also include a hint.

Amndeep7, line 1: Vul ID Mapping only ever shows the info icon, not a dropdown even tho there are two possible options.

libs/hdf-converters/package.json
Amndeep7, line 1: Add tests to confirm that the hdf converter validation functions actually throw errors, both hdf->ckl and ckl->hdf.

libs/hdf-converters/src/ckl-mapper/checklist-jsonix-converter.ts
```diff
@@ -22,6 +22,7 @@ import {
   Vulnattribute
 } from './checklistJsonix';
 import {coerce} from 'semver';
+import {validateChecklistProfileMetadata} from './checklist-mapper';
```
Amndeep7, line 25: Do not make the imports circular.  Either move everything here (if it makes sense) or put it into its own file.

apps/frontend/src/components/global/ExportCKLModal.vue
```diff
@@ -182,13 +196,15 @@
                         type="number"
                         :placeholder="profile.versionplaceholder"
                         class="pr-2"
+                        @keydown="(event) => preventNonNumeric(event)"
```
Amndeep7, line 199: Does not seem to handle all cases, and the behavior gets kinda weird with invalid stuff maybe turning into a 0 or something.  Better to just go back to the snackbar warning.

Amndeep7 (changes requested):

apps/frontend/src/components/global/ExportCKLModal.vue
kemley76: done

kemley76: done

apps/frontend/src/components/global/ExportCKLModal.vue
kemley76: Done

libs/hdf-converters/src/ckl-mapper/checklist-jsonix-converter.ts
kemley76: Done

libs/hdf-converters/package.json
kemley76: Done

Amndeep7 (approved):

PR #5910, opened 2024-06-11
Update hdf2html mapper

kemley76:
This PR is for the necessary changes to complete [this SAF PR](https://github.com/mitre/saf/pull/1689). hdf-converters now will be read the template files directly rather than using Axios to request them. The package.json and tailwind were configured to load the files into the hdf-converter's lib directory.


kemley76:
A completely different approach has been used that does not branch from any of these changes. New PR is #6004

PR #5952, opened 2024-07-01
GoSec Mapper `impact` Fix

charleshu-8:
- Fix `impact` to refer to GoSec's `severity` field.
- Other minor adjustments to mapper.

### DIFF
--- a/libs/hdf-converters/sample_jsons/gosec_mapper/gosec-hdf-withraw.json
+++ b/libs/hdf-converters/sample_jsons/gosec_mapper/gosec-hdf-withraw.json
@@ -129,6 +129,130 @@
           "Golang errors": {}
         }
       }
-    ]
+    ],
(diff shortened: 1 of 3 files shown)


charleshu-8:
Code smells are commented out lines for printing test files. Disregard.

Amndeep7 (changes requested):
I manually changed one of the results in the sample file to have LOW severity.  This caused the entire G304 section to be changed to LOW severity even though the other ones were at MEDIUM severity.  We need to determine if it is necessary to actually restructure this mapper so that the results are not consolidated by 'id' since even though metadata like severity is probably not gonna change due to being tied to the CWE, other metadata like confidence probably could change per result.  I think we should probably generate more sample files since the provided sample is not sufficient.

Amndeep7 (approved):

PR #5982, opened 2024-07-09
gosec Mapper Rework

charleshu-8:
- Rework gosec mapper to consider new discoveries on gosec results schema.
- Other QoL fixes.


charleshu-8:
Code smells are for intentionally commented out code to print test results. Safe to disregard.

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -44,8 +62,8 @@ export class GoSecMapper extends BaseConverter {
     statistics: {},
     profiles: [
       {
-        name: 'Gosec scanner',
-        title: 'gosec',
+        name: 'gosec Scan',
```
Amndeep7, line 65: Why the name change?

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -44,8 +62,8 @@ export class GoSecMapper extends BaseConverter {
...
+        title: 'gosec Scan',
```
Amndeep7, line 66: Why the title change?

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -92,9 +108,10 @@ export class GoSecMapper extends BaseConverter {
         return {
           auxiliary_data: [
             {
-              name: 'Gosec',
+              name: 'gosec',
```
Amndeep7, line 111: Why the name change?  And why does it not match the other changes?

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -22,11 +22,29 @@ function nistTag(input: Record<string, unknown>): string[] {
...
+  return `${_.get(input, 'nosec')}` === 'false' &&
+    `${_.get(input, 'suppressions')}` === 'null'
+    ? ExecJSON.ControlResultStatus.Failed
+    : ExecJSON.ControlResultStatus.Skipped;
+}
+
+// If a gosec rule violation is suppressed, forward the given justification
+function formatSkipMessage(input: Record<string, unknown>): string {
+  return `${_.get(input, 'suppressions')}` !== 'null'
+    ? `${_.get(input, 'suppressions[0].justification')}`
```
Amndeep7, line 36: is it only an array of size 1?  should we just combine all the suppressions texts into a single string?

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -22,11 +22,29 @@ function nistTag(input: Record<string, unknown>): string[] {
...
+    : '';
```
Amndeep7, line 37: should we provide some string explicitly stating that a reason wasn't provided?

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -22,11 +22,29 @@ function nistTag(input: Record<string, unknown>): string[] {
...
+// Check `nosec` and `suppressions` fields which denote whether the gosec rule violation should be suppressed/skipped
+function formatStatus(input: Record<string, unknown>): string {
+  return `${_.get(input, 'nosec')}` === 'false' &&
+    `${_.get(input, 'suppressions')}` === 'null'
+    ? ExecJSON.ControlResultStatus.Failed
+    : ExecJSON.ControlResultStatus.Skipped;
+}
+
+// If a gosec rule violation is suppressed, forward the given justification
+function formatSkipMessage(input: Record<string, unknown>): string {
```
Amndeep7, line 34: should we also provide the .kind attribute as well to say where the suppression came from?

Amndeep7 (changes requested):

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -47,7 +47,7 @@ function formatMessage(input: Record<string, unknown>): string {
   return `${_.get(input, 'confidence')} confidence of rule violation at:\n${_.get(input, 'code')}`;
 }
 
-export class GoSecMapper extends BaseConverter {
+export class GosecMapper extends BaseConverter {
```
Amndeep7, line 50: This might not be a change we can do without breaking a downstream project's stuff.

Amndeep7 (changes requested):

libs/hdf-converters/src/gosec-mapper.ts
Amndeep7: I've reached out to @jweissm to see if they can provide any input.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: The tool is stylized as 'gosec'.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: See above.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: See above for name change. Passthrough `name` fields also typically only involve just the tool name.

libs/hdf-converters/src/gosec-mapper.ts
```diff
@@ -22,11 +22,29 @@ function nistTag(input: Record<string, unknown>): string[] {
   return CWE_NIST_MAPPING.nistFilter(cwe, DEFAULT_NIST_TAG);
 }
 
+// Check `nosec` and `suppressions` fields which denote whether the gosec rule violation should be suppressed/skipped
+function formatStatus(input: Record<string, unknown>): string {
+  return `${_.get(input, 'nosec')}` === 'false' &&
+    `${_.get(input, 'suppressions')}` === 'null'
```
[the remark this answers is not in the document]
charleshu-8: Suppressions that are recorded in the `Issue` element only apply that that specific issue. If a suppression if applied for problem Y, only instances of problem Y will note that suppression; problem X instances will ignore suppressions of problem Y.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: Something like `N/A`?

Amndeep7: nah i would say something like "User didn't supply a reason" or something like that.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: Addressed.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: Addressed.

libs/hdf-converters/src/gosec-mapper.ts
charleshu-8: Addressed.

libs/hdf-converters/src/gosec-mapper.ts
Amndeep7: They didn't respond.  We move forward.

Amndeep7 (approved):

Amndeep7 (changes requested):
Providing both a message and a skip message doesn't work in Heimdall.  The message always takes priority: https://github.com/mitre/heimdall2/blob/0448daab17b98c23e70053617151e604c3cf0a48/apps/frontend/src/components/cards/controltable/ControlRowCol.vue#L79.  We need to either modify this mapper to only supply one or the other OR modify the frontend code to show both.  If we do the latter, then we should validate that this doesn't have any unintended side-effects with some of the other mappers.  I think the decision for which we do falls on @aaronlippold and @ejaronne.

Amndeep7 (approved):

PR #5994, opened 2024-07-11
Checklist Comment Fix

charleshu-8:
- Move CKL comment from control level to results level.


Amndeep7:
"On additional analysis, we realized that the checklist's comment field spans the context of the entire control, not the individual tests that may be run within the control. Each of those test outcomes are recorded in separate controls.results objects. The comments field content should be placed outside of the results objects as is done currently by using a labeled desc field. Recommend not trying to place comment into results objects. Please close this PR without merging. Thanks! Eugene."

PR #6007, opened 2024-07-18
Convert Microsoft Secure Score to OHDF

meme112233:
This PR introduces the converter for Microsoft SecureScore to OHDF.

Mapping is performed using the output of two Microsoft Graph API endpoints:
* https://graph.microsoft.com/v1.0/security/secureScores
  * This contains the report of the security assessment
* https://graph.microsoft.com/v1.0/security/secureScoreControlProfiles 
  * This contains the details of (many, but not all) the controls evaluated as part of the security assessment

The mapper takes the output of the two endpoints as a single json document.  Each document is included as an object under the keys `profiles` and `secureScore`.

```
{
   "secureScore":  <The full text of the secureScore endpoint output as JSON object>,
   "profiles":  <The full text of the secureScoreControlProfiles endpoint output as JSON object>
}
```
This combined file can be built with `jq`
```
jq -s '{"secureScore": .[0], "profiles": .[1]}' secureScore.json profiles.json
```

Sample combined secureScore/profiles document for testing:
[combined_msft.json](https://github.com/user-attachments/files/16354855/combined_msft.json)

# Mapper design decisions of interest
* impact: Derived from the maxScore of the Profile matching the controlName
  * When no matching Profile is found, a default impact of 0.5 is used
* status: 
  * success:
    * scoreInPercentage === 100
    * score === max value from `maxScore` field of any matching Profile
  * error:
    * score field is undefined (per API, should never happen)
  * fail:
    * any other condition
* tags
  * groups
    * controlCategory field from any matching Profiles
  * threats
    * threat field from any matching Profiles

A matching PR for the SAF-CLI is expected early next week that will support:
* CLI conversion of combined secureScore endpoints to OHDF.
  * Starting with pre-downloaded REST endpoint outputs
  * Starting from Graph API credentials


libs/hdf-converters/src/msft-config-mapper.ts
```diff
@@ -0,0 +1,131 @@
...
+                    .filter((p) => p.id === d.controlName)
+                    .map((p) => p.threats);
+                }
+              }
+            },
+            source_location: {},
+            results: [
+              {
+                status: {
+                  transformer: (d: ControlScore | any) => {
```
meme112233, line 97: Any Typescript gurus, Help needed to allow access to field that is often/always present in the data, but not in the Typescript defs provided by Microsoft.

charleshu-8: Hey, could you clarify which field you are talking about specifically?

ejaronne:
1. Profiles.value.title: recommend for controls.title
2. Perhaps the Impact should be based on Profiles.value.userImpact instead of Profiles.value.maxScore? I can't yet find the Microsoft documentation that explains the difference between these. Impact is intended to be the relative danger should this control be in a failed state.
3. Remediation: seems like a perfect opportunity as descriptions:[data:”<remediation>”, label:”fix”]
4. RemediationImpact: seems like a perfect opportunity as descriptions:[data:”<remediation>”, label:”rationale”]

libs/hdf-converters/src/msft-config-mapper.ts
```diff
@@ -0,0 +1,131 @@
...
+  ControlScore,
+  SecureScoreControlProfile
+} from '@microsoft/microsoft-graph-types';
+import {ExecJSON} from 'inspecjs';
+
+import {version as HeimdallToolsVersion} from '../package.json';
+
+import {BaseConverter, ILookupPath, MappedTransform} from './base-converter';
+
+export class MsftConfigMapper extends BaseConverter {
```
Amndeep7, line 12: strange that the mapper is called 'msgftconfigmapper' whereas the other things like the fingerprinting enum and the profile.name refer to it as 'microsoft secure score'.  probably better to consolidate on the one.

Amndeep7 (changes requested):

libs/hdf-converters/src/msft-config-mapper.ts
meme112233: `d. scoreInPercentage` exists in the returned JSON but not the typescript type def

charleshu-8:
```
if (d.scoreInPercentage === 100) {
```
It looks like you're already successfully grabbing it here, or am I looking at the wrong place?

libs/hdf-converters/package.json
```diff
@@ -59,7 +62,7 @@
     "yaml": "^2.1.1"
   },
   "devDependencies": {
-    "@types/jest": "^27.0.0",
+    "@types/jest": "^29.0.0",
```
charleshu-8, line 65: I believe https://github.com/mitre/heimdall2/pull/5700 details some issues occurring w/ Jest v29. Should play it safe and downgrade this back to v27.

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+            'sample_jsons/msft_secure_score_mapper/profiles.json',
+            {
+              encoding: 'utf-8'
+            }
+          )
+        )
+      })
+    );
+
+    /* tslint:disable */
```
charleshu-8, line 28: Remove this TSLint flag. I can handle the SonarCloud flag for this code if that's what prompted you to do this.

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    // fs.writeFileSync(
+    //   'sample_jsons/msft_secure_score_mapper/burpsuite-hdf.json',
```
charleshu-8, line 30: Please rename this to something that is descriptive of the file you generated (e.g., `msft-secure-score-hdf.json`).

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    //   JSON.stringify(mapper.toHdf(), null, 2)
+    // );
+    /* tslint:enable */
+
+    expect(omitVersions(mapper.toHdf())).toEqual(
+      omitVersions(
+        JSON.parse(
+          fs.readFileSync(
+            'sample_jsons/msft_secure_score_mapper/burpsuite-hdf.json',
```
charleshu-8, line 39: > Please rename this to something that is descriptive of the file you generated (e.g., msft-secure-score-hdf.json).

Ditto.

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+import {MsftConfigMapper} from '../../../src/msft-config-mapper';
+import {omitVersions} from '../../utils';
+
+describe('msft_secure_score_mapper', () => {
+  it('Successfully converts Microsoft Secure Score reports', () => {
+    const mapper = new MsftConfigMapper(
+      JSON.stringify({
+        secureScore: JSON.parse(
+          fs.readFileSync(
+            'sample_jsons/msft_secure_score_mapper/secureScore.json',
```
charleshu-8, line 11: Place all files generated from Secure Score under a sub-directory named `sample_input_report`. See other sample files for example.

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+            {
+              encoding: 'utf-8'
+            }
+          )
+        ),
+        profiles: JSON.parse(
+          fs.readFileSync(
+            'sample_jsons/msft_secure_score_mapper/profiles.json',
```
charleshu-8, line 19: > Place all files generated from Secure Score under a sub-directory named sample_input_report. See other sample files for example.

Ditto.

charleshu-8 (commented):
Part 1 of review, will look at mapper proper later.

libs/hdf-converters/src/msft-config-mapper.ts
meme112233: This was just to address the 'code smell' part of the CI/CD pipeline

libs/hdf-converters/src/msft-config-mapper.ts
meme112233: rename completed

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,202 @@
...
+        ],
+        sha256: ''
+      }
+    ],
+    passthrough: {
+      profiles: {
+        value: {
+          tramsformer: () => this.profiles
+        }
+      },
```
meme112233, line 198: @charleshu-8, @Amndeep7 , 
Thoughts on passing `this.profiles` through?  

* ```Property 'profiles' is used before its initialization.ts(2729)```
* Also, transformers don't seem to get executed within `passthrough`

charleshu-8: Pushed up a manual fix of passthrough to give you an idea of what it looks like in the other mappers.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,202 @@
...
+    statistics: {},
+    profiles: [
+      {
+        name: 'Microsoft Secure Score',
+        version: {path: 'v4'},
+        title: {
+          transformer: (d: SecureScore) =>
+            `Azure Secure Score report: TenantID: ${d.azureTenantId}`
+        },
+        summary: 'NAME',
```
charleshu-8, line 36: Change this to be something meaningful or just leave blank if you don't have anything to fill it with.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,202 @@
...
+                if (titles.length > 0) {
+                  return titles.join('... ');
+                } else {
+                  return `${d.controlCategory || ''}:${d.controlName || ''}`;
+                }
+              }
+            },
+            desc: {
+              transformer: (d: ControlScore) => d.description || ''
+            },
```
charleshu-8, line 63:
```suggestion
            desc: {path: 'description'},
```
Same behavior.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,202 @@
...
+            source_location: {},
+            code: {
+              transformer: (
+                d: ControlScore & {implementationStatus: string}
+              ) => {
+                const profiles = this.getProfiles(d.controlName || '');
+
+                if (profiles?.length === 0) {
+                  return d.implementationStatus;
```
charleshu-8, line 132: We're returning this if there are no profiles matching the control name? What if there are?

meme112233: just pushed a better implementation.

always include implementationStatus from control.  include from profiles if available.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,202 @@
...
+      release: HeimdallToolsVersion
+    },
+    version: HeimdallToolsVersion,
+    statistics: {},
+    profiles: [
+      {
+        name: 'Microsoft Secure Score',
+        version: {path: 'v4'},
+        title: {
+          transformer: (d: SecureScore) =>
```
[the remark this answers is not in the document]
meme112233: all updated

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,202 @@
...
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion
+    },
+    version: HeimdallToolsVersion,
+    statistics: {},
+    profiles: [
+      {
+        name: 'Microsoft Secure Score',
+        version: {path: 'v4'},
```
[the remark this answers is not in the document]
meme112233: Removed as no version provided by Microsoft and field is optional

ejaronne (approved):
Looks good to me!  Thanks!

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+          ...(this.withRaw && {raw: data})
+        };
+      }
+    }
+  };
+  constructor(secureScore_and_profiles_combined: string, withRaw = false) {
+    const rawParams = JSON.parse(secureScore_and_profiles_combined);
+    super(rawParams.secureScore.value[0], true);
+    this.withRaw = withRaw;
+    this.profiles = rawParams.profiles.value as SecureScoreControlProfile[];
```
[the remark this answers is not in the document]
Amndeep7: same question for the securescore one as well

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+      release: HeimdallToolsVersion
+    },
+    version: HeimdallToolsVersion,
+    statistics: {},
+    profiles: [
+      {
+        name: 'Microsoft Secure Score Scan',
+        title: {
+          transformer: (data: SecureScore) =>
+            `Azure Secure Score report - TenantID: ${data.azureTenantId}`
```
Amndeep7, line 33: what's the difference between the azuretenantid and the id?

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                `${data.controlCategory}:${data.controlName}`
+            },
+            title: {
+              transformer: (data: ControlScore) => {
+                const titles = this.getProfiles(data.controlName || '')
+                  .filter((profile) => profile.title !== undefined)
+                  .map((profile) => profile.title);
+
+                if (titles.length > 0) {
+                  return titles.join('... ');
```
Amndeep7, line 53: worth doublechecking, but i think control titles can be multiline so we could maybe just use a newline here

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                } else {
+                  return `${data.controlCategory || ''}:${data.controlName || ''}`;
```
Amndeep7, line 55: if a controlcategory is not supplied, it'd be weird to have a colon just hangin out there as a prefix lol

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+        controls: [
+          {
+            path: 'controlScores',
+            id: {
+              transformer: (data: ControlScore) =>
+                `${data.controlCategory}:${data.controlName}`
+            },
+            title: {
+              transformer: (data: ControlScore) => {
+                const titles = this.getProfiles(data.controlName || '')
```
Amndeep7, line 48: optional: one thing we could do to not have to repeatedly call getProfiles is make it so that controls uses a transformer that would act as a closure.  probably would look like the following:

```
blah

makeControls(controlScores: ControlScore[]): MappedTransform<
  ExecJSON.Control[] & ILookupPath,
  ILookupPath
> {
const controlProfiles = this.getProfiles(data.controlName || '')
return the subsection of the mapping you've already written but now without needing to call getProfiles a bunch of times since now you can just use that variable
}

mapping = {
blah,
profiles: [
{
blah,
controls: {
path: 'controlScores',
transformer: makeControls or something like that
}
}
]
}
```

Haven't counted how many times we do the search.  Might not be necessary if we don't do it that often / there's not that much data to search through so there won't be a significant speed savings.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+
+                if (titles.length > 0) {
+                  return titles.join('... ');
+                } else {
+                  return `${data.controlCategory || ''}:${data.controlName || ''}`;
+                }
+              }
+            },
+            desc: {path: 'description'},
+            impact: {
```
Amndeep7, line 60: would profile.rank have an impact on impact?  https://github.com/microsoftgraph/msgraph-typescript-typings/blob/main/microsoft-graph.d.ts#L8357-L8358 this says that it's a stack ranking of the control but idk what that actually implies

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                if (knownMaxScores.length === 0) {
+                  return 0.5;
+                }
+
+                const highMaxScore = Math.max(...knownMaxScores);
+                return highMaxScore / 10.0;
+              }
+            },
+            refs: [],
+            tags: {
```
Amndeep7, line 76: image

when tags are empty, do not provide them.  i think you can just return undefined in the transformer but if that doesn't work then you can fall back on the boolean spread operator trick to not include the attribute in the object at all.  example: https://github.com/mitre/heimdall2/blob/master/libs/hdf-converters/src/xccdf-results-mapper.ts#L422-L429

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+              group: {
```
Amndeep7, line 77: why are we using a different naming scheme than the original format?  shouldn't we just call this 'controlcategory' or 'category'?

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                  return this.getProfiles(data.controlName || '').map(
+                    (profile) => profile.tier
+                  );
+                }
+              },
+              threats: {
+                transformer: (data: ControlScore) => {
+                  // return unique threats from the profile document where its id matches the controlName
+                  const uniqs: Set<string> = new Set();
+                  this.getProfiles(data.controlName || '').forEach(
```
Amndeep7, line 97: prefer for..of over foreach

https://2ality.com/2021/01/looping-over-arrays.html

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                  // return tiers from the profile document where its id matches the controlName
+                  return this.getProfiles(data.controlName || '').map(
+                    (profile) => profile.tier
+                  );
+                }
+              },
+              threats: {
+                transformer: (data: ControlScore) => {
+                  // return unique threats from the profile document where its id matches the controlName
+                  const uniqs: Set<string> = new Set();
```
Amndeep7, line 96: can probably just use https://lodash.com/docs#uniq instead of handrolling

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                  this.getProfiles(data.controlName || '').forEach(
+                    (profile: SecureScoreControlProfile) =>
+                      profile.threats?.forEach((threat) => uniqs.add(threat))
+                  );
+                  return [...uniqs];
+                }
+              },
+              services: {
+                transformer: (data: ControlScore) => {
+                  // return thrserviceeats from the profile document where its id matches the controlName
```
Amndeep7, line 106: typo

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                  // return userImpacts from the profile document where its id matches the controlName
+                  return this.getProfiles(data.controlName || '')
+                    .filter((profile) => profile.userImpact !== undefined)
+                    .map((profile) => profile.userImpact);
+                }
+              },
+              nist: ['SA-11', 'RA-5']
+            },
+            source_location: {},
+            descriptions: [
```
Amndeep7, line 123: did not address @ejaronne's request for a rationale subdescription based off of profile.remediationimpact

libs/hdf-converters/src/msft-secure-score-mapper.ts
Amndeep7, line 60: would profile.deprecated have an impact?  https://github.com/microsoftgraph/msgraph-typescript-typings/blob/main/microsoft-graph.d.ts#L8349-L8350 I am unsure if it being deprecated means that we should set the impact to 0 or if we leave it be.  this'd be a @ejaronne question.

Amndeep7, line 60: the types also show a userImpact field which is something that @ejaronne mentioned in his review.  there isn't a comment associated with it in the types, but it seems to have been provided in many of the profiles.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,203 @@
...
+                },
+                start_time: {transformer: () => this.data.createdDateTime}
+              }
+            ]
+          }
+        ],
+        sha256: ''
+      }
+    ],
+    passthrough: {
```
Amndeep7, line 186: a key thing that we try to do during the mapping process is ensure that all data is brought through into the hdf file, even if it just gets dumped into passthrough.

we are missing significant sections of the output at the moment.  i've already mentioned the URLs.  we also have secureScore.averageComparitiveScores, securescore.enabledServices, securescore.vendorInformation, etc.  we still have a couple fields in controlscore that are unmapped such as lastsynced.  doing the wholesale dump of profiles is fine.

Amndeep7 (changes requested):

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
```diff
@@ -0,0 +1,90 @@
+import fs from 'fs';
+import {MsftSecureScoreMapper} from '../../../src/msft-secure-score-mapper';
+import {omitVersions} from '../../utils';
+
+describe('msft_secure_score_mapper', () => {
+  it('Successfully converts Microsoft Secure Score reports', () => {
+    const mapper = new MsftSecureScoreMapper(
+      JSON.stringify({
```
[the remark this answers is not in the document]
JH-Klevin: This seems like an extra spot for inconsistency.

libs/hdf-converters/index.ts
```diff
@@ -40,3 +40,4 @@ export * from './src/utils/fingerprinting';
 export * from './src/veracode-mapper';
 export * from './src/xccdf-results-mapper';
 export * from './src/zap-mapper';
+export * from './src/msft-secure-score-mapper';
```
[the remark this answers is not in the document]
JH-Klevin: this has been updated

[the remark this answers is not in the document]
JH-Klevin: README entry added.

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -175,7 +175,8 @@ export class MsftSecureScoreMapper extends BaseConverter {
                     data: ControlScore & {implementationStatus: string}
                   ) => data.implementationStatus
                 },
-                start_time: {transformer: () => this.data.createdDateTime}
+                start_time: {transformer: () => this.data.createdDateTime},
+                run_time: 0
```
Amndeep7, line 179: Why is it necessary to have here explicitly?  Can't you just assume a runtime duration of 0 if this field is absent?

Amndeep7 (changes requested):

libs/hdf-converters/src/msft-secure-score-mapper.ts
JH-Klevin: agreed.  the current repeated lookup approach is based off of recommendations from initial design review.  as the data is small, postpone this change pending performance analysis.

libs/hdf-converters/test/mappers/forward/msft_secure_score_mapper.spec.ts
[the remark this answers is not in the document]
Amndeep7: I'm fine either way here

libs/hdf-converters/src/msft-secure-score-mapper.ts
JH-Klevin: azureTenantId is roughly the billing account for the Microsoft account.  id is the uuid for the secureScore report.  likely each generation of the report would have a different id.

meme112233: azureTenantId is roughly the billing account for the Microsoft account. id is the uuid for the secureScore report. likely each generation of the report would have a different id.

libs/hdf-converters/src/msft-secure-score-mapper.ts
meme112233: `profile.rank` is microsoft's recommended order in which failed controls should be addressed.  It takes into consideration:
*severity (points available)
* ease of implementation
* minimal impact to users
* complexity
https://learn.microsoft.com/en-us/defender-xdr/microsoft-secure-score-improvement-actions#ranking

Bsed on the number of factors that feed this number, I'd vote against its use as an impact indicator.

meme112233: `User impact is what the users will experience if the recommended action is enacted, and the Users affected are the people who will be impacted.`

https://learn.microsoft.com/en-us/defender-xdr/microsoft-secure-score-improvement-actions#assess-information-and-review-user-impact

meme112233: 1) profile.deprecated is always TRUE in the data that we've seen thus far.
2) `// Flag to indicate if a control is depreciated.`
     `deprecated?: NullableOption<boolean>;`
3) I suspect that if a control is deprecated, it would no longer be presented under the `LIST controlScore` endpoint, rendering this discussion moot.

This might be good to revisit after we first see a deprecated control.

libs/hdf-converters/src/msft-secure-score-mapper.ts
meme112233: updated to `category`

libs/hdf-converters/src/msft-secure-score-mapper.ts
meme112233: good change... implemented.

libs/hdf-converters/src/msft-secure-score-mapper.ts
meme112233: implemented

libs/hdf-converters/src/msft-secure-score-mapper.ts
Amndeep7: Makes sense.  Maybe the uuids for the securescore report should be concatenated and put into the profile name or something?

libs/hdf-converters/src/msft-secure-score-mapper.ts
Amndeep7: Makes sense.  The only factor we consider is severity (however it is defined).  Its complexity or ease of implementation is irrelevant for us.  This would be another thing that we should make sure to include in the tags though.

Amndeep7: Ok so irrelevant info for us, but still something that should be put in the tags.

libs/hdf-converters/src/msft-secure-score-mapper.ts
Amndeep7: Currently does not work, but we'll just fix heimdall to make it work.

libs/hdf-converters/src/msft-secure-score-mapper.ts
meme112233: added rank as a tag entry

Amndeep7: We'll worry about this resulting in false positives when someone complains I guess.

libs/hdf-converters/src/msft-secure-score-mapper.ts
Amndeep7: Will discuss with sponsor about necessity.

combined_msft.json
Amndeep7, line 1: please delete this file or move it to a more appropriate location

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,358 @@
...
+
+  constructor(combinedJson: string, withRaw = false) {
+    this.data = JSON.parse(combinedJson);
+    this.withRaw = withRaw;
+  }
+
+  toHdf(): ExecJSON.Execution[] {
+    const results: ExecJSON.Execution[] = [];
+
+    this.data.secureScore.value.forEach((element) => {
```
Amndeep7, line 41: prefer for..of in terms of loops, but here you could probably just chain maps

return this.data.securescore.value.map(e => whatever to get the new mapper).map(m => m.toHdf())

or even just

return this.data.securescore.value.map(e => get new mapper and then immediately call toHdf and return that)

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,358 @@
...
+                  ),
+                  (() => {
+                    const result = this.getProfiles(data.controlName || '')
+                      .map((profile) => profile.service)
+                      .filter((v) => Boolean(v));
+                    return result.length > 0;
+                  })()
+                ),
+                ...conditionallyProvideAttribute(
+                  'rank',
```
Amndeep7, line 210: rank is done twice but this latter one has a uniq on it for some reason

libs/hdf-converters/src/msft-secure-score-mapper.ts
```diff
@@ -0,0 +1,358 @@
...
+                  ),
+                  (() => {
+                    const result = this.getProfiles(data.controlName || '')
+                      .map((profile) => profile.userImpact)
+                      .filter((v) => Boolean(v));
+                    return result.length > 0;
+                  })()
+                )
+              }),
+              nist: ['SA-11', 'RA-5']
```
Amndeep7, line 238: instead of hardcoding this in, please use the constants defined here: https://github.com/mitre/heimdall2/blob/master/libs/hdf-converters/src/utils/global.ts#L12

Amndeep7 (changes requested):

libs/hdf-converters/src/msft-secure-score-mapper.ts
Amndeep7: Decided to go the middle road and just memoize that function instead.

Amndeep7 (approved):

PR #6013, opened 2024-07-22
Trufflehog mapper

andytang99:
Completed trufflehog mapper with sample jsons.


certs/mitrecerts.sh
charleshu-8, line 1: Revert this change. Shouldn't be updating this.

libs/hdf-converters/test/mappers/forward/trufflehog_mapper.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      fs.readFileSync(
+        'sample_jsons/trufflehog_mapper/sample_input_report/trufflehog.json',
+        {encoding: 'utf-8'}
+      )
+    );
+
+     fs.writeFileSync(
+       'sample_jsons/trufflehog_mapper/trufflehog-hdf.json',
+       JSON.stringify(mapper.toHdf(), null, 2)
+    );
```
charleshu-8, line 17:
```suggestion
     // fs.writeFileSync(
     //   'sample_jsons/trufflehog_mapper/trufflehog-hdf.json',
     //   JSON.stringify(mapper.toHdf(), null, 2)
     // );
```
Ensure that we aren't pushing up unexpected file writes to the repo.

libs/hdf-converters/test/mappers/forward/trufflehog_mapper.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+        'sample_jsons/trufflehog_mapper/sample_input_report/trufflehog.json',
+        {encoding: 'utf-8'}
+      ),
+      true
+    );
+
+     fs.writeFileSync(
+       'sample_jsons/trufflehog_mapper/trufflehog-hdf-withraw.json',
+       JSON.stringify(mapper.toHdf(), null, 2)
+     );
```
charleshu-8, line 47:
```suggestion
     // fs.writeFileSync(
     //   'sample_jsons/trufflehog_mapper/trufflehog-hdf-withraw.json',
     //   JSON.stringify(mapper.toHdf(), null, 2)
     // );
```
Same as above.

charleshu-8 (commented):
Quick note on stuff that stood out to me. Haven't looked at your mapper proper since it's WIP.

certs/mitrecerts.sh
andytang99: I don't think I changed anything here.

Amndeep7:
I'll resolve the sonarqube nonsense when the PR is ready to be merged.  There's one actual code smell where you didn't get of an unused import line, but the other code smells are the ones that we approve of with the commented out code in the testing area.  The remainder are quote unquote vulnerabilities because we have what are hopefully just sample/testing secrets in the test json files that it's getting triggered over though we know that they are (hopefully) actually just examples.

Amndeep7:
@andytang99 I enabled workflows on this pr.  Please make sure that all the workflows pass.  Currently you're failing the linting.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+      target_id: null  
+    },
+    version: HeimdallToolsVersion,
+    statistics: {
+      duration: null  
+    },
+    profiles: [
+      {
+        name: {path: 'SourceName'},              
+        title: {path: 'SourceName'},
```
charleshu-8, line 29:
```suggestion
        name: {path: 'wrapper[0].SourceName'},              
        title: {path: 'wrapper[0].SourceName'},           
```
Need to access the first element of the wrapper array field.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+    profiles: [
+      {
+        name: {path: 'SourceName'},              
+        title: {path: 'SourceName'},           
+        version: null,         
+        maintainer: null,      
+        summary: null,         
+        license: null,         
+        copyright: null,       
+        copyright_email: null,
```
charleshu-8, line 35:
```suggestion
```
Remove if applicable field is not available.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+        title: {path: 'SourceName'},           
+        version: null,         
+        maintainer: null,      
+        summary: null,         
+        license: null,         
+        copyright: null,       
+        copyright_email: null, 
+        supports: [],          
+        attributes: [],        
+        depends: [],
```
charleshu-8, line 38:
```suggestion
```
Remove if applicable field is not available.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+                nist: 'IA-5(7)',
+                cci: [
+                  'CCI-004069',
+                  'CCI-000202',
+                  'CCI-000203',
+                  'CCI-002367'
+                ],
+                severity: 'medium'
+            },             
+            descriptions: [],
```
charleshu-8, line 55:
```suggestion
```
Remove if applicable field is not available.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+                    return _.get(data, 'DetectorType') + '_' + _.get(data, 'DecoderName');
+                }
+            },               
+            desc: {
+                transformer: (data: Record<string, unknown>): string => {
+                    return 'Found ' + _.get(data, 'DetectorName') + ' secret using ' + _.get(data, 'DecoderName') + ' decoder'
+                }
+            },           
+            impact: 0.5,            
+            code: null,
```
charleshu-8, line 74:
```suggestion
```
Remove if applicable field is not available.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+                  transformer: (data: Record<string, unknown>): string => {
+                    return `${JSON.stringify(_.get(data, 'SourceMetadata'))}`;
+                  }
+                },                                
+                message: {
+                    transformer: (data: Record<string, unknown>): string => {
+                        return `${JSON.stringify(_.omit(data, 'SourceMetadata'))}`;
+                    }
+                },                                
+                run_time: null,
```
charleshu-8, line 88:
```suggestion
```
Remove if applicable field is not available.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+                status: ExecJSON.ControlResultStatus.Failed,  
+                code_desc: {
+                  transformer: (data: Record<string, unknown>): string => {
+                    return `${JSON.stringify(_.get(data, 'SourceMetadata'))}`;
+                  }
+                },                                
+                message: {
+                    transformer: (data: Record<string, unknown>): string => {
+                        return `${JSON.stringify(_.omit(data, 'SourceMetadata'))}`;
+                    }
```
charleshu-8, line 86: This is illegible in Heimdall. Consult @ejaronne for a fix.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,110 @@
...
+            ]
+          }
+        ],
+        sha256: ''
+      }
+    ],
+    passthrough: {
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
+        return {
+          auxiliary_data: [{name: '', data: _.omit([])}],  //Insert service name and mapped fields to be removed
```
charleshu-8, line 100: Store any unused fields here. If all fields are used, remove this section.

charleshu-8 (commented):
Review by me & @Rlin232

libs/hdf-converters/src/trufflehog-mapper.ts
charleshu-8: Remember to consult Heimdall while making these mappers to get a good idea of how our tools are showing them/how others see them downstream.

libs/hdf-converters/sample_jsons/trufflehog_mapper/trufflehog_to_hdf_example.json
charleshu-8, line 1: Delete this.

libs/hdf-converters/test/mappers/forward/trufflehog_mapper.spec.ts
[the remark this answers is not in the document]
charleshu-8: Any file in `sample_input_report` should get a test, so make one for ` trufflehog_saf_example.json`.

certs/mitrecerts.sh
Amndeep7, line 1: You changed the file permissions for this file.  Please revert this change.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+    passthrough: {
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
+        return {
+          ...(this.withRaw && {raw: data})
+        };
+      }
+    }
+  };
+  constructor(exportJson: string, withRaw = false) {
+    super({wrapper: JSON.parse(exportJson)}, true);
```
Amndeep7, line 94: what is the need for a wrapper object?  if the answer is to be able to do the array access for profile.name and other places, you can just do that directly without anything in front.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+              transformer: (data: Record<string, unknown>): string => {
+                return (
+                  _.get(data, 'DetectorName') + '_' + _.get(data, 'DecoderName')
+                );
+              }
+            },
+            id: {
+              transformer: (data: Record<string, unknown>): string => {
+                return (
+                  _.get(data, 'DetectorType') + '_' + _.get(data, 'DecoderName')
```
Amndeep7, line 48: The detector name and type are directly correlated.  What is the point of showing different values here.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+              nist: 'IA-5(7)',
+              cci: ['CCI-004069', 'CCI-000202', 'CCI-000203', 'CCI-002367'],
+              severity: 'medium'
+            },
+            refs: [],
+            source_location: {},
+            title: {
+              transformer: (data: Record<string, unknown>): string => {
+                return (
+                  _.get(data, 'DetectorName') + '_' + _.get(data, 'DecoderName')
```
Amndeep7, line 41: Why use underscores instead of just spaces?

libs/hdf-converters/src/trufflehog-mapper.ts
Amndeep7, line 1: Sponsor provided sample file fails on load

image

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+        supports: [],
+        attributes: [],
+        groups: [],
+        status: 'loaded',
+        controls: [
+          {
+            path: 'wrapper',
+            key: 'id',
+            tags: {
+              nist: 'IA-5(7)',
```
Amndeep7, line 32: this should probably be an array of size one, not a string

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+            id: {
+              transformer: (data: Record<string, unknown>): string => {
+                return (
+                  _.get(data, 'DetectorType') + '_' + _.get(data, 'DecoderName')
+                );
+              }
+            },
+            desc: {
+              transformer: (data: Record<string, unknown>): string => {
+                return (
```
Amndeep7, line 54: use template strings

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+  > = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion
+    },
+    version: HeimdallToolsVersion,
+    statistics: {},
+    profiles: [
+      {
+        name: {path: 'wrapper[0].SourceName'},
```
Amndeep7, line 21: i feel like sourceid should probably be here as well

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+            impact: 0.5,
+            results: [
+              {
+                status: ExecJSON.ControlResultStatus.Failed,
+                code_desc: {
+                  transformer: (data: Record<string, unknown>): string => {
+                    return `${JSON.stringify(_.get(data, 'SourceMetadata'), null, 2)}`;
+                  }
+                },
+                message: {
```
Amndeep7, line 72: i really don't like this approach of just shoving everything into the message or code_desc.  here's the result struct (https://github.com/trufflesecurity/trufflehog/blob/55fe05d0b48a35bf3de294adde2b8aaee09a3004/pkg/output/json.go#L27) - it seems pretty consistent so we can make decisions on what to do with the data.  it allows for spaces that can include arbitrary data but we can treat those places appropriately.  

ex. if we're using the source name etc as the profile info, then we should probably not repeat all that jazz to infinity in here.  similarly if the detector information is being used to describe the control, then it doesn't need to all be repeated in the results area.  then there's the attributes that are empty strings or nulls: those can probably be removed from the output - just make sure to not exclude on all falsy values since there might be times like the 'verified' attribute where a true/false is an acceptable value.  i think only the following info should be retained in this message area (https://github.com/trufflesecurity/trufflehog/blob/55fe05d0b48a35bf3de294adde2b8aaee09a3004/pkg/output/json.go#L42-L53) and only if it meets the non-default/empty criteria as specified above.

libs/hdf-converters/src/trufflehog-mapper.ts
```diff
@@ -0,0 +1,97 @@
...
+                  _.get(data, 'DecoderName') +
+                  ' decoder'
+                );
+              }
+            },
+            impact: 0.5,
+            results: [
+              {
+                status: ExecJSON.ControlResultStatus.Failed,
+                code_desc: {
```
Amndeep7, line 67: i guess code_desc can stay as it is though

Amndeep7 (changes requested):

libs/hdf-converters/index.ts
```diff
@@ -33,6 +33,7 @@ export * from './src/scoutsuite-mapper';
 export * from './src/snyk-mapper';
 export * from './src/sonarqube-mapper';
 export * from './src/splunk-mapper';
+export * from './src/trufflehog-mapper';
```
Amndeep7, line 36: add to the hdf converters readme

Amndeep7, line 36: add the mapper to the list of mappers supported in the supported formats tooltip in the load modal within heimdall

Amndeep7 (changes requested):

libs/hdf-converters/src/trufflehog-mapper.ts
andytang99: I will just show the detector type as the id instead.

libs/hdf-converters/src/trufflehog-mapper.ts
andytang99: Resolved to use space instead of underscore

libs/hdf-converters/src/trufflehog-mapper.ts
andytang99: resolved to use array of size one

Amndeep7 (approved):

PR #6018, opened 2024-07-23
Swapped adm-zip out for zip.js

Amndeep7:
`adm-zip` was causing issues such as the following

```
 warning  in ../../node_modules/adm-zip/util/fileSystem.js

Module not found: Error: Can't resolve 'original-fs' in '/Users/amann/work/auto/dependency-track-mapper/heimdall2/node_modules/adm-zip/util'
```

which was causing heimdall to have runtime errors and not display itself properly.  This had been throwing warnings before but had started to actually cause this runtime problem when making other changes so I decided to fix it.

Easiest thing to do was to replace it wholesale.  Sidebenefit is that I found `zip.js` which has built-in types so we have that benefit.

The only problem is that it is not a commonjs dependency so we have to do a dynamic import which is a bit ugly due to the lack of top level await that we're currently suffering from.  I'm working on getting us able to use that functionality, but that shouldn't be a blocker for this pr.

Also I cleaned up the files we touched a little bit.

### DIFF
--- a/apps/frontend/src/utilities/export_util.ts
+++ b/apps/frontend/src/utilities/export_util.ts
@@ -1,10 +1,18 @@
-import ZipFile from 'adm-zip';
 import {saveAs} from 'file-saver';
+let zip: any;
+try {
+  (async () => {
+    zip = await import('@zip.js/zip.js');
+  })();
+} catch (e) {
+  console.log(`zip module import failed: ${e}`);
+}
 
 type File = {
   filename: string;
   data: string;
 };
+
 export async function saveSingleOrMultipleFiles(
   files: File[],
   filetype: string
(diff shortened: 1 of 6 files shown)


apps/frontend/src/utilities/export_util.ts
```diff
@@ -1,10 +1,18 @@
-import ZipFile from 'adm-zip';
 import {saveAs} from 'file-saver';
+let zip: any;
```
Amndeep7, line 2: Unfortunately necessary for now due to dynamic import.  The blocker is top level await.  I'm working on getting that enabled.

Amndeep7:
Currently failing the frontend tests due to node not finding the TransformStream object for some reason even though node supports it as of v18.  It works fine on my machine but that's cause it's a browser running the frontend code.  I need to see if we need a polyfill or if I can muck with stuff to get node to work properly.

PR #5986, opened 2024-07-10
SBOM Mapper

charleshu-8:

PR #6035, opened 2024-07-26
SBOM View

kemley76:
Adding ability to view SBOM results both in the main results table and in a new, separate SBOM specific view.


kemley76:
## Status as of 8/16/24
Dependent on #5986 

### Features added
- SBOM view with component table and dependency tree
- ability to choose what columns to display in component table (name, description, version, number of dependencies, etc.)
- any vulnerabilities affecting a component appear in the table as a button for navigating back to results view
- each component has an expandable section that contains all the rest of the component information (properties, external references, licenses, vulnerabilities, dependencies, parents, etc.)
- ability to filter components by severity, bom-ref, and freeform search
- vulnerabilities that impact an SBOM component have a button to display them in the SBOM view's component table
- dependency tree view that shows the dependency relationships between components
- ability to navigate to components that match a given filter
- an indicator for if a component in the tree has any vulnerabilities

### What is left to add
***I think the SBOM view is functional as it is, but these are just what I would probably add if I had enough time*** 
- it would be nice to display the vulnerabilities in the dependency tree view a bit better. Colored chips might be nice, but there can be any number of vulnerabilities present on a component, so that might be tricky. It would also be nice to indicate if a component has any vulnerabilities in any of its descendants. 
- Information panels and tooltips in various menus (search bar, settings icon, filter icon, SBOM view as a whole) to explain how to use the SBOM view to its fullest. 
- Automated frontend/Cypress tests. None of the SBOM view is being validated by tests at the moment. The tests should at the very least, load in a good sample file and ensure that the right amount of components load in the table and the filtering works.
- There seems to be a small issue with navigation in the tree view. Some components either aren't present in the dependency tree (might be an issue with the SBOM itself) or cannot be found with the filter navigation feature ("Go" chip that takes user to dependency tree view). I noticed this with the [dropwizard-vulns](https://github.com/mitre/heimdall2/blob/94f4879f38009a02d4a8962b541006e8a0a3c13e/libs/hdf-converters/sample_jsons/cyclonedx_sbom_mapper/sample_input_report/dropwizard-vulns.json) sample file.
- A change over time view. This has not been started at all, so it can be a separate PR. It would allow users to compare multiple SBOMs of the same target and see how it evolved over time (packages added/removed, version updates, authorship changes, vulnerabilities/patches, etc.)

PR #6052, opened 2024-08-01
Remove Tag Cleaning on Frontend

### LINKED ISSUE #6038: HTML export doesn't render html properly in some cases [BUG]
**Describe the bug**
When exporting some hdf files to HTML, the resulting HTML file contains unrendered HTML tags that are not present in the original HDF. 

**To Reproduce**
1. Go to [https://heimdall-lite.mitre.org/](https://heimdall-lite.mitre.org/)
2. Import sample "Fortify Heimdall_tools Sample"
3. Note the titles of the first controls (no HTML tags present)
4. Export as HTML
5. View HTML and note unrendered html tags in output

**Expected behavior**
The tags should be rendered as HTML properly, not as raw text. The tags appear to be coming from the html template file used to generate the output HTML. 

image

charleshu-8:
- Make frontend display any tags given in the OHDF object.


charleshu-8:
The tags stem from how the Fortify mapper doesn't seem to clean up the XML tags that are converted and used in its mappings. Adding a tag filter on the HTML export side removes these tags to mirror how we do things on the frontend, but it does seem like some context is lost as a result of the tags encoding some information.

EX)
- Pre-filter:
`&lt;Content&gt;&lt;Paragraph&gt;Attackers are able to control the file system path argument to &lt;Replace key&#x3D;&quot;PrimaryCall.name&quot;&#x2F;&gt; at &lt;Replace key&#x3D;&quot;PrimaryLocation.file&quot;&#x2F;&gt; line &lt;Replace key&#x3D;&quot;PrimaryLocation.line&quot;&#x2F;&gt;, which allows them to access or modify otherwise protected files.&lt;AltParagraph&gt;Allowing user input to control paths used in file system operations could enable an attacker to access or modify otherwise protected system resources.&lt;&#x2F;AltParagraph&gt;&lt;&#x2F;Paragraph&gt;&lt;&#x2F;Content&gt;`

- Post-filter:
`Attackers are able to control the file system path argument to  at  line , which allows them to access or modify otherwise protected files.Allowing user input to control paths used in file system operations could enable an attacker to access or modify otherwise protected system resources.`

As this behavior is mirrored on the frontend as well, I recommend addressing this issue in a rework of the Fortify mapper instead.

Amndeep7 (changes requested):
I do not agree with this PR.

I am actually fine with the raw HTML showing up in the results cause now we can see the full information.

I think the correct direction would be to fix both the frontend and the HTML export to allow for displaying HTML formatted data.

charleshu-8:
Pivoting this to instead remove tag cleaning on Heimdall frontend.

PR #6112, opened 2024-08-16
Remove passport openid

Amndeep7:


georgedias (approved):
very straight forward

PR #6155, opened 2024-09-04
Remove libxmljs2

Amndeep7:
Manually delete libxmljs2 which is an optional dependency for @cyclonedx/cyclonedx-library. this optional dependency adds architecture specific binaries to the node_modules, which we do not want since they get packaged with hdf-converters. hdf-converters is used in the saf cli where we are trying to make noarch rpm binaries.

### DIFF
--- a/apps/frontend/package.json
+++ b/apps/frontend/package.json
@@ -127,6 +127,10 @@
     "ts-jest": "^29.1.0",
     "vue-jest": "^3.0.7"
   },
+  "resolutions": {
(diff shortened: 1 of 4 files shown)


Amndeep7:
I'll fix the merge conflict once this is approved and all the other tests pass @wdower.  Dependabot got some changes in while I was working this I guess.

Amndeep7:
nvm there seems to be a netlify issue that requires more thinking @wdower

PR #6195, opened 2024-09-17
Neuvector mapper

jtquach1:
Related to https://github.com/mitre/saf/pull/2912


aaronlippold:
Does this have both forward and backwards conversion? 
Does this also integrate into the SAF CLI?

jtquach1:
> Does this have both forward and backwards conversion? Does this also integrate into the SAF CLI?

@Amndeep7 mentioned I didn't need to do backwards conversion (forwards conversion only). I didn't make a PR against saf-cli yet, but going to put that up and link/reference here

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+  };
+}
+
+export class NeuvectorMapper extends BaseConverter {
+  withRaw: boolean;
+  rawData: NeuvectorScanJson;
+  getModule: (moduleName: string) => RESTScanModule | undefined;
+
+  memoizedGetModule(): (moduleName: string) => RESTScanModule | undefined {
```
Amndeep7, line 176: i don't think this function is actually doing any memoization since you have no cache to store intermediate results.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  mappings: MappedTransform<
+    ExecJSON.Execution & {passthrough: unknown},
+    ILookupPath
+  > = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion,
+      target_id: null
```
Amndeep7, line 191: you do not need to supply optional attributes that have no-op values in them such as this null

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+    },
+    version: HeimdallToolsVersion,
+    statistics: {
+      duration: null
+    },
+    profiles: [
+      {
+        name: 'Neuvector',
```
Amndeep7, line 199:
```suggestion
        name: 'NeuVector Scan',
```

Please correct the capitalization of 'NeuVector' to match their preference wherever you use it, especially in user facing areas.  Also please match the patterns that we've set in other mappers, ex. making 'name' equal to 'Tool Scan'.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+        title: null,
```
Amndeep7, line 200: This should probably specify target information, ex. some combination of registry, repository, and tag (and maybe even the imageid/digest if we feel like we can work it in)

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+                path: 'description',
+                transformer: (description: string) =>
+                  nistTags(cweTags(description))
+              },
+              ghsa: {path: 'name', transformer: ghsaTag},
+              rhsa: {path: 'name', transformer: rhsaTag},
+              // `score` is confirmed to be CVSS v2 in https://github.com/neuvector/scanner/blob/765fb1db2cf678ea6c6d386f3eb0f720311d745a/cvetools/cvesearch.go#L1416
+              score: {path: 'score'},
+              vectors: {path: 'vectors'},
+              // In the neuvector/scanner dashboard, the CVSS v2 and v3 scores are selectable by a dropdown.
```
Amndeep7, line 236: unless you were able to spin up the application, we didn't confirm this.  it was just our conjecture.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+                  this.rawData.report.modules
+                    ?.find((module) => module.cves?.find(cveIdMatches(name)))
+                    ?.cves?.find(cveIdMatches(name))?.status
+              },
+              feed_rating: {path: 'feed_rating'},
+              link: {path: 'link'},
+              // Both fields are Unix epoch timestamps, in seconds.
+              published_timestamp: {path: 'published_timestamp'},
+              last_modified_timestamp: {path: 'last_modified_timestamp'},
+              ...universalTags()
```
Amndeep7, line 258: i don't see why they need to be split out into a separate function.  just leave them here.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+        status: 'loaded',
+        controls: [
+          {
+            path: 'report.vulnerabilities',
+            key: 'id',
+            tags: {
+              cves: {path: 'cves'},
+              cpes: {
+                path: 'cpes',
+                transformer: (cpes: string[]) => (!cpes ? undefined : cpes)
```
Amndeep7, line 220: https://github.com/mitre/heimdall2/blob/master/libs/hdf-converters/src/utils/global.ts#L82

if you want to conditionally supply attributes, you might be interested in leveraging this pre-existing helper function

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+            source_location: {ref: {path: 'file_name'}},
+            id: {
+              transformer: (data: Record<string, any>) =>
+                `${data.name}/${data.package_name}/${data.package_version}`
+            },
+            desc: {path: 'description'},
+            impact: {
+              transformer: (data: Record<string, any>) => {
+                const score = data.score_v3 ?? data.score;
+                return Number((score / 10).toFixed(1));
```
Amndeep7, line 270: I don't think we need to cap the number of decimal points, but we can confirm with aaron or eugene.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+              }
+            },
+            code: null,
+            results: [
+              {
+                status: ExecJSON.ControlResultStatus.Failed,
+                code_desc: '',
+                message: {
+                  transformer: (data: Record<string, any>) => {
```
Amndeep7, line 279: another example where i prefer unknown over any

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+            ]
+          }
+        ],
+        sha256: ''
+      }
+    ],
+    passthrough: {
+      transformer: (data: Record<string, any>): Record<string, unknown> => {
+        return {
+          auxiliary_data: [{name: '', data: _.omit([])}], //Insert service name and mapped fields to be removed
```
Amndeep7, line 298: Don't forget to fill this out.  We're only going to omit report.vulnerabilities and whatever report.whatevers gets mapped such as potentially the target information since we're going to use that in the profile.title.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+import _ from 'lodash';
+import {version as HeimdallToolsVersion} from '../package.json';
+import {BaseConverter, ILookupPath, MappedTransform} from './base-converter';
+import {CweNistMapping} from './mappings/CweNistMapping';
+import {DEFAULT_UPDATE_REMEDIATION_NIST_TAGS} from './utils/global';
+
+/* Types are generated with Tygo, from original Golang source code to TypeScript, and tweaked to reflect actual outputted JSON. */
+type RESTVulnerability = {
+  name: string;
+  score: number /* float32 */;
```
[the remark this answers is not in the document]
jtquach1: Yeah, the `/* ... */`'s are artifacts from Tygo. My best guess/understanding is that Tygo uses the long comments when the types are getting converted to a less precise type, like `number` or `any` where `any` is a special type. Tygo uses `//` when trying to preserve normal comments in the source code. 
There was a `TODO` comment in the generated RESTScanSecret.suggestion that triggered a code smell, which I manually removed.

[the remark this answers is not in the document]
jtquach1: I could prepend the Tygo-generate comments with `Tygo: ` to not get them confused with my code manual comments

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+  automated: boolean;
+  description: string;
+  remediation: string;
+  tags?: string[]; // Tags provide list of compliance that related to the cis test item.
+  tags_v2?: Record<string, any /* share.TagDetails */>; // TagsV2 provide compliance details for each compliance tag
+};
+
+type RESTScanSecret = {
+  type: string; // the secret description
+  evidence: string; // found in a cloaked string
```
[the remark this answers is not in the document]
jtquach1: these are Tygo-generated, but also, not sure myself

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+export type NeuvectorScanJson = {
+  report: RESTScanRepoReport;
+  error_message: string;
+};
+
+const CWE_NIST_MAPPING = new CweNistMapping();
+
+function cweTags(description: string): string[] | undefined {
+  const regex = /CWE-\d{3}/g;
+  return description?.match(regex)?.flat();
```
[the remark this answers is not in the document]
jtquach1: I definitely missed that

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,311 @@
...
+function nistTags(cweTags: string[] | undefined): string[] {
+  const identifiers =
+    cweTags?.map((tag: string) => tag.match(/\d{3}/g)?.[0] ?? [])?.flat() ?? [];
+  return CWE_NIST_MAPPING.nistFilter(
+    identifiers,
+    DEFAULT_UPDATE_REMEDIATION_NIST_TAGS
+  );
+}
+
+function ghsaTag(name: string): string | undefined {
```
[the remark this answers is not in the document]
jtquach1: I can remove them, but just wondering, are GHSAs and RHSAs unnecessary to list in the tags because CVEs/CWEs are more precise in what information they display or that they're primarily mapped in heimdall2?

libs/hdf-converters/src/neuvector-mapper.ts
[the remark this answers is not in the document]
Amndeep7: nah, the weird /* inline */ syntax is distinguishing enough

libs/hdf-converters/src/neuvector-mapper.ts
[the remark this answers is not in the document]
Amndeep7: if we're unsure, then let's remove it.

libs/hdf-converters/src/neuvector-mapper.ts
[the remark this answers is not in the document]
Amndeep7: usually cves or other unique identifiers end up being control.ids.  cwes and nist tags are a) more general purpose information and b) used for doing mappings.

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,262 @@
...
+    ExecJSON.Execution & {passthrough: unknown},
+    ILookupPath
+  > = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion
+    },
+    version: HeimdallToolsVersion,
+    statistics: {
+      duration: null
```
Amndeep7, line 131: please get rid of all the unused attributes such as the ones that get assigned null

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,262 @@
...
+              severity: {path: 'severity'},
+              source: {
+                path: 'package_name',
+                transformer: (packageName: string) =>
+                  this.getModule(packageName)?.source
+              },
+              status: {
+                path: 'name',
+                transformer: (name: string) =>
+                  this.rawData.report.modules
```
Amndeep7, line 185: i would still do the memoized function, just implement it correctly

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,262 @@
...
+                    ?.find((module) => module.cves?.find(cveIdMatches(name)))
+                    ?.cves?.find(cveIdMatches(name))?.status
+              },
+              feed_rating: {path: 'feed_rating'},
+              link: {path: 'link'},
+              published_timestamp: {path: 'published_timestamp'},
+              last_modified_timestamp: {path: 'last_modified_timestamp'},
+              envs: {
```
Amndeep7, line 193: the conditional attribute thing applies here too

libs/hdf-converters/src/neuvector-mapper.ts
```diff
@@ -0,0 +1,262 @@
...
+          }
+        ],
+        sha256: ''
+      }
+    ],
+    passthrough: {
+      transformer: (
+        data: NeuVectorScanJson
+      ): {
+        auxiliary_data: {name: string; data: Pick<never[], keyof never[]>}[];
```
Amndeep7, line 241: i feel like the 'never's aren't helping here lol

Amndeep7 (changes requested):

Amndeep7 (approved):

Issue #4727, opened 2023-07-18
Update Heimdall Ironbank Deployment to have a "release-latest" tag


em-c-rod:

Amndeep7:
heimdall2/heimdall-lite provide a release-latest

heimdall2-mainline/heimdall-lite-mainline provide a latest

The Heimdall2 entry is now merged with both heimdall2 and mainline, but I am still working on having the superfluous Iron Bank entry deleted.

Heimdall-lite and the SAF CLI are still WIP for merge->delete.

As such I am going to mark this as closed.

PR #6448, opened 2024-12-04
test workflow dispatch

Amndeep7:


.github/workflows/release-server-to-docker.yml
```diff
@@ -37,20 +46,28 @@ jobs:
...
+      - name: Clone and update local copy of Iron Bank repo to have the latest tags and digest
+        working-directory: ../ironbank_heimdall
+        run: |
+          git clone https://repo1.dso.mil/dsop/mitre/security-automation-framework/heimdall2.git .
+          git switch -c "${{ env.VERSION_UNDERSCORED }}" # swap to sha for mainline
+          yq e -i '.args.HEIMDALL_VERSION="${{ steps.format-tag.outputs.replaced }}" | .tags[0]="${{ steps.format-tag.outputs.replaced }}" | .labels."org.opencontainers.image.version"="${{ steps.format-tag.outputs.replaced }}" | .resources[1].tag="mitre/heimdall2:${{ steps.format-tag.outputs.replaced }}" | .resources[1].url="docker://docker.io/mitre/heimdall2@${{ env.DOCKER_SHA }}"' hardening_manifest.yaml
+          sed -i s/HEIMDALL_VERSION=\.\*/HEIMDALL_VERSION=${{ steps.format-tag.outputs.replaced }}/ Dockerfile
+          git diff
+          git add hardening_manifest.yaml Dockerfile
+          git -c "user.name=Automated Heimdall Release" -c "user.email=saf@groups.mitre.org" commit -s -m "Updating Heimdall to ${{ steps.format-tag.outputs.replaced }}"
```
em-c-rod, line 66: Do you still want to change the user.email? This seems appropriate to me.

em-c-rod:
https://github.com/mitre/heimdall2/pull/6448

.github/workflows/release-server-to-docker.yml
```diff
@@ -9,18 +9,27 @@ on:
...
-      - name: Run string replace # remove the v from the version number before using it in the docker tag
+      - name: Run string replace to remove the v from the version number before using it in the docker tag
         uses: frabert/replace-string-action@v2
         id: format-tag
         with:
           pattern: 'v'
           string: '${{ github.event.release.tag_name || github.event.inputs.version}}'
           replace-with: ''
           flags: 'g'
+      - name: Convert periods to underscores in version string
```
Amndeep7, line 29: wrote underscores, but meant hyphens.  need to change the var name and comment

.github/workflows/release-server-to-docker.yml
Amndeep7: I thought it'd be more exact to have the releaser's id tied to the pr but you're right that we can probs just use the group email

.github/workflows/release-server-to-docker.yml
```diff
@@ -9,18 +9,27 @@ on:
         description: 'Version'     
         required: true
 
+env:
+  IRONBANK_HEIMDALL_PROJECT_ID: 5450 # this is for heimdall (non mainline) - I think these can be in-line envs instead of supplied by github repo/org level secrets/values since each push/release workflow will have a unique ironbank id due to the mainline vs release + heimdalllite vs heimdall matrix
```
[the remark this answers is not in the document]
Amndeep7: this comment was mostly just a note for myself.  plan is to delete it once the work is completed.  basically just some word vomit to explain that i think it's fine to leave the project id as an envvar here that will change in each workflow instead of using a repo level secret.

Amndeep7:
> #4727

@em-c-rod is explicitly calling it 'release-latest' vs 'latest' necessary since they will be two projects with the current 'heimdall' one tracking releases and the 'heimdall-mainline' tracking the master branch?

.github/workflows/release-server-to-docker.yml
Amndeep7: https://github.com/mitre/heimdall2/pull/6448/commits/0e0e0baa5a50fd394c36c234fc2b87ae56985a45

.github/workflows/release-server-to-docker.yml
```diff
@@ -30,27 +39,36 @@ jobs:
...
+          DEFAULT_BRANCH=$(curl --header "PRIVATE-TOKEN: ${{ secrets.SAF_IRONBANK_REPO1_PAT }}" "https://repo1.dso.mil/api/v4/projects/${{ env.IRONBANK_HEIMDALL_PROJECT_ID }}" | jq -r ".default_branch") # likely 'development'
+          git push 'https://${{ secrets.SAF_IRONBANK_REPO1_USERNAME }}:${{ secrets.SAF_IRONBANK_REPO1_PAT }}@repo1.dso.mil/dsop/mitre/security-automation-framework/heimdall2.git' "${{ env.VERSION_HYPHENATED }}":"${ISSUE}-update-heimdall-to-${{ env.VERSION_HYPHENATED }}"
+          curl -X POST --header "PRIVATE-TOKEN: ${{ secrets.SAF_IRONBANK_REPO1_PAT }}" --header "Content-Type: application/json" --data "{\"allow_collaboration\": true, \"squash\": true, \"remove_source_branch\": true, \"source_branch\": \"${ISSUE}-update-heimdall-to-${{ env.VERSION_HYPHENATED }}\", \"target_branch\": \"${DEFAULT_BRANCH}\", \"title\": \"Resolve \\\"Update Heimdall to ${{ steps.format-tag.outputs.replaced }}\\\"\", \"description\": \"Closes #${ISSUE}\"}" "https://repo1.dso.mil/api/v4/projects/${{ env.IRONBANK_HEIMDALL_PROJECT_ID }}/merge_requests"
```
Amndeep7, line 74: They explicitly asked to not have 'closes' in the description since it messes up their process.  The only complaint for this happened on a non-release issue, but I guess we should implement it here too.

.github/workflows/release-server-to-docker.yml
```diff
@@ -30,27 +39,36 @@ jobs:
...
+          sed -i s/HEIMDALL_VERSION=\.\*/HEIMDALL_VERSION=${{ steps.format-tag.outputs.replaced }}/ Dockerfile
+          git diff
+          git add hardening_manifest.yaml Dockerfile
+          git -c "user.name=Automated Heimdall Release" -c "user.email=saf@groups.mitre.org" commit -s -m "Updating Heimdall to ${{ steps.format-tag.outputs.replaced }}"
+      - name: Create issue, branch, and merge request for Iron Bank repo
+        working-directory: ../ironbank_heimdall
+        run: |
+          ISSUE=$(curl -X POST --header "PRIVATE-TOKEN: ${{ secrets.SAF_IRONBANK_REPO1_PAT }}" --header "Content-Type: application/json" --data '{"title": "Update Heimdall to ${{ steps.format-tag.outputs.replaced }}", "labels": "Status::Review"}' "https://repo1.dso.mil/api/v4/projects/${{ env.IRONBANK_HEIMDALL_PROJECT_ID }}/issues" | jq ".iid")
```
Amndeep7, line 71: new workflow process

if the issue title is new (ex. update heimdall to v3 and has a description that specifies the tag), then make the mr and update the issue to include the generated mr id - probably will require the issue description to be structured data in some capacity so we can process it
if the issue title is the same (ex. update heimdall to latest commit) then update the description (specify the new commit), make a new mr and update the issue with the new id, and take the old id and use it to close the old mr

.github/workflows/release-server-to-docker.yml
Amndeep7: implemented in action

.github/workflows/release-server-to-docker.yml
Amndeep7: For now I no longer think that this will be an issue since workflows only trigger on human action.

Amndeep7:
> > #4727
> 
> @em-c-rod is explicitly calling it 'release-latest' vs 'latest' necessary on the ironbank side since they will be two projects with the current 'heimdall' one tracking releases and the 'heimdall-mainline' tracking the master branch? i.e. you will be able to use 'latest' on either.

@em-c-rod please ignore this comment.  My intent is to combine them on the user side so we will be retaining the release-latest vs latest nomenclature.

PR #6538, opened 2025-01-17
Rename Non-packaged libraries to be under "common"

em-c-rod:
This is cherry-picked work taken from the FIPs PR to nest the interfaces library under the name "common". The idea is that additional custom packages can be put in this "common" library instead of adding them separating as their own thing.

I separated it to make the FIPs PR easier to debug and review.


Amndeep7 (approved):
Wanted to see if I could get the nested structure to work properly but it was getting to be too large since it started getting into the tsconfig stuff all over the place.  This builds and runs so it's fine.

PR #6579, opened 2025-01-29
Cypress console logs

Amndeep7:


Amndeep7:
image

https://github.com/mitre/heimdall2/actions/runs/13025550260/job/36333906688?pr=6579

Passing run but note how we can look at the logs of the test that eventually decided to pass after repeated attempts.  You can see that the problem was just with the request to the splunk server blowing up for no reason and returning a 401 as opposed to ex. you breaking the app.

Does this mean that we should probably fix up the splunk<->heimdall code so that it's less likely to 401?  Yes.  But at least now we can more confidently review the e2e cypress test logs and say "flakey" and move on as opposed to being worried.

aaronlippold (approved):

Issue #6583, opened 2025-01-29
[BUG]  null reference in sonarqube parser


Jiri-Stary:
**Describe the bug**
Inccorect handling of sonarqube issues such as kubernetes one in sonarqube-mapper.ts that do not have summary or CVE

 TypeError: Cannot read properties of undefined (reading 'match')

**To Reproduce**

 mitre/saf:1.4.8 convert  sonarqube2hdf -n *** -u *** --auth *** -o ./hdf/sonarqube_scan.json

use example sonar issue

  {
            "key": "57b3733b-d416-4b4e-87da-b77d4edde2ab",
            "rule": "kubernetes:S6870",
            "severity": "MAJOR",
            "component": "myproject:helm/mlr/templates/jupyter/deployment.yaml",
            "project": "myproject",
            "line": 34,
            "hash": "723c0daa435bdafaa7aa13d3ae06ca5e",
            "textRange": {
                "startLine": 34,
                "endLine": 34,
                "startOffset": 19,
                "endOffset": 30
            },
            "flows": [],
            "status": "OPEN",
            "message": "Specify a storage limit for this container.",
            "effort": "5min",
            "debt": "5min",
            "author": "",
            "tags": [],
            "creationDate": "2025-01-23T14:01:27+0000",
            "updateDate": "2025-01-23T14:01:27+0000",
            "type": "VULNERABILITY",
            "branch": "Jiri-Stary-patch",
            "scope": "MAIN",
            "quickFixAvailable": false,
            "messageFormattings": [],
            "codeVariants": [],
            "cleanCodeAttribute": "COMPLETE",
            "cleanCodeAttributeCategory": "INTENTIONAL",
            "impacts": [
                {
                    "softwareQuality": "MAINTAINABILITY",
                    "severity": "MEDIUM"
                },
                {
                    "softwareQuality": "SECURITY",
                    "severity": "MEDIUM"
                }
            ],
            "issueStatus": "OPEN",
            "prioritizedRule": false
        },

**Expected behavior**
not crash

- Can you please add better undefined input handling ? 
  issue?.summary instead of   issue.summary should be enough

```
function parseNistTags(issue: Issue): string[] {
  const tags: string[] = [];
  issue.sysTags?.forEach((sysTag) => {
    if (sysTag.toLowerCase().startsWith('owasp-')) {
      const identifier = [
        sysTag.toLowerCase().replace('owasp-', '').toUpperCase()
      ];
[...]

PR #6610, opened 2025-02-07
fix: fix null reference in sonarqube mapper

Jiri-Stary:
fix: https://github.com/mitre/heimdall2/issues/6583


Amndeep7:
OBE due to sonarqube mapper rewrite https://github.com/mitre/heimdall2/pull/6952

PR #6616, opened 2025-02-10
Fix Apply Attestation: "TypeError: Cannot read properties of undefined (reading 'status')"

em-c-rod:
When applying an attestation to an overlay file, we were getting the following error:
`"TypeError: Cannot read properties of undefined (reading 'status')"`
because the overlay file had the same control listed in multiple profile objects of the HDF file. Not every control instance has results because of it being an overlay file. This will catch that issue and apply the attestation once the results exist.
- [x] Added a catch for overlay files which may have controls with no results. 
- [x] Test added

### DIFF
--- a/libs/hdf-converters/sample_jsons/attestations/triple_overlay_profile_sample.json
+++ b/libs/hdf-converters/sample_jsons/attestations/triple_overlay_profile_sample.json
@@ -1,0 +1,1247 @@
+{
+    "platform": {
+        "name": "centos",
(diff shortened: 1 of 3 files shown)


Amndeep7 (changes requested):
We should probably add another test to check for this overlay case.

Amndeep7:
Also I pushed up a commit @em-c-rod, feel free to revert if you don't want it.

Amndeep7 (approved):

Issue #6816, opened 2025-04-15
[BUG] Callback Redirect to OIDC when using okta strategy


clem-field:
@Amndeep7 

**Describe the bug**
Okta integration is successfully retrieving a token and it is being passed back to the application through NGINX. The callback is setting the client to 'disabled' and overriding to OIDC and no longer performing Okta flow. Overriding the exception handler results in an 'Unauthorized' 401 in the okta flow at time of callback.

**To Reproduce**
1. Using Heimdall Server 
2. Deploy with Okta strategy and remove all local login and self registration (requires on boarding to okta [domain, client id, client secret]
3. Navigate to deployed location
4. Attempt login via okta

**Expected behavior**
Should follow okta path to authenticate user, retrieve openID, email, profile and log user in (build user if not already built)

**Screenshots**
Follow path with RCFederation and / or respective browser dev tool/s.

**Desktop (please complete the following information):**
 - OS: Windows 11+, Mac Sequoia 15.4
 - Browser Safari, Chrome, Edge
 - Version [Multiple]

**Additional context**
[apps/backend/src/authn/authn.contoller.ts ](https://github.com/mitre/heimdall2/blob/f974ab0be1bd078716b480b061c9deffd52547f2/apps/backend/src/authn/authn.controller.ts#L103) implements a AuthenticationExceptionFilter function that overrides the implemented strategy to OIDC https://github.com/mitre/heimdall2/blob/f974ab0be1bd078716b480b061c9deffd52547f2/apps/backend/src/filters/authentication-exception.filter.ts#L17

Teams have implemented more verbose logging in the authentication filter:

import {ArgumentsHost, Catch, ExceptionFilter} from '@nestjs/common';
import * as _ from 'lodash';
import {ConfigService} from '../config/config.service';

@Catch(Error)
export class AuthenticationExceptionFilter implements ExceptionFilter {
  configService = new ConfigService();
  catch(exception: Error, host: ArgumentsHost): void {
    const ctx = host.switchToHttp();
    const request = ctx.getRequest();
    const response = ctx.getResponse();
    // Log detailed error info
    console.error('Authentication Error', {
        message: exception.message,
        stack: exception.stack,
        authInfo: _.get(request, 'authInfo'),
        query: request.query,
        headers: request.headers,
    });
    // Remove OIDC-specific check
    // Optionally, add logic for actual auth flow (e.g., SAML or OAuth)
    const authError = _.get(request, 'authInfo.message') || exception.message;
[...]

Issue #3980, opened 2023-01-18
OIDC PKCE Support


sudheer-v:
Hi Team

I'm trying to use OIDC external authentication(Pingid). PingID has PKCE enforced on IDP, does Heimdall2 OIDC config supports PKCE?

when I try to configure Heimdall2 it is throwing code_challenge required error. What is the env variable to turn on PKCE when configuring oidc?

please guide me on this.

Regards,
Sudheer

Amndeep7:
Hello @sudheer-v, I apologize for the delay in response to this issue.  If this is still an issue you're facing, please see https://github.com/mitre/heimdall2/pull/6820 and supply either the OIDC_USES_PKCE_S256 or OIDC_USES_PKCE_PLAIN envvars (set to true) as necessary.  For more details, see the underlying implementation here: https://github.com/GovTechSG/passport-openidconnect/tree/master

PR #6820, opened 2025-04-16
OIDC and Okta auth strategy improvements and bugfixes

Amndeep7:
Resolves: #6837
Resolves: #6816
Resolves: #3980

NOTE: This will require a version bump to occur on next release


aaronlippold:
Thank you for working on these authentication improvements! I noticed a potential issue with removing the 'openid' scope from the OIDC and Okta strategies.

According to the OpenID Connect specification, the 'openid' scope is required for OpenID Connect requests. Removing it could break compliance with the protocol.

From the [OpenID Connect Core 1.0 specification](https://openid.net/specs/openid-connect-core-1_0.html):
> "OpenID Connect requests MUST contain the openid scope value. If the openid scope value is not present, the behavior is entirely unspecified."

If you're experiencing issues with the authentication process, there might be other solutions that maintain protocol compliance while addressing the specific problems you're encountering.

I suggest keeping the 'openid' scope while incorporating your other valuable improvements to the exception handling:

```typescript
// For OIDC strategy
{
  // other configuration...
  scope: 'openid profile email' // Keep 'openid' for compliance
}

// For Okta strategy
{
  // other configuration...
  scope: 'openid email profile' // Keep 'openid' for compliance
}
```

The authentication exception filter improvements you've made are excellent and would be valuable to incorporate regardless of the scope issue.

aaronlippold:
I really like the improvements to the `AuthenticationExceptionFilter`! The parameterized constructor for handling different authentication types is a great enhancement. Here are some additional ideas that could further improve it:

1. **Add correlation ID tracking**: This would help with traceability across logs.

```typescript
import {createLogger, generateCorrelationId} from '../utils/logger';

// In the catch method:
const correlationId = generateCorrelationId();
const startTime = Date.now();

// Add to logs:
this.logger.warn({
  correlationId,
  message: {
    event: 'authentication_exception', 
    error: exception.message,
    authenticationType: this.authenticationType,
    duration_ms: Date.now() - startTime
  }
});
```

2. **Add correlation ID to response headers**: This helps with client-side troubleshooting.

```typescript
// Before redirecting:
response.header('X-Correlation-ID', correlationId);
```

3. **Categorize authentication failures**: This can help with security analysis.

```typescript
// Determine the failure category
let failureCategory = 'unknown';
const errorMessage = exception.message.toLowerCase();

if (errorMessage.includes('invalid') || errorMessage.includes('incorrect')) {
  failureCategory = 'invalid_credentials';
} else if (errorMessage.includes('not found') || errorMessage.includes('unknown')) {
  failureCategory = 'unknown_user';
} else if (errorMessage.includes('disabled') || errorMessage.includes('locked')) {
  failureCategory = 'account_disabled';
}

// Include in logs:
this.logger.warn({
  correlationId,
  message: {
    event: 'authentication_exception',
    failureCategory,
    // other details...
  }
});
```

4. **Sanitize headers in logs**: To avoid logging sensitive information.

```typescript
const sanitizedHeaders = _.omit(request.headers, ['cookie', 'authorization']);
```

These enhancements would build on your excellent work with the parameterized authentication exception filter while adding more observability and security features.

aaronlippold:
Here's a more comprehensive review with additional suggestions:

## Overall Improvements

Your PR makes several valuable improvements:
1. Parameterized exception filter to handle different auth types
2. Changed redirects from 301 to 302 (good practice for auth redirects)
3. Enhanced error messages by combining auth info with exception message
4. Added better logging for auth errors
5. Fixed typo in comment ("occours" -> "occurs")
6. Added reference to GitHub issue for the state verification bug

## Additional Enhancement Suggestions

### 1. Make the logger more reusable

Instead of creating a new logger inside the filter, consider using the centralized logger utility:

```typescript
import {createLogger} from '../utils/logger';

// Inside the class:
private readonly logger = createLogger('AuthExceptionFilter');
```

### 2. Support all authentication types

Currently, the type parameter only supports 'oidc' and 'okta'. Consider expanding it to all auth strategies:

```typescript
constructor(
  readonly authenticationType?: 'oidc' | 'okta' | 'github' | 'gitlab' | 'google' | 'local' | 'ldap'
) {}
```

### 3. Improve error message formatting

The current implementation concatenates authInfo message and exception message with a newline, but this could be more robust:

```typescript
const authError = `${_.get(request, 'authInfo.message') || ''}${
  _.get(request, 'authInfo.message') ? '\n' : ''
}${exception.message}`.trim();
```

### 4. Add type safety for error information

Instead of using a plain object for error info, consider using a proper type:

```typescript
interface AuthErrorInfo {
  message: string;
  stack?: string;
  authInfo?: any;
  query: Record<string, any>;
  headers: Record<string, any>;
  authenticationType?: string;
  timestamp: string;
}

// Then use it:
const errInfo: AuthErrorInfo = {
  message: exception.message,
  stack: exception.stack,
  // other properties...
};
```

### 5. Enhance the logger output format

The current format could be improved for better readability:

```typescript
this.logger.warn({
  message: {
    event: 'authentication_exception',
    error: exception.message,
    authenticationType: this.authenticationType,
    timestamp: new Date().toISOString(),
    requestPath: request.url,
    // other properties...
  }
});
```

### 6. Add proper testing

Consider adding tests for your enhanced exception filter to ensure it works correctly with different authentication strategies:

```typescript
// In authentication-exception.filter.spec.ts
describe('AuthenticationExceptionFilter', () => {
  it('should handle OIDC state verification errors', () => {
    // Test implementation
  });
  
  it('should combine auth info with exception message', () => {
    // Test implementation
  });
  
  it('should redirect to the correct authentication endpoint', () => {
    // Test implementation
  });
});
```

### 7. Consider using HttpStatus constants

Instead of hardcoding status code numbers:

```typescript
import {HttpStatus} from '@nestjs/common';

// Then use:
return response.redirect(HttpStatus.FOUND, `/authn/${this.authenticationType}`);
```

### 8. Add option for structured JSON logging

For better integration with logging systems:

```typescript
// Add option to log in JSON format:
const errorJson = JSON.stringify({
  timestamp: new Date().toISOString(),
  level: 'warn',
  context: 'AuthExceptionFilter',
  message: errInfo
});

// Or format as needed for your logging infrastructure
```

These suggestions would make your improvements even more robust while maintaining the excellent core enhancements you've already implemented.

aaronlippold:
**Update on OIDC scope**: After further investigation of the passport-openidconnect source code, I discovered that you're correct to remove the explicit 'openid' scope. The library automatically adds 'openid' to the scope parameter in all requests.

From the library's source code:
```javascript
// From passport-openidconnect/lib/strategy.js
if (scope) {
    params.scope = 'openid ' + scope;
} else {
    params.scope = 'openid';
}
```

So including 'openid' in our explicit scope was actually causing duplication in the request. This change makes our implementation more efficient while still maintaining OpenID Connect compliance. Sorry for the confusion in my initial comment.

Amndeep7:
If we can get confirmation that this is working in the sponsor environment, the only changes that would need to be made are converting the console.logs into actual logs as necessary and swapping the passport strategy library to be the actual one instead of my fork with injected logging.

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -36,14 +37,26 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
         userInfoURL: configService.get('OIDC_USER_INFO_URL') || 'disabled',
         clientID: configService.get('OIDC_CLIENTID') || 'disabled',
         clientSecret: configService.get('OIDC_CLIENT_SECRET') || 'disabled',
-        callbackURL: `${configService.get('EXTERNAL_URL')}/authn/oidc/callback`,
-        scope: 'openid profile email'
+        callbackURL: `${configService.get('EXTERNAL_URL')}/authn/oidc_callback`,
+        pkce:
+          configService.get('OIDC_USES_PKCE_S256') === 'true'
+            ? 'S256'
+            : configService.get('OIDC_USES_PKCE_PLAIN') === 'true'
```
em-c-rod, line 44: What does plain vs. S256 mean? Do you want to make a variable which required the user to specify one of the options instead of several true or false vars? What about if both are set to true? Are they mutually exclusive? How can we make it fool proof for the user?

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -36,14 +37,26 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
...
+              ? 'plain'
+              : undefined,
```
em-c-rod, line 46: Are we catching that this could be undefined anywhere else?

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -36,14 +37,26 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
...
+        scope: ['openid', 'email', 'profile'],
+        skipUserProfile: false,
+        proxy:
+          configService.get('OIDC_USE_HTTPS_PROXY') === 'true'
+            ? true
+            : undefined,
```
em-c-rod, line 52: What is the purpose of setting this to undefined instead of false?

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -36,14 +37,26 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
         userInfoURL: configService.get('OIDC_USER_INFO_URL') || 'disabled',
         clientID: configService.get('OIDC_CLIENTID') || 'disabled',
         clientSecret: configService.get('OIDC_CLIENT_SECRET') || 'disabled',
-        callbackURL: `${configService.get('EXTERNAL_URL')}/authn/oidc/callback`,
-        scope: 'openid profile email'
+        callbackURL: `${configService.get('EXTERNAL_URL')}/authn/oidc_callback`,
+        pkce:
+          configService.get('OIDC_USES_PKCE_S256') === 'true'
```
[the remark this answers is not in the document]
em-c-rod: https://github.com/mitre/heimdall2/wiki/Environment-Variables-Configuration

apps/backend/src/authn/okta.strategy.ts
```diff
@@ -26,50 +25,54 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
...
-        }/oauth2/default/v1/authorize`,
-        tokenURL: `https://${
-          configService.get('OKTA_DOMAIN') || 'disabled'
-        }/oauth2/default/v1/token`,
-        userInfoURL: `https://${
-          configService.get('OKTA_DOMAIN') || 'disabled'
-        }/oauth2/default/v1/userinfo`,
+        issuer:
+          configService.get('OKTA_ISSUER_URL') ||
+          `https://${configService.get('OKTA_DOMAIN')}` ||
```
em-c-rod, line 30: Is the `/oath2/default` not the right location anymore? Was it ever? And is that path determined on the Heimdall side rather than related to the Okta client that Heimdall would be integrated with?

apps/backend/src/authn/okta.strategy.ts
```diff
@@ -26,50 +25,54 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
...
+          'disabled',
+        authorizationURL:
+          configService.get('OKTA_AUTHORIZATION_URL') ||
+          `https://${configService.get('OKTA_DOMAIN') || 'disabled'}/oauth2/v1/authorize`,
```
em-c-rod, line 34: Are the new paths from switching the library?

apps/backend/src/authn/okta.strategy.ts
```diff
@@ -26,50 +25,54 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
...
+        tokenURL:
+          configService.get('OKTA_TOKEN_URL') ||
+          `https://${configService.get('OKTA_DOMAIN') || 'disabled'}/oauth2/v1/token`,
+        userInfoURL:
+          configService.get('OKTA_USER_INFO_URL') ||
+          `https://${configService.get('OKTA_DOMAIN') || 'disabled'}/oauth2/v1/userinfo`,
```
em-c-rod, line 40: If disabled is used, the webpage won't load, right?

apps/backend/src/authn/okta.strategy.ts
```diff
@@ -26,50 +25,54 @@ export class OktaStrategy extends PassportStrategy(Strategy, 'okta') {
...
         clientID: configService.get('OKTA_CLIENTID') || 'disabled',
         clientSecret: configService.get('OKTA_CLIENTSECRET') || 'disabled',
-        callbackURL: `${configService.get('EXTERNAL_URL')}/authn/okta/callback`,
-        scope: 'openid email profile'
+        callbackURL: `${configService.get('EXTERNAL_URL')}/authn/okta_callback`,
```
em-c-rod, line 43: Was switching the `okta/callback` to `/okta_callback` stylistic, or did it change library use or functionality or something?

apps/backend/src/filters/authentication-exception.filter.ts
```diff
@@ -1,24 +1,46 @@
...
+    this.logger.warn(
+      `Authentication Error\n${JSON.stringify(errInfo, null, 2)}`
+    );
+    const authError =
+      `${_.get(request, 'authInfo.message')}\n${exception.message}`.trim();
+    response.cookie('authenticationError', authError, {
       secure: this.configService.isInProductionMode()
     });
-    response.redirect(301, '/');
+    response.redirect(302, '/');
```
em-c-rod, line 44: Do you mind explaining the switch from 301 to 302, so now we are saying there is a temporary redirect? Just wanting to understand.

apps/backend/src/main.ts
```diff
@@ -65,13 +65,17 @@ async function bootstrap() {
         }),
         proxy: configService.isInProductionMode() ? true : undefined,
         cookie: {
-          maxAge: 60 * 60,
+          maxAge: 60 * 60 * 1000,
```
em-c-rod, line 68: Is this unit milliseconds?

test/integration/login.cy.ts
```diff
@@ -70,6 +70,10 @@ context('Login', () => {
         cy.login(BAD_LOGIN_AUTHENTICATION);
         toastVerifier.toastTextContains('Incorrect Username or Password');
       });
+      it('fails to authenticate an oidc user with unverified email', () => {
```
em-c-rod, line 73: Good catch to add this test. Is there a way to add a test to catch the double login issue? Do any of them already capture that?

apps/backend/src/authn/authn.controller.ts
```diff
@@ -104,26 +104,31 @@ export class AuthnController {
   async loginToOkta(
     @Req() req: Request
   ): Promise<{userID: string; accessToken: string}> {
+    console.log('in the okta login func');
+    console.log(JSON.stringify(req.session, null, 2));
     return this.authnService.login(req.user as User);
   }
 
-  @Get('okta/callback')
+  @Get('okta_callback')
```
em-c-rod, line 112: I though the path for the get (`okta_callback`) was related to the path for the AuthGuard (`okta`) but George told me it isn't. Mostly just noting this here for understanding purposes.

em-c-rod (commented):
Left some comments and questions. Nice improvements here. Is there a way to confirm functionality on the Heroku deployment with the MITRE Okta integration?

Amndeep7:
> Left some comments and questions. Nice improvements here. Is there a way to confirm functionality on the Heroku deployment with the MITRE Okta integration?

@em-c-rod yeah i set our heimdall prod to be this branch 😅 so you can test it right there by going to heimdall-demo.mitre.org

apps/backend/package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@casl/ability": "^6.0.0",
+    "@govtechsg/passport-openidconnect": "https://github.com/Amndeep7/passport-openidconnect#logging",
```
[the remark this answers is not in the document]
Amndeep7: > Is this going to be back on the upstream of this library now that you have debugged the issue with your logging branch?
yes - it will be the actual govtechsg one
> Are you pushing anything you changed upstream for this library?
no - all i did was add a bunch of console logs in there

Amndeep7:
I transitioned to using @govtechsg/passport-openidconnect instead of passport-openidconnect since it's a fork that seems to be more actively maintained and willing to merge in open PRs.  They've also improved the documentation.  Immediately after using this library we started getting more accurate errors as to what was going wrong, which helped with the debugging process.

apps/backend/package.json
```diff
@@ -95,7 +97,6 @@
     "passport-ldapauth": "^3.0.1",
     "passport-local": "^1.0.0",
     "passport-oauth2": "^1.5.0",
-    "passport-openidconnect": "^0.1.1",
```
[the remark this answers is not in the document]
Amndeep7: Done: https://github.com/mitre/heimdall2/pull/6820#issuecomment-2895992381

apps/backend/src/authn/oidc.strategy.ts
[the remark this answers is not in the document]
Amndeep7: Yes I've introduced several new envvars.  I will be documenting them and getting them added to the helm chart when this PR is approved.

apps/backend/src/authn/oidc.strategy.ts
```diff
@@ -36,14 +37,26 @@ export class OidcStrategy extends PassportStrategy(Strategy, 'oidc') {
...
+              : undefined,
+        scope: ['openid', 'email', 'profile'],
+        skipUserProfile: false,
+        proxy:
+          configService.get('OIDC_USE_HTTPS_PROXY') === 'true'
+            ? true
+            : undefined,
+        agent:
+          configService.get('OIDC_USE_HTTPS_PROXY') === 'true'
+            ? new HttpsProxyAgent(configService.get('HTTPS_PROXY') ?? '')
```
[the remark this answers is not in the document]
Amndeep7: https://xxproxy.organization.com:port

apps/backend/src/authn/oidc.strategy.ts
Amndeep7: > What does plain vs. S256 mean?
I don't have an in-depth understanding but my skim makes me think that s256 is sha256 whereas plain is unencrypted.  We had a request for PKCE support from one of the linked issues so I figured it'd be straightforward enough to add since the library we're using exposes the functionality with that attribute.  
> Do you want to make a variable which required the user to specify one of the options instead of several true or false vars? 
The attribute takes in a string enum of 'S256' or 'plain' so I just expose that set of enums via multiple envvars instead of making the user do something like "OIDC_PKCE=S256|plain" themselves which is liable for typos/unallowed strings from them.  I'm not tied to this implementation decision, but figured it'd be easier for the end user.
> What about if both are set to true? 
If both are true, then we use S256 as per how I implemented the ternary operator chain
> Are they mutually exclusive?
Yes.  The documentation I will add will hopefully make this clear to the user.
> How can we make it fool proof for the user?
I thought that this would be the more fool proof solution but based off of your questions that might not be the case.  Do you have an alternative approach in mind?

apps/backend/src/authn/oidc.strategy.ts
Amndeep7: the type of the pkce attribute is `pkce?: 'S256' | 'plain'`.  setting it explicitly to undefined makes it so that we are disabling pkce support.

apps/backend/src/authn/oidc.strategy.ts
Amndeep7: Idk just personal preference to only set optional parameters to things when I am using them.  in this case, i'd only be using this parameter if we set it to true so i let it do its default behavior otherwise.

apps/backend/src/authn/okta.strategy.ts
Amndeep7: Please see PM for answers to the first two questions.

> And is that path determined on the Heimdall side rather than related to the Okta client that Heimdall would be integrated with?

No, this is an Okta configuration thing that Heimdall needs to respect.

In any case, this leads to the problem of what is a good default location.  IMO having a good enough default (which is one without a tenant defined OR having the tenant be 'default' but I chose the former cause see PM) and then the user can manually override however they want is the easiest solution.  Alternatively we can leave it as 'okta_domain' and then try to do some magic by inserting some 'okta_tenant' envvar into the url in the appropriate location if it is supplied, but what if the url is just completely different for some reason?  

On the whole, my preference would be to remove 'OKTA_DOMAIN' entirely and not try to be smart about default values at all and force a user to fill out an OIDC-like set of envvars based off of what they see at the .well-known url.  Primary concern is unnecessary breaking changes but like it wouldn't be that bad of a break...

apps/backend/src/authn/okta.strategy.ts
Amndeep7: No, the paths come purely from okta; however, trying to be smart about what the url will be is not great for our own MITRE deployment and was causing problems in the sponsor environment.

apps/backend/src/authn/okta.strategy.ts
Amndeep7: It seems like Okta might have some weird bug in the resolution of the callback path where it tries to go to /okta instead of /okta/callback for some reason iirc.  Workaround is just to flatten the path.

See: https://github.com/okta/okta-oidc-middleware/issues/20

apps/backend/src/filters/authentication-exception.filter.ts
Amndeep7: It came up in discussion with the sponsor team how 302 made a bit more sense instead of 301 since you aren't permanently going to go to the '/' page based off of what could be a temporary auth problem.

apps/backend/src/main.ts
Amndeep7: As it turns out, yes it is in ms instead of s.

test/integration/login.cy.ts
Amndeep7: The double login should be entirely resolved now that the /okta/callback -> /okta_callback + /oidc/callback -> /oidc_callback change happened.  In my manual testing, I have never seen it show up again esp after removing the section in the authexceptionfilter where it would automatically try to do the repeated login but use the wrong url since it was only kinda written correctly for oidc.

apps/backend/src/authn/authn.controller.ts
Amndeep7: They are entirely independent urls.  The /okta url's more or less sole purpose is to run into the guard that says you gotta log in via okta to hit it and replaces our implementation for that endpoint with the stuff that kicks off the auth process.  The /okta_callback url is what we supply to okta as a callback back into our application, has one last check to confirm that we're actually logged in w/ the state appropriately set and all, and then we can go off and create our user within the app.

apps/backend/package.json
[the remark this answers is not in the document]
em-c-rod: Cool 👍

apps/backend/src/authn/oidc.strategy.ts
em-c-rod: "I thought that this would be the more fool proof solution but based off of your questions that might not be the case. Do you have an alternative approach in mind?" I totally get what you mean from trying to mitigate user error. From a user understanding perspective, maybe you only need one var set to true | false and leave the other option as the default? Like have `OIDC_USES_PKCE_S256` and if this is not set or set to false, use the plain?

apps/backend/src/authn/okta.strategy.ts
em-c-rod: So from our own docs on this, we say `Your callback URL should be https://<your.heimdall.instance>/authn/okta/callback`. Does that mean, for example, a new release of this code would require us to update existing okta integrations accordingly?

apps/backend/src/main.ts
em-c-rod: Do you want to add a comment or, if the variable name is defined by us, add maxAgeInMsec?

test/integration/login.cy.ts
em-c-rod: Amazing

apps/backend/src/authn/okta.strategy.ts
Amndeep7: Yes, this will be a breaking change for our OKTA and OIDC users.  We will have to update the minor version number on our next release and state that this change is coming.

apps/backend/src/main.ts
Amndeep7: I can add a comment.  The parameter is not defined by us.

apps/backend/src/authn/okta.strategy.ts
em-c-rod: Ok. I have noticed it to _seem_ pretty consistent, which could put less work on the user, but that's just my opinion.

apps/backend/src/authn/okta.strategy.ts
em-c-rod: Okay, very important to know.

apps/backend/src/authn/okta.strategy.ts
Amndeep7: Nominally the user has been told in the docs that OKTA_DOMAIN is a required envvar if they're using this login method so we technically don't need to do anything tho I agree it's not the best experience.  We should probably have some sorta configuration vetting process that happens on spinup to ensure that all vars are set properly, but I think that'd be outside of the scope of this pr.

PR #7059, opened 2025-07-12
Use flag to force tenable to run on the frontend

georgedias:
- [x] Better error notification - inform user when Tenable connection failed due to DNS error (ENOTFOUND - address cannot be resolved)
- To solve the DNS Tenable host address resolution (hosting platform) implement one of these possible solutions:
   - [x] Flag to force tenable to run on the frontend
     - [x] Add flag to the [Helm Chart Values](https://github.com/mitre/heimdall2-helm/blob/main/heimdall2/values.yaml)
   - [ ] SSH tunneling from Heroku


Amndeep7:
Instead of this being a flag why not just automatically try doing it from the frontend and if it errors out for any reason then it goes and tries the backend and if that also fails to give the user some error message?

PR #7085, opened 2025-07-22
github workflow that builds the heimdall container and saves it as an artifact for every PR

Amndeep7:


Amndeep7:
image

note the tag matches this branch's name

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
+name: Build the Heimdall container and save it as an artifact in every PR
```
dhessmtr, line 1: Start review

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
```
dhessmtr, line 3: [master] branch - This PR is just helping us get one of the parts that the folks use. Eventually they will be able to just use the master branch for heimdall2.

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  docker:
+    runs-on: ubuntu-22.04
```
dhessmtr, line 12: ubuntu will be the base image.

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - name: Convert branch name into a safe docker tag by replacing invalid characters with underscores
```
dhessmtr, line 14: Regex for docker tags because it has less characters that it can use. E.g. "/" will not work so it'd have to be convered to "_" and ensure the starting character is "c" if it's a symbol

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+                      if (c ~ /[A-Za-z0-9_.-]/) {
+                          out = out c
+                      } else {
+                          out = out "_"
+                      }
+                  }
+                  print out
+              }'
+          )" >> $GITHUB_ENV
+      - name: Set up Docker Buildx
```
dhessmtr, line 39: "uses" - The docker plugins or packages required to help create the heimdall.tar ball?

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+          echo "DOCKER_TAG=$(
+            echo $GITHUB_HEAD_REF | awk '
+              {
+                  c = substr($0, 1, 1)
+                  if (c ~ /[A-Za-z0-9_]/) {
+                      out = c
+                  } else {
+                      out = "_"
+                  }
+                  for (i = 2; i <= length($0) && i <= 128; i++) {
```
NickShumar, line 28: Good catch on starting from the second character here, wouldn't have known tags can't start with a period, had to consult the docs

.github/workflows/build-save-docker.yml
```diff
@@ -0,0 +1,58 @@
...
+          context: .
+          push: false
+          platforms: linux/amd64
+          tags: mitre/heimdall2:${{ env.DOCKER_TAG }}
+          outputs: type=docker,dest=${{ runner.temp }}/heimdall.tar
+      - name: Upload docker tarball
+        uses: actions/upload-artifact@v4
+        with:
+          name: heimdall.tar
+          path: ${{ runner.temp }}/heimdall.tar
```
wdower, line 58: I see two files in the artifact run:
image

What's the `.dockerbuild` file? Do we need it? how is it added?

.github/workflows/build-save-docker.yml
Amndeep7: Yeah the intent of this is to allow folks to temporarily use functionality that are locked away in branches/PRs without having to build the container themselves.  Once the PR is merged in, then they'll use the actual release.

.github/workflows/build-save-docker.yml
Amndeep7: No, this is the designation for the runner.

.github/workflows/build-save-docker.yml
Amndeep7: 'c' is just the variable name and is shorthand for 'character'.  we assign that character to our output if it's a valid character otherwise we replace it with an underscore.

.github/workflows/build-save-docker.yml
Amndeep7: 'uses' is github actions/workflows syntax for specifying that we are using some 3rd party module, i.e. an 'action'.  in this case, we are setting up the docker binaries and configuration that is needed for our latter step that actually builds heimdall.

.github/workflows/build-save-docker.yml
Amndeep7: image

If you look slightly up, you'll see more details there.  It's something automatically generated by the docker build/push action to drive people to use Docker Desktop.  I don't think it's necessary.  I don't know how to configure it to be removed.  I don't think it's a problem to leave it be.

.github/workflows/build-save-docker.yml
Amndeep7: I didn't either which is why I referred to the docs and linked them as comments haha

.github/workflows/build-save-docker.yml
wdower: Ah. So I have to RTFM. Can do!

wdower (approved):

Issue #7337, opened 2025-10-23
[BUG] Cannot load or export CKL files with multiple Host IP addresses


em-c-rod:
**Describe the bug**
Cannot load or export a CKL file with 2 host IP addresses.

**To Reproduce**
Importing:
1. Load a CKL file with 2 Host IP addresses into Heimdall
2. Receive error "Error: Invalid checklist metadata fields: Host IP (123.123.1.123, 123.123.2.123)"

Exporting:
1. Try to export a file to DISA Checklist
2. Try to include the following Host IP addresses: "123.123.1.123, 123.123.2.123"
3. The field is red and does not allow the export.

**Expected behavior**
Expect to be able to load and export CKL files with multiple Host IP addresses.

**Desktop (please complete the following information):**
 - iOS
 - Chrome
 -  141.0.7390.77

**Additional context**
I cannot upload a CKL file, but you can recreate by exporting a sample file to CKL from Heimdall, then opening in a text editor and adding an additional (so there are more than one) Host IP addresses in the XML.

https://github.com/mitre/heimdall2/pull/7344

em-c-rod:
Reference this PR that fixed the same problem for multiple Host MAC addresses:
https://github.com/mitre/heimdall2/pull/6679

PR #7344, opened 2025-10-24
Update Checklist Mapper Import to Allow Multiple IP Addresses

DMedina6:
Resolves the input part of https://github.com/mitre/heimdall2/issues/7337

refactor checklist mappper test
add test for file with multiple IP addresses
update checklist_metadata_utils to be able to validate a host IP address string with multiple IP addresses
allows export with more than one IP address

### DIFF
--- a/libs/hdf-converters/package.json
+++ b/libs/hdf-converters/package.json
@@ -25,7 +25,7 @@
     "xml2json": "tsx data/converters/xml2json.ts"
   },
   "dependencies": {
-    "@aws-sdk/client-config-service": "3.917.0",
+    "@aws-sdk/client-config-service": "3.919.0",
     "@e965/xlsx": "^0.20.0",
     "@mdi/js": "^7.0.96",
     "@microsoft/microsoft-graph-types": "^2.40.0",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "typescript": "^4.9.3"
   },
   "devDependencies": {
-    "dotenv-cli": "^10.0.0"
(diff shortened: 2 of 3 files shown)


DMedina6:
adding copilot as a reviewer just to see what it does

DMedina6:
One issue: 

image

It wants me to use replaceAll() instead of replace(), but this requires eslint2021, and we use 2019 in the tsconfig

image

libs/hdf-converters/test/mappers/forward/checklist_mapper.spec.ts
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs';
+import fs from 'node:fs';
```
Amndeep7: I think we usually prefer to not put the 'node' preface in front of these libraries

Amndeep7:
> One issue:
> image
> 
> It wants me to use replaceAll() instead of replace(), but this requires eslint2021, and we use 2019 in the tsconfig
> image

Yeah life is true suffering with all the tech debt we've accrued.  I would try doing the bump and if anything breaks too egregiously then we can revert and ignore the sonarqube error since it's just a code smell not a vuln or anything.

libs/hdf-converters/test/mappers/forward/checklist_mapper.spec.ts
DMedina6: Oh okay. I only changed it because Sonarqube had an issue with it, although I'm not sure why it decided now it had a problem since I never changed it before this.

PR #7362, opened 2025-10-28
add dialogue box for a tree map legend

em-c-rod:
Information on the control families was collected from the following sources:
https://nist-sp-800-53-r5.bsafes.com/docs/chapter-three-the-controls/
https://csrc.nist.gov/CSRC/media/Publications/sp/800-53/rev-5/final/documents/sp800-53r4-appj-to-r5-comparison.xlsx
And Eugene

### DIFF
--- a/libs/hdf-converters/package.json
+++ b/libs/hdf-converters/package.json
@@ -25,7 +25,7 @@
     "xml2json": "tsx data/converters/xml2json.ts"
   },
   "dependencies": {
-    "@aws-sdk/client-config-service": "3.920.0",
+    "@aws-sdk/client-config-service": "3.921.0",
     "@e965/xlsx": "^0.20.0",
     "@mdi/js": "^7.0.96",
     "@microsoft/microsoft-graph-types": "^2.40.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -252,534 +252,534 @@
     "@smithy/util-utf8" "^2.0.0"
     tslib "^2.6.2"
 
-"@aws-sdk/client-config-service@3.920.0":
-  version "3.920.0"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-config-service/-/client-config-service-3.920.0.tgz#02cfe5d911359c100e8848b99b6d4928816d9291"
-  integrity sha512-0vKds9ZlLSQS8JQMJpa4KWojb/U6bcL5aPXUAoT2lhHSyORyquwUjxy6+Sqe3vxo//Y5YKa8EMl5nSQ1J9Xriw==
+"@aws-sdk/client-config-service@3.921.0":
+  version "3.921.0"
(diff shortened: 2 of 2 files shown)


DMedina6:
This all looks good to me!

apps/frontend/src/views/Results.vue
```diff
@@ -147,7 +147,68 @@
...
+                <v-spacer />
+                <v-btn class="mr-2" icon small @click="isActiveDialog = true">
+                  <v-icon
+                    b-tooltip.hover
+                    title="Tree Map Legend"
+                    color="primary"
+                  >
+                    mdi-information-outline
+                  </v-icon>
+                </v-btn>
```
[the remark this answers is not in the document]
DMedina6: disagree on the basis that this is consistent with how the rest of the app is already doing it

apps/frontend/src/views/Results.vue
```diff
@@ -147,7 +147,68 @@
...
+                <v-dialog v-model="isActiveDialog" width="500">
+                  <v-card>
+                    <v-card-title>Tree Map Legend</v-card-title>
+                    <v-card-text class="text-h7">
+                      The NIST 800-53 Control Families have the following
+                      acronyms:
+                      <br /><br />
```
Amndeep7, line 168: It seems like it should be possible to use CSS to handle linebreaks in multiline strings instead of using the `br` tag according to genai.  Can we try doing it with CSS instead?

See AI block below:

In Vuetify 2, you can use the white-space: pre-line; CSS property to display multiline strings inside a v-card-text block without using <br>. This property ensures that line breaks in the string (e.g., \n) are respected.

Here’s an example:
```
<template>
  <v-card>
    <v-card-text class="multiline-text">
      This is line one.
      This is line two.
      This is line three.
    </v-card-text>
  </v-card>
</template>

<style scoped>
.multiline-text {
  white-space: pre-line;
}
</style>
```
In this example:
The white-space: pre-line; CSS property ensures that any line breaks in the text are rendered as new lines.
You can write your multiline string directly in the template, or bind it dynamically using Vue's data or computed properties.

apps/frontend/src/views/Results.vue
```diff
@@ -147,7 +147,68 @@
...
+                    title="Tree Map Legend"
+                    color="primary"
+                  >
+                    mdi-information-outline
+                  </v-icon>
+                </v-btn>
+                <v-dialog v-model="isActiveDialog" width="500">
+                  <v-card>
+                    <v-card-title>Tree Map Legend</v-card-title>
+                    <v-card-text class="text-h7">
```
Amndeep7, line 165: If the css mechanism works, then I also want us to put the text on the js side that is then templated in as opposed to hard coding into the html.

Amndeep7 (dismissed):

apps/frontend/src/views/Results.vue
DMedina6: Emily tried this, but it didn't look as expected in the local build. She's going to look into this further, but in the mean time we want to be able to integrate this and release this as soon as possible. Since it works currently, I've made a [separate issue](https://github.com/mitre/heimdall2/issues/7380) for this improvement to the PR, and we're going merge this for now.

PR #7548, opened 2026-01-04
Multi platform build

### LINKED ISSUE #7547: Mac MX-compatible container build
Right now the Docker build of Heimdall is done on platform linux/amd64 only, which makes it unsupported on Mac chips.

A small modification to the build workflow could produce another container image for linux/arm64.

wdower:
Closes #7547 

Also uses an updated Docker build pattern by using [Docker Bake](https://docs.docker.com/build/bake/) and a declarative HCL file to define images to build for simplicity.

Also adds a manual dispatch option to use the Docker Build Cloud driver to do the build which is helpful because the ARM64 build can take a solid hour to run. We don't really need fast builds for things like Dependabot updates and we only have so many free build minutes, but it's good to have the option.

I've tested the pattern using push-server-to-docker.yml to build and push AMD and ARM builds, tagged multi-platform-build on DockerHub. I've now fixed the workflow to only trigger on master as usual, with the addition of a workflow_dispatch trigger (that we can use to run the fast build on Docker Build Cloud).

### DIFF
--- a/libs/hdf-converters/package.json
+++ b/libs/hdf-converters/package.json
@@ -29,7 +29,7 @@
     "xml2json": "tsx data/converters/xml2json.ts"
   },
   "dependencies": {
-    "@aws-sdk/client-config-service": "3.988.0",
+    "@aws-sdk/client-config-service": "3.990.0",
     "@e965/xlsx": "^0.20.0",
     "@mdi/js": "^7.0.96",
     "@microsoft/microsoft-graph-types": "^2.40.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -252,24 +252,24 @@
     "@smithy/util-utf8" "^2.0.0"
     tslib "^2.6.2"
 
-"@aws-sdk/client-config-service@3.988.0":
-  version "3.988.0"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-config-service/-/client-config-service-3.988.0.tgz#f30fa9b76b5de63bf200b7ce51fd2fd8c41098d2"
-  integrity sha512-C7Rox0shg4TVGMa6pPPijwqrX872GhWWOV7mFsQwxg27yXIp6eB0IE22NAc5SgUFLrpKEbfzoBnv3Ht10yCWpw==
+"@aws-sdk/client-config-service@3.990.0":
+  version "3.990.0"
+  resolved "https://registry.yarnpkg.com/@aws-sdk/client-config-service/-/client-config-service-3.990.0.tgz#444d2de498b80b8069e8b50131fdcaeb2ef7b291"
+  integrity sha512-FK4apXgbODH/6qdQpA7Uz4IN+crVE6A3HUolpvUZqOKXTzYTxE4VvkWCmVonoMri7TOwBbPEjCtZz++HkKlHFw==
   dependencies:
     "@aws-crypto/sha256-browser" "5.2.0"
     "@aws-crypto/sha256-js" "5.2.0"
-    "@aws-sdk/core" "^3.973.8"
-    "@aws-sdk/credential-provider-node" "^3.972.7"
+    "@aws-sdk/core" "^3.973.10"
+    "@aws-sdk/credential-provider-node" "^3.972.9"
     "@aws-sdk/middleware-host-header" "^3.972.3"
     "@aws-sdk/middleware-logger" "^3.972.3"
     "@aws-sdk/middleware-recursion-detection" "^3.972.3"
-    "@aws-sdk/middleware-user-agent" "^3.972.8"
+    "@aws-sdk/middleware-user-agent" "^3.972.10"
     "@aws-sdk/region-config-resolver" "^3.972.3"
     "@aws-sdk/types" "^3.973.1"
-    "@aws-sdk/util-endpoints" "3.988.0"
+    "@aws-sdk/util-endpoints" "3.990.0"
     "@aws-sdk/util-user-agent-browser" "^3.972.3"
-    "@aws-sdk/util-user-agent-node" "^3.972.6"
+    "@aws-sdk/util-user-agent-node" "^3.972.8"
     "@smithy/config-resolver" "^4.4.6"
     "@smithy/core" "^3.23.0"
     "@smithy/fetch-http-handler" "^5.3.9"
@@ -298,31 +298,31 @@
     tslib "^2.6.2"
 
 "@aws-sdk/client-s3@^3.425.0":
-  version "3.988.0"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-s3/-/client-s3-3.988.0.tgz#f445d820ad59cbedce98e80458ab641cba023cbc"
-  integrity sha512-mt7AdkieJJ5hEKeCxH4sdTTd679shUjo/cUvNY0fUHgQIPZa1jRuekTXnRytRrEwdrZWJDx56n1S8ism2uX7jg==
+  version "3.990.0"
+  resolved "https://registry.yarnpkg.com/@aws-sdk/client-s3/-/client-s3-3.990.0.tgz#5cafb93eba5e8fd26e16a6f7fc27776a85d08878"
+  integrity sha512-XnsM8RgB35Atn2+aYSocitCybDG82x9yYf/s2D23ytpyHCupmuZN3LzK2a0WxmKO6Zf7EtEIYy0mHGY4tLp9YA==
   dependencies:
     "@aws-crypto/sha1-browser" "5.2.0"
     "@aws-crypto/sha256-browser" "5.2.0"
     "@aws-crypto/sha256-js" "5.2.0"
-    "@aws-sdk/core" "^3.973.8"
-    "@aws-sdk/credential-provider-node" "^3.972.7"
+    "@aws-sdk/core" "^3.973.10"
+    "@aws-sdk/credential-provider-node" "^3.972.9"
     "@aws-sdk/middleware-bucket-endpoint" "^3.972.3"
     "@aws-sdk/middleware-expect-continue" "^3.972.3"
-    "@aws-sdk/middleware-flexible-checksums" "^3.972.6"
+    "@aws-sdk/middleware-flexible-checksums" "^3.972.8"
     "@aws-sdk/middleware-host-header" "^3.972.3"
     "@aws-sdk/middleware-location-constraint" "^3.972.3"
     "@aws-sdk/middleware-logger" "^3.972.3"
     "@aws-sdk/middleware-recursion-detection" "^3.972.3"
-    "@aws-sdk/middleware-sdk-s3" "^3.972.8"
+    "@aws-sdk/middleware-sdk-s3" "^3.972.10"
     "@aws-sdk/middleware-ssec" "^3.972.3"
-    "@aws-sdk/middleware-user-agent" "^3.972.8"
+    "@aws-sdk/middleware-user-agent" "^3.972.10"
     "@aws-sdk/region-config-resolver" "^3.972.3"
-    "@aws-sdk/signature-v4-multi-region" "3.988.0"
+    "@aws-sdk/signature-v4-multi-region" "3.990.0"
     "@aws-sdk/types" "^3.973.1"
-    "@aws-sdk/util-endpoints" "3.988.0"
+    "@aws-sdk/util-endpoints" "3.990.0"
     "@aws-sdk/util-user-agent-browser" "^3.972.3"
-    "@aws-sdk/util-user-agent-node" "^3.972.6"
+    "@aws-sdk/util-user-agent-node" "^3.972.8"
     "@smithy/config-resolver" "^4.4.6"
     "@smithy/core" "^3.23.0"
     "@smithy/eventstream-serde-browser" "^4.2.8"
@@ -358,23 +358,23 @@
     "@smithy/util-waiter" "^4.2.8"
     tslib "^2.6.2"
 
-"@aws-sdk/client-sso@3.988.0":
-  version "3.988.0"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-sso/-/client-sso-3.988.0.tgz#0b6d5673e1b69d6cce03c78c11c4f705ea8fb303"
-  integrity sha512-ThqQ7aF1k0Zz4yJRwegHw+T1rM3a7ZPvvEUSEdvn5Z8zTeWgJAbtqW/6ejPsMLmFOlHgNcwDQN/e69OvtEOoIQ==
+"@aws-sdk/client-sso@3.990.0":
+  version "3.990.0"
+  resolved "https://registry.yarnpkg.com/@aws-sdk/client-sso/-/client-sso-3.990.0.tgz#a5471e810b848f740e5c6a8abe90eb2304670c8b"
+  integrity sha512-xTEaPjZwOqVjGbLOP7qzwbdOWJOo1ne2mUhTZwEBBkPvNk4aXB/vcYwWwrjoSWUqtit4+GDbO75ePc/S6TUJYQ==
   dependencies:
     "@aws-crypto/sha256-browser" "5.2.0"
     "@aws-crypto/sha256-js" "5.2.0"
-    "@aws-sdk/core" "^3.973.8"
+    "@aws-sdk/core" "^3.973.10"
     "@aws-sdk/middleware-host-header" "^3.972.3"
     "@aws-sdk/middleware-logger" "^3.972.3"
     "@aws-sdk/middleware-recursion-detection" "^3.972.3"
-    "@aws-sdk/middleware-user-agent" "^3.972.8"
+    "@aws-sdk/middleware-user-agent" "^3.972.10"
     "@aws-sdk/region-config-resolver" "^3.972.3"
     "@aws-sdk/types" "^3.973.1"
-    "@aws-sdk/util-endpoints" "3.988.0"
+    "@aws-sdk/util-endpoints" "3.990.0"
     "@aws-sdk/util-user-agent-browser" "^3.972.3"
-    "@aws-sdk/util-user-agent-node" "^3.972.6"
+    "@aws-sdk/util-user-agent-node" "^3.972.8"
     "@smithy/config-resolver" "^4.4.6"
     "@smithy/core" "^3.23.0"
     "@smithy/fetch-http-handler" "^5.3.9"
@@ -535,23 +535,23 @@
     tslib "^2.6.2"
 
 "@aws-sdk/client-sts@^3.427.0":
-  version "3.988.0"
-  resolved "https://registry.yarnpkg.com/@aws-sdk/client-sts/-/client-sts-3.988.0.tgz#767ee88e89caf4d15580e57416ff5248fdb02a40"
-  integrity sha512-VgHC1UgtxR+CSBLk3dj3jIku3NML8HDALU6JvfNVlokkYfk638jZUCNWbvo2XOUy+/qaUYnXrxxTwtIcWqTnNg==
+  version "3.990.0"
+  resolved "https://registry.yarnpkg.com/@aws-sdk/client-sts/-/client-sts-3.990.0.tgz#19942092345fe056a736d0cde7d98e39ee64ba7c"
+  integrity sha512-w1MfhIKV1GmQJI2gN64gvtwSPJR55n5TIkujIoZZy3V3OlksAOWCnfUBHoyfu8x4aZn70cINfa+ixPtUHtnEew==
   dependencies:
     "@aws-crypto/sha256-browser" "5.2.0"
     "@aws-crypto/sha256-js" "5.2.0"
-    "@aws-sdk/core" "^3.973.8"
-    "@aws-sdk/credential-provider-node" "^3.972.7"
+    "@aws-sdk/core" "^3.973.10"
+    "@aws-sdk/credential-provider-node" "^3.972.9"
     "@aws-sdk/middleware-host-header" "^3.972.3"
     "@aws-sdk/middleware-logger" "^3.972.3"
     "@aws-sdk/middleware-recursion-detection" "^3.972.3"
-    "@aws-sdk/middleware-user-agent" "^3.972.8"
+    "@aws-sdk/middleware-user-agent" "^3.972.10"
     "@aws-sdk/region-config-resolver" "^3.972.3"
     "@aws-sdk/types" "^3.973.1"
-    "@aws-sdk/util-endpoints" "3.988.0"
+    "@aws-sdk/util-endpoints" "3.990.0"
     "@aws-sdk/util-user-agent-browser" "^3.972.3"
-    "@aws-sdk/util-user-agent-node" "^3.972.6"
+    "@aws-sdk/util-user-agent-node" "^3.972.8"
     "@smithy/config-resolver" "^4.4.6"
     "@smithy/core" "^3.23.0"
     "@smithy/fetch-http-handler" "^5.3.9"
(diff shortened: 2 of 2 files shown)


.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
 on:
   push:
     branches: [master]
+  workflow_dispatch:
+    inputs:
+      use_build_cloud:
+        description: 'Use Docker Build Cloud (fast, uses minutes)'
```
Amndeep7, line 9: Ideally there is an API call that returns the number of minutes remaining (so something aside from looking at the GUI) so that we can have some safeguarding logic to prevent us from exceeding the number of minutes we have left based off of the average duration of a heimdall build

.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
...
+        default: false
 
 permissions:
   contents: read
 
 jobs:
   docker:
-    runs-on: ubuntu-24.04
+    name: Build and Push to Docker Hub
+    runs-on: ubuntu-latest
```
Amndeep7: I prefer having explicit versions instead of 'latest' so stuff doesn't break underneath us

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+}
+
+# Heimdall Server (full) - Local platform only
+target "server" {
+  inherits = ["_common"]
+  tags = [
+    "${TAG_PREFIX}:latest",
+    "${TAG_PREFIX}:server"
+  ]
+  platforms = ["${LOCAL_PLATFORM}"]
```
Amndeep7: instead of local platform, please make separate targets per arch and then have the multiarch one

.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
...
     steps:
       - name: Checkout the Heimdall Repository
         uses: actions/checkout@v5
-      - name: Set up QEMU
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ vars.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PAT }}
```
Amndeep7, line 30: Would be nice if we could fully transition off of the DOCKERHUB_USERNAME/DOCKERHUB_TOKEN combo across all of the repos to instead use this new credential.  At minimum the SAF CLI and Vulcan.

.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
...
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - name: Build and push the container image to Dockerhub
-        id: docker_build
-        uses: docker/build-push-action@v6
+          driver: cloud
+          endpoint: "mitre/mitre-builder"
+
+      - name: Set up Docker Buildx (Standard - Free)
```
Amndeep7: "standard-free" makes me think that it's still reaching out to the internet as opposed to being local.  can we just call this step variant 'local' and the other one 'docker cloud'

.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
...
+        if: github.event.inputs.use_build_cloud != 'true'
```
Amndeep7: can we have both this above the cloud step so that the order is qemu, buildx (local), buildx (docker cloud) in order to have both the local ones together

.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push multi-platform lite image
+        uses: docker/bake-action@v5
+        env:
+          TAG_SUFFIX: ${{ github.sha }}
```
Amndeep7: do not forget to re-add the latest tag

.github/workflows/push-lite-to-docker.yml
```diff
@@ -3,35 +3,57 @@ name: Push Heimdall Lite to Docker Hub on every merge to master and tag as lates
...
         with:
-          context: .
-          file: Dockerfile.lite
+          files: docker-bake.hcl
+          targets: lite-ci
           push: true
-          platforms: linux/amd64
-          tags: mitre/heimdall-lite:latest,mitre/heimdall-lite:${{ github.sha }}
-      - name: Get Docker SHA since the Iron Bank release requires us to specify the exact resources we need them to pull into the environment
```
Amndeep7, line 34: prefer reverting the comment

DOCKER-BAKE.md
Amndeep7, line 1: Can we move this into the wiki instead

DOCKER-BAKE.md
```diff
@@ -0,0 +1,228 @@
+# Building Heimdall Images Locally with Docker Bake
+
+This guide shows developers how to **build** Heimdall Server and Heimdall Lite Docker images from source using Docker Bake.
+
+> **Note:** This is for developers who want to build custom images or test changes. If you just want to **run** Heimdall using pre-built images, see the [README's Docker setup instructions](README.md#heimdall-server---docker) and use `./setup-docker-env.sh` instead.
+
+
+## Prerequisites
+
+1. **Docker Desktop** (macOS) or **Docker Engine** (Linux)
```
Amndeep7: why is docker desktop explicitly labeled as macos, wouldn't it also work for windows?  and nominally we should have windows instructions as well

DOCKER-BAKE.md
```diff
@@ -0,0 +1,228 @@
...
+| `lite` | Frontend-only static site | Your machine only | Local development |
+| `server-multiarch` | Full Heimdall | amd64 + arm64 | Testing multi-platform |
+| `lite-multiarch` | Frontend-only | amd64 + arm64 | Testing multi-platform |
+| `server-dev` | Development build | Your machine only | Development with NODE_ENV=development |
+| `server-ci` | CI build with caching | amd64 + arm64 | Same as CI |
+| `lite-ci` | CI lite with caching | amd64 + arm64 | Same as CI |
+
+## Common Development Workflows
+
+### 1. Quick Local Test (Fastest)
```
Amndeep7: rename these headers to be more exact, ex. this is just 'local build'.  the next one is 'local build - multi-platform', etc.  fastest or not is irrelevant, it's what the outcome is that needs to be most clearly presented

DOCKER-BAKE.md
```diff
@@ -0,0 +1,228 @@
...
+
+Build for your Mac's architecture only:
```
Amndeep7: like mentioned elsewhere i would like to have explicit arch designations at all times instead of being dependent on the state of the user's arch at runtime.

anyways these commands apply to unix more broadly not just macs

DOCKER-BAKE.md
```diff
@@ -0,0 +1,228 @@
...
+
+# Verify the image
+docker images | grep heimdall2
+```
+
+**Build time:** ~5-10 minutes (native build, no emulation)
+
+### 2. Test Multi-Platform Build (Before PR)
+
+Verify your changes work on both Intel and ARM:
```
Amndeep7: intel-> amd

DOCKER-BAKE.md
```diff
@@ -0,0 +1,228 @@
...
+# - lite-multiarch (amd64 + arm64)
+```
+
+### 5. Test the Exact CI Build
+
+Replicate what runs in GitHub Actions:
+
+```bash
+# Set the version tag like CI does
+export TAG_SUFFIX=$(git rev-parse HEAD)
```
Amndeep7: This should not be hidden away in the CI section.  Should be brought up earlier and specify that a) if you do not provide it then it just does 'latest', b) this example puts in the git sha of the commit, and c) you can change it to be whatever you want.

DOCKER-BAKE.md
```diff
@@ -0,0 +1,228 @@
...
+## Available Build Targets
+
+| Target | What It Builds | Platforms | Use Case |
+|--------|---------------|-----------|----------|
+| `server` | Full Heimdall (backend + frontend) | Your machine only | Local development |
+| `lite` | Frontend-only static site | Your machine only | Local development |
+| `server-multiarch` | Full Heimdall | amd64 + arm64 | Testing multi-platform |
+| `lite-multiarch` | Frontend-only | amd64 + arm64 | Testing multi-platform |
+| `server-dev` | Development build | Your machine only | Development with NODE_ENV=development |
+| `server-ci` | CI build with caching | amd64 + arm64 | Same as CI |
```
Amndeep7: can we get rid of caching?  for gha, I do not want to use caching since there's a chance that layers are not updated until we manually force clear the cache or update something about the layer which makes the cache irrelevant.  we've been burned by this before so it's just a lot easier to have no cache every time.  locally it's whatever and a lot easier to fix if you do need to clear out the cache.

.github/workflows/push-lite-to-docker.yml
Amndeep7, line 1: can we confirm that this all still works on IB side?

Amndeep7, line 1: can we add support for arm builds on IB (idk if it will require new repos over there for the different arch or if we can just combine them somehow, but it'll probably have to be separate repos)

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+
+variable "NODE_ENV" {
+  default = "production"
+}
+
+variable "YARNREPO_MIRROR" {
+  default = "https://registry.npmjs.org"
+}
+
+# Detect local platform (macOS needs linux override)
```
Amndeep7: as mentioned elsewhere, i do not want local platform stuff at all - specify the arch every time for the sake of clarity.  and once you've typed it in once, you can just scroll up / search your history for the bake command so it's not like specifying the arch is gonna be more onerous than a few extra characters the first time you do it

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+  dockerfile = "Dockerfile"
+  args = {
+    BASE_CONTAINER = "${BASE_CONTAINER}"
+    NODE_ENV = "${NODE_ENV}"
+    YARNREPO_MIRROR = "${YARNREPO_MIRROR}"
+  }
+  labels = {
+    "org.opencontainers.image.source" = "https://github.com/mitre/heimdall2"
+    "org.opencontainers.image.licenses" = "Apache-2.0"
+    "org.opencontainers.image.description" = "Heimdall2 - Security Automation Framework"
```
Amndeep7:
```suggestion
    "org.opencontainers.image.description" = "Heimdall - Security Automation Framework"
```

Amndeep7: Think we should slowly continue stripping out the '2' moniker everywhere.  If the repo name contains it, it's whatever, but we should stop labeling it as that ourselves.

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+  args = {
+    BUILD_CONTAINER = "${BASE_CONTAINER}"
+    BASE_CONTAINER = "nginx:alpine"
+    NODE_ENV = "${NODE_ENV}"
+    YARNREPO_MIRROR = "${YARNREPO_MIRROR}"
+  }
+  labels = {
+    "org.opencontainers.image.source" = "https://github.com/mitre/heimdall2"
+    "org.opencontainers.image.licenses" = "Apache-2.0"
+    "org.opencontainers.image.description" = "Heimdall Lite - Security Automation Framework (Frontend Only)"
```
Amndeep7:
```suggestion
    "org.opencontainers.image.description" = "Heimdall Lite - Security Automation Framework"
```

.github/workflows/release-lite-to-docker.yml
```diff
@@ -24,26 +31,40 @@ jobs:
...
         with:
-          context: .
-          file: Dockerfile.lite
+          files: docker-bake.hcl
+          targets: lite-ci
           push: true
-          platforms: linux/amd64
-          tags: mitre/heimdall-lite:release-latest,mitre/heimdall-lite:${{ steps.format-tag.outputs.replaced }}
+          set: |
+            lite-ci.tags=mitre/heimdall-lite:release-latest,mitre/heimdall-lite:${{ steps.format-tag.outputs.replaced }}
```
Amndeep7: why is this doing the 'set' paradigm whereas the pushes do the tag_suffix paradigm?  is there a way to make it consistent across the board?

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+group "all" {
+  targets = ["server-multiarch", "lite-multiarch"]
+}
+
+# Heimdall Server (full) - Local platform only
+target "server" {
+  inherits = ["_common"]
+  tags = [
+    "${TAG_PREFIX}:latest",
+    "${TAG_PREFIX}:server"
```
Amndeep7: why do we have a 'server' tag?  it's unnecessary imo.  if it's cause there's no way to distinguish server from lite due to them having the same tag_prefix, then we should have separate tag prefixes for them.

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+}
+
+# Heimdall Server (full) - Multi-architecture
+target "server-multiarch" {
+  inherits = ["_common"]
+  tags = concat(
+    [
+      "${DOCKER_HUB_REPO}:latest"
+    ],
+    TAG_SUFFIX != "" ? ["${DOCKER_HUB_REPO}:${TAG_SUFFIX}"] : []
```
Amndeep7: why do only the multiarch ones respect tag_suffix?

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+  )
+  platforms = ["linux/amd64", "linux/arm64"]
+}
+
+# Heimdall Lite (frontend-only) - Local platform only
+target "lite" {
+  inherits = ["_common_lite"]
+  tags = [
+    "${TAG_PREFIX}:lite",
+    "${TAG_PREFIX}:lite-latest"
```
Amndeep7: having the separate prefixes also gets rid of this lite-latest stuff so the user only has to keep an eye out for 'latest' period

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+
+# Development target with custom NODE_ENV
+target "server-dev" {
+  inherits = ["_common"]
+  args = {
+    NODE_ENV = "development"
+  }
+  tags = ["${TAG_PREFIX}:dev"]
+  platforms = ["${LOCAL_PLATFORM}"]
+}
```
Amndeep7: imo local/bare-metal dev is the only place where node_env should be development.  like i don't foresee any requirement for my docker containers to have the dev versions of my dependencies.  let's get rid of this target.

docker-bake.hcl
```diff
@@ -0,0 +1,143 @@
...
+  cache-from = ["type=gha,scope=server"]
+  cache-to = ["type=gha,mode=max,scope=server"]
+}
+
+# CI target for lite with GitHub Actions cache
+target "lite-ci" {
+  inherits = ["lite-multiarch"]
+  cache-from = ["type=gha,scope=lite"]
+  cache-to = ["type=gha,mode=max,scope=lite"]
+}
```
Amndeep7: As discussed and mentioned above, we've been burned by cached layers in docker containers before.  I would rather not deal with that problem again.  Please remove these targets as well.

Amndeep7:
@wdower seems like hdf converters randomly decided to break underneath you cause that test used to pass just fine.  I would've said that it seems like something with one of the dependency updates broke it but that frankly makes no sense cause attestations have nothing to do with the aws sdk which is what got changed.  regardless, it needs to be fixed whether in here or some other pr.  i won't have a chance to look at it for a little bit.

.github/workflows/push-lite-to-docker.yml
wdower: Can't find any API call that implements this.

docker-bake.hcl
wdower: done

.github/workflows/push-lite-to-docker.yml
wdower: done

.github/workflows/push-lite-to-docker.yml
wdower: done

.github/workflows/push-lite-to-docker.yml
wdower: created issues on those repos to track

.github/workflows/push-lite-to-docker.yml
wdower: done

.github/workflows/push-lite-to-docker.yml
wdower: done

DOCKER-BAKE.md
wdower: done

docker-bake.hcl
wdower: fixed

docker-bake.hcl
wdower: removed and refactored

docker-bake.hcl
wdower: removed

docker-bake.hcl
wdower: done

wdower:
@Amndeep7 Some more investigation of IB shows that if you want to have both amd and arm builds available in the registry, you need to drop off a specific build file on the P1 side. We can work that separately.

I can see from earlier testing that this workflow does, in fact, trigger a MR on P1 as expected, and that MR can pull the image we build with the Docker Bake process just fine -- it's just only getting the amd build right now since that's the only one we have configured it to pull.

I would argue that merging in this PR before we build out an arm dockerfile on the P1 side doesn't lose us anything. We don't currently support arm builds over there at all at the moment anyway.

wdower:
> @wdower seems like hdf converters randomly decided to break underneath you cause that test used to pass just fine. I would've said that it seems like something with one of the dependency updates broke it but that frankly makes no sense cause attestations have nothing to do with the aws sdk which is what got changed. regardless, it needs to be fixed whether in here or some other pr. i won't have a chance to look at it for a little bit.

Looks like one of the tests was failing because our example data struct for attestations was just out of date. The TTL for the example attestation had simply passed (so attestations functionality was ironically working as expected). I could fix that test. The other ones for Sonarqube and Splunk seem to be much more involved.

.beads/.gitignore
Amndeep7, line 1: i think these bead files should be removed

.github/workflows/push-lite-to-docker.yml
Amndeep7: @wdower said it wouldn't stop IB builds from happening: https://github.com/mitre/heimdall2/pull/7548#issuecomment-3795798314

Amndeep7: @wdower can you make an issue for this either here or on the ib repo and then put in any notes you had on the topic.  once that's done it can be marked as resolved.

.gitattributes
Amndeep7, line 1: this should also be removed alongside the beads stuff

.github/workflows/push-lite-to-docker.yml
```diff
@@ -29,28 +29,28 @@ jobs:
           username: ${{ vars.DOCKER_USER }}
           password: ${{ secrets.DOCKER_PAT }}
 
-      - name: Set up QEMU (for emulated builds)
```
Amndeep7: why was this comment removed?

it still applies in the situations where we're going to be building cross-platform images from within the ci here and not via docker build cloud

.github/workflows/push-lite-to-docker.yml
Amndeep7, line 1: can we update this workflow https://github.com/mitre/heimdall2/blob/master/.github/workflows/build-save-docker.yml to also use docker bake so that we're fully transitioned to using that technique however since it's gonna be regular ci it shouldn't do multiplatform builds only linux/amd64 unless a flag is set to do multiplatform.  it should also have the flag of doing it on docker cloud but that would only apply if the multiplatform one was also set to true.

docker-bake.hcl
```diff
@@ -59,33 +55,23 @@ target "_common_lite" {
...
+# Group: Build all variants (multi-arch)
 group "default" {
   targets = ["server", "lite"]
 }
 
-# Group: Build all with multi-arch
-group "all" {
-  targets = ["server-multiarch", "lite-multiarch"]
+# Group: Build single-arch variants for all platforms
+group "all-single-arch" {
```
Amndeep7: what is the purpose of this group?  feels like it's just 'default' but with more steps.

docker-bake.hcl
```diff
@@ -118,26 +118,27 @@ target "lite-multiarch" {
...
-    NODE_ENV = "development"
-  }
-  tags = ["${TAG_PREFIX}:dev"]
-  platforms = ["${LOCAL_PLATFORM}"]
+# Heimdall Lite (frontend-only) - AMD64 only
+target "lite-amd64" {
+  inherits = ["_common_lite"]
+  tags = concat(
+    [
+      "mitre/heimdall-lite:amd64"
```
Amndeep7: we should make another variable - DOCKER_HUB_REPO should be renamed to show that it is for heimdall and then another variable should be created that is for heimdall-lite

.github/workflows/push-lite-to-docker.yml
Amndeep7: i only see an issue on vulcan but not on the saf cli

.github/workflows/push-lite-to-docker.yml
Amndeep7: done

DOCKER-BAKE.md
Amndeep7, line 1: the docker bake stuff should be linked to from the TOC in the wiki

Amndeep7, line 1: should make it clearer in the advanced usecase setting that the envvars and inline setting are equivalent (which i think they are, but please confirm so)

Amndeep7, line 1: for the one-off build cloud one, you presumably would still need to create a cloud builder first before you could use that line.

i would separate out the one-time setup to be just make the builder, and then the 'use cloud' line would be in the build with cloud section with the comment attached being something that makes it clearer that the 'use cloud' line is a persistent setting

DOCKER-BAKE.md
Amndeep7: done

DOCKER-BAKE.md
Amndeep7: done

DOCKER-BAKE.md
Amndeep7: done

DOCKER-BAKE.md
Amndeep7: OBE

DOCKER-BAKE.md
Amndeep7: done

DOCKER-BAKE.md
Amndeep7: done

docker-bake.hcl
Amndeep7: done

docker-bake.hcl
Amndeep7: done

.github/workflows/release-lite-to-docker.yml
Amndeep7: @wdower this was not addressed

docker-bake.hcl
Amndeep7: done

Amndeep7 (changes requested):
I really like the wiki page now aside from the minor improvements listed in my comments.  it is substantially better than before.

i feel like the bake stuff (targets / vars / etc) is also considerably simpler and more consistent other than the couple of caveats that are also in my comments.

.github/workflows/push-lite-to-docker.yml
wdower: Whoops. Now done on SAF CLI.

.beads/.gitignore
wdower: done

docker-bake.hcl
wdower: struck

.github/workflows/push-lite-to-docker.yml
wdower: Done

.github/workflows/push-lite-to-docker.yml
wdower: re added

.github/workflows/push-lite-to-docker.yml
wdower: Updated. Turns out Docker-formatted images can't be saved to a local archive with both architectures inside; has to be one or the other. Tried switching to using OCI format instead, which could export multi-arch tarballs, but those can't be loaded by Docker, thereby defeating the whole point.

Settled for just letting the job build one or the other based on user input. Default behavior is to build amd64, which is exactly what it was doing before. Verified I can load the resulting image with Docker when we do it this way.

.github/workflows/release-lite-to-docker.yml
wdower: Now has been. We've settled on a) using an envar and logic in the bakefile itself to govern tags, instead of explicitly setting them in the jobfile, and b) updated the envar (TAG_SUFFIXES) to take a comma-delimited list instead of just one extra tag.

Issue #7494, opened 2025-12-09
[BUG] - SonarQube error


bfield308:
**Describe the bug**
Getting a 400 (malformed request) response during saf convert sonarqube2hdf with a saf container. 

**To Reproduce**
From a pipeline or docker running SAF Workstation
1. saf convert sonarqube2hdf -n <project_key> -u <url> -a [your_key] -o sonar-hdf-results.json​
2. connection starts
4. Dec-09-2025 15:07:07 +00:00] SonarQube2HDF response
[Dec-09-2025 15:07:07 +00:00] SonarQube2HDF 400
[Dec-09-2025 15:07:07 +00:00] SonarQube2HDF [object Object]

**Expected behavior**
Would expect the file to be generated

**Screenshots**
request: /api/issues/search?componentKeys=<project_key>&statuses=OPEN,REOPENED,CONFIRMED,RESOLVED&p=101
{"errors":[{"msg":"Can return only the first 10000 results. 10100th result asked."}]}

**Desktop (please complete the following information):**
 - mitre/saf/1.5.1

**Additional context**
Believe the issue to be related to the '101` restriction in the request as changing this manually on an API successfully resolves the issue. 

See feature request to add an 'ignore' funtion

Issue #7575, opened 2025-12-29
Sonarqube converter - failed paging


eskornev:
Hello, I was trying to get SonarQube (Community Edition v10.5.1 90531) report and was getting error 400:
```
[Dec-29-2025 16:22:52 +03:00] SonarQube2HDF Generating HDF for https://sonar.*** version: 10.5.1.90531
[Dec-29-2025 16:23:16 +03:00] SonarQube2HDF response
[Dec-29-2025 16:23:16 +03:00] SonarQube2HDF 400
[Dec-29-2025 16:23:16 +03:00] SonarQube2HDF [object Object]
[Dec-29-2025 16:23:16 +03:00] SonarQube2HDF request
...
responseUrl: 'https://sonar.***/api/issues/search?componentKeys=***&statuses=OPEN,REOPENED,CONFIRMED,RESOLVED&p=101'
```
While testing by curl with page parameter = 100 my test was successful, but with 101 like in logs - it was also failed with error 400, i.e. there is no page 101 in report. Is there any errors with paging while reading sonar reports?
Increased verbosity level was not helpful also.
Thanks in advance.

Amndeep7:
Hello @eskornev, sorry for the delay in response.  #7576 should resolve the problem you're experiencing, and will be available in the next release of the SAF CLI.  

In short, the problem was that SonarQube uses ElasticSearch underneath the hood to provide the list of findings per the given search query; however, ElasticSearch only allows up to 10k results per query so you need to find ways to modify the query to make it give you all the results.  I believe I have an approach now that should be able to handle this case far better than before.

If you find any further issues with this integration or any other bugs, please do file another issue!

PR #7576, opened 2026-01-15
Sonarqube more fixes

Amndeep7:


Amndeep7:
Resolves https://github.com/mitre/heimdall2/issues/7494 and https://github.com/mitre/heimdall2/issues/7575 as of the above commits.  Sonarqube has a limit of 10k returnable results for a given search even though the total number of issues might exceed that and there is no way to retrieve them without modifying the search in some capacity: https://community.sonarsource.com/t/cannot-get-more-than-10000-results-through-web-api/3662.

A future iteration of this might attempt to use the components api to iterate through all the individual components (since ideally you don't have 10k findings in a single file), ~~but was considered out of the scope of this PR due to a) the degree of complexity that would be added for b) the relatively minimal return - just fix your findings and the problem starts to go away by itself.~~

EDIT: That future iteration is WIP.  Additionally, functionality might be provided to apply filtering to the results (or a jq command that will do the equivalent, but ensure that there is at least an original source file that contains ALL the findings).

PR #7791, opened 2026-03-11
fix: Replace hardcoded SonarQube issue statuses with dynamic discovery from server

mikeBoterf:
## Summary

The SonarQube-to-HDF converter previously used a hardcoded list of issue status
filter values when calling the `/api/issues/search` endpoint. In our environment
running SonarQube 2025.x, one of the hardcoded values (`IN_SANDBOX`) was not
recognized as a valid issue status by the server. The valid statuses returned by
the API parameter definition were `OPEN`, `CONFIRMED`, `FALSE_POSITIVE`,
`ACCEPTED`, and `FIXED`. This required manual patching of the source at deploy
time to work around the mismatch.

Rather than simply updating the hardcoded list, this change introduces dynamic
status discovery so the converter adapts to whatever statuses a given SonarQube
server supports.

### Dynamic Status Discovery

The converter now queries the SonarQube server's `/api/webservices/list` endpoint
to discover the valid issue statuses at runtime. This endpoint returns the
`possibleValues` for the `issueStatuses` (or `statuses` for pre-10.4) parameter
directly from the server's API definition. This approach:

- Works across all SonarQube versions without version-specific hardcoding
- Automatically adapts if SonarQube adds or removes statuses in future releases
- Does not require a project key or any project-scoped permissions
- Is a single lightweight metadata request

### User Override via Constructor Parameter

A new optional `issueStatuses` parameter has been added to the `SonarqubeResults`
constructor. When provided, it bypasses automatic discovery entirely and uses the
supplied comma-separated status list directly. User-supplied values are normalized
to uppercase for consistency with the SonarQube API.

### Fallback Behavior

If the `/api/webservices/list` endpoint is unavailable or returns unexpected data,
the converter falls back to sensible defaults:

- Pre-10.4: `OPEN,REOPENED,CONFIRMED,RESOLVED`
- 10.4+: `OPEN,CONFIRMED,FALSE_POSITIVE,ACCEPTED,FIXED`

All three tiers (user override, server discovery, fallback) log their behavior so
operators can verify which statuses are being used.

### Three-Tier Priority

1. **User-supplied override** - exact control when needed
2. **Server discovery** - automatic, version-agnostic, no maintenance required
3. **Hardcoded fallback** - safety net if the metadata endpoint is unreachable

## Testing

Verified against a SonarQube 2025.x server. The `/api/webservices/list` endpoint
correctly returns the valid issue statuses for the server, and all project issues
were successfully retrieved and converted to HDF format.


mikeBoterf:
@aaronlippold @Amndeep7

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -891,10 +892,76 @@ export class SonarqubeResults {
...
+      logger.warn(
+        `Failed to discover issue statuses via /api/webservices/list, falling back to defaults`
+      );
+      logger.debug(inspect(e, {depth: 3}));
+    }
+
+    // Priority 3: Fallback defaults
+    const fallback = isLegacy
+      ? 'OPEN,REOPENED,CONFIRMED,RESOLVED'
+      : 'OPEN,CONFIRMED,FALSE_POSITIVE,ACCEPTED,FIXED';
```
Amndeep7: https://docs.sonarsource.com/sonarqube-server/2025.6/user-guide/issues/solution-overview#life-cycle

We can see here that at least as of 2025.6, that sandbox status is valid, and would consist of values that we should be examining.

I think instead of doing an allow-list approach we should flip the script and do a deny-list approach.  Then we can exclude from our end statuses that we know are not relevant for us, such as 'FALSE_POSITIVE' which we intentionally left out of the set of issues we were collecting.  It would also handle scenarios like yall are experiencing where you do not have the sandbox status yet.

The user configuration should probably be additional statuses we do not want (not a full replacement of the set that we're looking at).  Maybe enabling that flag should also do a nag at the user to file an issue so that we can fix it upstream - what do you think?

mikeBoterf:
@Amndeep7 Thanks for the review — the feedback made a lot of sense. Here's a summary of the changes made to address it:

**Switched from allow-list to deny-list approach** as requested. The converter now discovers all valid statuses from the server via `/api/webservices/list`, then excludes a default set rather than cherry-picking which ones to include. This naturally handles servers that do or don't support newer statuses like `IN_SANDBOX`.

**Default deny-list (10.4+):** `FALSE_POSITIVE`, `FIXED`
- `FALSE_POSITIVE` — intentionally excluded as before (user-designated non-issues)
- `FIXED` — no longer present in the code, purged after 30 days
- `IN_SANDBOX` is **not** in the default deny-list — if the server advertises it, it will be included in queries since those are real issues worth examining (per your note about the 2025.6 docs). Servers that don't support it simply won't advertise it, so it handles itself.

**User override (`excludeIssueStatuses`) replaces defaults** rather than adding to them, giving operators full control. If someone passes `--excludeIssueStatuses OPEN`, only `OPEN` is excluded — `FALSE_POSITIVE` and `FIXED` come back into the result set.

**Smart logging around the user flag:**
- If the user's exclusion list matches the defaults exactly → informs them they can omit the flag
- If it differs from defaults → suggests filing an upstream issue so the default deny-list can be updated for everyone

**Logging at every step** so operators can see what's happening:
1. Available statuses discovered from server (or fallback used)
2. Which deny-list is active (defaults vs. user-supplied)
3. Final query statuses and what was excluded

Existing tests remain compatible — the legacy fallback (pre-10.4) still produces `OPEN,REOPENED,CONFIRMED,RESOLVED` after excluding `CLOSED`, matching the original hardcoded behavior.

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -891,10 +892,127 @@ export class SonarqubeResults {
...
+    if (this.excludeIssueStatuses) {
+      // User-supplied deny-list REPLACES the defaults entirely
+      const userExclusions = this.excludeIssueStatuses
+        .split(',')
+        .map(s => s.trim().toUpperCase())
+        .filter(s => s.length > 0);
+      denySet = new Set(userExclusions);
+
+      // Smart nag: compare user list against defaults
+      const defaultSet = new Set(defaultDenyList.map(s => s.toUpperCase()));
```
Amndeep7: aren't they already allcaps?

libs/hdf-converters/src/sonarqube-mapper.ts
```diff
@@ -891,10 +892,127 @@ export class SonarqubeResults {
...
+      const userSet = new Set(userExclusions);
+      const sameAsDefault = defaultSet.size === userSet.size &&
```
Amndeep7: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/symmetricDifference

check if this is equal to 0.  we're using nodejs 22+ and also people should use modern browsers

libs/hdf-converters/src/sonarqube-mapper.ts
Amndeep7: make sure to fix in the other places where you do this map of already allcaps stuff

Amndeep7:
@mikeBoterf looks good aside from some nitpicks on my part and then satisfying the linter.  ping me when you've got those in place and we can merge this.

mikeBoterf:
@Amndeep7 Fixed both — removed the redundant `.toUpperCase()` calls on the already-uppercase constants and swapped the manual set comparison for `Set.symmetricDifference()`. Thanks for the cleanup catches.

Amndeep7:
my life is suffering @mikeBoterf 

you can either revert the change for the symmetric differences function or you're going to have to wait until i finish my pr for updating the version of typescript (which is old enough to not accept that function existing).  that pr might take a while, so actually yeah let's just go with the 'revert' on that particular change though please file an issue for me to come back and fix it when typescript is updated.

additionally please run the linter

mikeBoterf:
@Amndeep7 Reverted `symmetricDifference` back to manual comparison, ran the linter, and filed #7816 to track switching to `symmetricDifference()` once TypeScript is updated.

libs/hdf-converters/src/sonarqube-mapper.ts
mikeBoterf: Fixed — removed redundant `.toUpperCase()` in both places.

libs/hdf-converters/src/sonarqube-mapper.ts
mikeBoterf: Reverted to manual set comparison — current TS version lacks the type defs. Filed #7816 to revisit once TypeScript is updated.

PR #7803, opened 2026-03-13
Vitest concurrent

Amndeep7:


Amndeep7:
sonarqube might technically be right but it fails on build so we're stuck with it

PR #7883, opened 2026-03-30
BridgeCrew Checkov to HDF Mapper Added

clem-field:
## Summary

Adds a new BridgeCrew Checkov to HDF mapper to `@mitre/hdf-converters`, enabling conversion of Checkov JSON scan results into Heimdall Data Format (HDF).

### What's included:
- **CheckovMapper** (`libs/hdf-converters/src/checkov-mapper.ts`) — Converts Checkov JSON output (passed, failed, and skipped checks) into HDF, extending `BaseConverter`
- **CheckovCciMappingData** (`libs/hdf-converters/src/mappings/CheckovCciMappingData.ts`) — 1,341 Checkov check-to-CCI mappings derived from CIS v8 to NIST SP 800-53 Rev 5 control mappings
- **Fingerprinting** — Registers `INPUT_TYPES.CHECKOV` with identifying keys (`check_type`, `results`, `passed_checks`, etc.) for auto-detection
- **Frontend integration** — Wired into `report_intake.ts` for Heimdall2 UI upload support
- **Test suite** — 13 test cases covering conversion, severity mapping, CCI/NIST derivation, status mapping, source location, references, passthrough data, and raw data inclusion
- **Sample data** — Checkov JSON input samples and expected HDF output for regression testing
- **README** — Updated supported formats list (now #7)

### Mapper capabilities:
- Maps Checkov severity levels to HDF impact scores (critical=0.9, high=0.7, medium=0.5, low=0.3, info=0.0)
- Derives NIST SP 800-53 tags from CCI mappings, with fallback to default static code analysis tags (SA-11, RA-5)
- Supports `withRaw` flag to include raw Checkov data in HDF passthrough
- Handles guideline URLs with fallback to Bridgecrew documentation
- Covers 20+ check frameworks (AWS, Azure, GCP, Kubernetes, Docker, Terraform, Ansible, CloudFormation, etc.)

## Test plan
- [x] `yarn run test --reporter=verbose ./test/mappers/forward/checkov_mapper.spec.ts` — 13/13 tests passing
- [ ] Verify Heimdall2 UI upload with sample Checkov JSON output
- [ ] Confirm fingerprinting auto-detects Checkov JSON files


libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -0,0 +1,255 @@
+import {describe, expect, it} from 'vitest';
+import {CheckovMapper} from '../../../src/checkov-mapper';
+import {ExecJSON} from 'inspecjs';
```
Amndeep7: Preferred import order is

node built-ins in alpha order
3rd party deps in alpha order
relative deps in alpha order

```suggestion
import {ExecJSON} from 'inspecjs';
import {describe, expect, it} from 'vitest';
import {CheckovMapper} from '../../../src/checkov-mapper';
```

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+
+describe('CheckovMapper', () => {
+  const sampleCheckovJson = JSON.stringify({
```
Amndeep7: don't hardcode a string sample.  put it under sample_jsons/checkov-mapper/sample_input_report and load it in using fs.readFileSync

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
Amndeep7, line 1: please try to match the structure that the other tests have, ex. the burpsuite tests

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -0,0 +1,255 @@
...
+      passed: 1,
+      failed: 2,
+      skipped: 0,
+      parsing_errors: 0,
+      resource_count: 3,
+      checkov_version: '3.2.497'
+    }
+  });
+
+  it('should convert Checkov native JSON to HDF', () => {
```
Amndeep7: we can keep these tests if you'd like, but the way we usually write our tests is to have a straight up regression test where we confirm that the mapper is able to wholesale replicate a previously generated file

(separately this is why we have the fs.writefilesync lines commented out so that we can uncomment them out and regen the files as necessary if something changes such as a mapper improvement)

libs/hdf-converters/index.ts
Amndeep7, line 1: for lack of a better place to put it, please make sure to sign your commits using the --signoff option

libs/hdf-converters/sample_jsons/checkov-mapper/sample-checkov-hdf.json
Amndeep7, line 1: please make it clear which sample input matches up with the sample output and also ensure that there is an output for every input.

Amndeep7, line 1: see my comment re: tests, but usually it is worthwhile to have a separate describe block (or more if you're also testing withraw) for each sample input

libs/hdf-converters/src/mappings/CheckovToCciAndNistMappingData.ts
```diff
@@ -0,0 +1,2699 @@
+// CCI Mappings for Checkov/Bridgecrew rules
+// Updated: 2026-03-29
+//
+// Mapping Methodology:
```
Amndeep7, line 4: was this file handwritten or generated?  if generated, can you write a small script in the libs/hdf-converters/converters dir that will be able to regenerate it via an invocation from the scripts section of the package.json (see xml2json as an example)

ideally please make sure that if you're writing that script in typescript that node by itself should be able to run it.  this means you can't use certain typescript features.  see https://nodejs.org/learn/typescript/run-natively for more info

libs/hdf-converters/src/mappings/CheckovCciMappingData.ts
```diff
@@ -0,0 +1,2699 @@
...
+//
+//   Chain: CKV check → NIST SP 800-53 Rev 5 control → DISA CCI
+//
+//   Each mapped entry includes a comment documenting:
+//     - The NIST 800-53 Rev 5 control(s) for traceability
+//     - The Checkov policy name for context
+//
+//   Total checks mapped: 1341
+//   Source: checkov-checks-enriched.json (checkov --list + NIST/CCI enrichment)
+export const data: Record<string, string[]> = {
```
Amndeep7: can you provide both the nist control as well as the cci number?  

if you've done the mapping, might as well provide it as data instead of doing original source -> you mapped it to nist -> you mapped it to cci -> *the pre-existing utility maps it back into nist but maybe not the same way you did* -> inject into the tags array in the hdf file

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS
+} from './utils/global';
+
+const IMPACT_MAPPING: Map<string, number> = new Map([
+  ['critical', 0.9],
+  ['high', 0.7],
+  ['medium', 0.5],
+  ['low', 0.3],
+  ['info', 0.1]
+]);
```
Amndeep7, line 96: looking here in their sarif reporter (https://github.com/bridgecrewio/checkov/blob/main/checkov/common/output/sarif.py#L17-L23) seems like they use 'none' instead of 'info'

also we should probably adjust the scores to match what they're using (after scaling it to fit in [0,1] instead of [0,10])

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+
+function impactMapping(severity: unknown): number {
+  if (typeof severity === 'string') {
+    return IMPACT_MAPPING.get(severity.toLowerCase()) ?? 0.5;
+  }
+  return 0.5;
+}
```
Amndeep7: the samples you provided all had what appeared to be a default value of null.  the test with the inline example that you supplied had non-null values for the 'severity' field, but are they the actual values?  i couldn't find a schema or specification.  if you can find one that we could base our work off of, that would be great.

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
+import {ExecJSON} from 'inspecjs';
+import * as _ from 'lodash';
+import {version as HeimdallToolsVersion} from '../package.json';
+import {BaseConverter, ILookupPath, MappedTransform} from './base-converter';
+import {data as CheckovCciMappingData} from './mappings/CheckovCciMappingData';
+import {data as CciNistMappingData} from './mappings/CciNistMappingData';
+import {
+  DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS
+} from './utils/global';
```
Amndeep7, line 10: fix import ordering

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+      if (baseControl) {
+        nistTags.push(baseControl[0]);
+      }
+    }
+  }
+  const uniqueTags = [...new Set(nistTags)];
+  return uniqueTags.length > 0
+    ? uniqueTags
+    : DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS;
+}
```
Amndeep7: like i said in that other comment, if we already generated the nist control families in the first place, then we should just use that mapping instead of this one here.

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+
+interface CheckovCheck {
```
Amndeep7: ideally interfaces and enums are placed before functions

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  check_result: {result: string};
+  file_path?: string;
+  file_line_range?: number[];
+  resource?: string;
+  resource_address?: string;
+  severity?: string;
+  guideline?: string;
+  bc_check_id?: string;
+  code_block?: Array<[number, string]>;
+}
```
Amndeep7: for lack of a schema from them, we should try to generate one of our own so as to make types from it

i think you can use utilities like quicktype to maybe make one for you

ensure that the types are comprehensive for describing the files though they don't need to be exhaustive for all the different attributes that could pop up

i also prefer using types as opposed to interfaces

ex. i'd probably have like a CheckovReport type that would be like

```
type CheckovReport = {
    check_type: string;
    results: {
      passed_checks: CheckovCheck[],
      failed_checks: CheckovCheck[],
      skipped_checks: CheckovCheck(if it's the same type of output)[],
      parsing_errors: CheckovParseErrors(or whatever makes sense here)[]
    };
    summary: CheckovSummary; (having a separate checkovsummary type would allow for us to handle this scenario as a possible input too: https://github.com/bridgecrewio/checkov/issues/7462)
    url?: string; (what gets populated here if we do have a subscription?  does this parameter only show up on the free tier or does it get replaced with some url containing the enhanced analysis or what)
  }
}
```

and then for CheckovCheck itself it'd be something like 

```
type CheckovCheck = {
  check_id: string;
  what you have there but ideally with a better idea of what attributes are actually consistently there or not;
  check_result: {
    result: 'PASSED' | 'FAILED' | 'WHATEVER' or an enum;
    frequently seen but optional parameters?: structure;
    [property: string]: any; (handles the fact that there is more stuff out there that we should keep in mind exists but we likely won't directly use those attributes so who cares what their names are)
  }
}
```

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  const failedChecks = (_.get(results, 'failed_checks') || []) as CheckovCheck[];
+  const skippedChecks = (_.get(results, 'skipped_checks') || []) as CheckovCheck[];
+
+  const allChecks = [...passedChecks, ...failedChecks, ...skippedChecks];
+
+  return {
+    ...data,
+    all_checks: allChecks
+  };
+}
```
Amndeep7: while this nominally makes things easier, i really prefer to try to leave the original data structures as they are if at all possible since it makes it a lot easier to reason about the code if the data structure i see on disk matches what's in memory.

the usual reason for doing this sort of pre-processing is to do some major restructuring of the data.

take a look at the veracode mapper (https://github.com/clem-field/heimdall2/blob/22cb035df0ffc17bcae391c941fee9903ae976a4/libs/hdf-converters/src/veracode-mapper.ts) for an example of when the data restructuring is necessary but also how we can use the controls array to our advantage.

in this case, i would probably do something like 

```
controls: [
  {
    path: 'results.passed_checks',
    ...controlMapping()
  },
  {
    path: 'results.failed_checks',
    ...controlMapping()
  },
  {
    path: 'results.skipped_checks',
    ...controlMapping()
  },
]
```

and have controlmapping be a function that returns a mappedtransform which contains the mapping for your checkovcheck type

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  const parts: string[] = [];
+  if (resource) {
+    parts.push(`Resource: ${resource}`);
+  }
+  if (filePath) {
+    parts.push(`File: ${filePath}`);
+  }
+  if (lineRange && Array.isArray(lineRange) && lineRange.length >= 2) {
+    parts.push(`Line: ${lineRange[0]}-${lineRange[1]}`);
+  }
```
Amndeep7: seems like a code_block attribute is provided as well sometimes so we should include it esp since my skimming shows it gets provided when you are also provided a file_line_range

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+  return {
+    ...data,
+    all_checks: allChecks
+  };
+}
+
+export class CheckovMapper extends BaseConverter {
+  withRaw: boolean;
+
+  mappings: MappedTransform<
```
Amndeep7, line 133: this mapping is incomplete - more data can be put into the results that get shown in heimdall

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+    ExecJSON.Execution & {passthrough: unknown},
+    ILookupPath
+  > = {
+    platform: {
+      name: 'Heimdall Tools',
+      release: HeimdallToolsVersion,
+      target_id: 'Checkov (Bridgecrew)'
```
Amndeep7:
```suggestion
```

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+    },
+    version: HeimdallToolsVersion,
+    statistics: {},
+    profiles: [
+      {
+        name: 'Checkov',
+        version: {path: '$.summary.checkov_version'},
```
Amndeep7: you don't need the dollar sign since you're at the already at the root of the recursion into the data object.  the dollar sign is used if you need to break out of the recursion.

please make the fix in all other locations as relevant.

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  const allChecks = [...passedChecks, ...failedChecks, ...skippedChecks];
+
+  return {
+    ...data,
+    all_checks: allChecks
+  };
+}
+
+export class CheckovMapper extends BaseConverter {
```
Amndeep7: if we define that type for CheckovReport, you can use it to improve the typing of the 'data' attribute by doing something like this

```suggestion
export class CheckovMapper extends BaseConverter<CheckovReport> {
```

key value for this is that the transformer functions and mappings can be better type checked

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+    },
+    version: HeimdallToolsVersion,
+    statistics: {},
+    profiles: [
+      {
+        name: 'Checkov',
+        version: {path: '$.summary.checkov_version'},
+        title: {
+          path: '$.check_type',
+          transformer: (checkType: unknown): string => {
```
Amndeep7: see comment i left re: `extends BaseConverter<CheckovReport>`, but is there a chance that the typeof checktype is ever not string?  the only possibility seems to be that it's not provided at all, which seems like it shouldn't ever be the case

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+    ],
+    passthrough: {
+      transformer: (data: Record<string, unknown>): Record<string, unknown> => {
+        return {
+          auxiliary_data: [
+            {
+              name: 'Checkov',
+              data: {
+                check_type: _.get(data, 'check_type'),
```
Amndeep7: already included in the title
```suggestion
```

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+        sha256: ''
+      }
+    ],
+    passthrough: {
+      transformer: (data: Record<string, unknown>): Record<string, unknown> => {
+        return {
+          auxiliary_data: [
+            {
+              name: 'Checkov',
+              data: {
```
Amndeep7, line 193: i would put `url` and `parsing_errors` in here too - we do not want to leave any data unmapped, even if it is just dumping it in the passthrough

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+                transformer: (checkId: unknown): string[] => {
+                  if (typeof checkId === 'string') {
+                    const ccis = CheckovCciMappingData[checkId] || [];
+                    return ccis.length > 0
+                      ? deriveNistFromCCI(ccis)
+                      : DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS;
+                  }
+                  return DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS;
+                }
+              },
```
Amndeep7: re: comments on using the extends BaseConverter<CheckovResults> makes it so that we can confidently (or at least as confident we feel in our types) know that checkId will be a string, which will simply the implementation here

re: comments on already having done the nist mapping, just use that data instead

would make checkovccimappingdata look like this

```
export const data: Record<string, Record<string, string[]>> = {
  // CKV2_ADO_1: Ensure at least two approving reviews for PRs
  'CKV2_ADO_1': {
    cci: ['CCI-001501', 'CCI-001510'],
    nist: ['CM-3(2)', 'CM-5(1)']
  },
 ...
```

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+              severity: {path: 'severity'},
+              checkov_id: {path: 'check_id'},
+              resource: {path: 'resource'}
```
Amndeep7: items/metadata that are globally tied to the check itself are what should go in tags / the stuff that pops up in the details tab within heimdall

resource_address also often seems to be supplied and from what i saw has the same values as `resource` - so you might want to consider shoving it into the tags area too but since it's kinda duplicative you could also consider dumping it into the code tab which when we're not dealing with inspec is often used as a dumping ground for data that we'd like the user to be able to see outside of just the passthrough block

Amndeep7: your checkovcheck type says that resource is conditionally supplied so you might want to use the conditionallyprovidedattribute utility function here

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+            },
+            refs: [
+              {
+                transformer: (check: unknown) => {
+                  const guideline = _.get(check, 'guideline') as
+                    | string
+                    | null
+                    | undefined;
+                  if (
+                    typeof guideline === 'string' &&
```
Amndeep7: i prefer using lodash isString since it's implementation is a bit more robust

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+                    guideline.length > 0
+                  ) {
+                    return {url: guideline};
+                  }
+                  const checkId = _.get(check, 'check_id') as
+                    | string
+                    | undefined;
+                  if (typeof checkId === 'string') {
+                    return {
+                      url: `https://docs.bridgecrew.io/docs/${checkId.toLowerCase()}`
```
Amndeep7: typing in some random checkids does not make this resolve at all, even the base bridgecrew.io does not resolve

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+                  | number[]
+                  | undefined;
+                return _.omitBy(
+                  {
+                    ref: filePath,
+                    line: lineRange?.[0]
+                  },
+                  (value) =>
+                    value === undefined || value === null
+                );
```
Amndeep7: use the conditionallyprovideattribute util function

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+              }
+            },
+            title: {path: 'check_name'},
+            id: {path: 'check_id'},
+            desc: {path: 'check_name'},
```
Amndeep7: desc is an optional attribute so we don't need to provide it if it is goign to be the same as title, i.e. check_name; however, i see some "description" attributes that are all null.  maybe do a conditionallyprovideattribute here in that if it's non-null then we'll supply it?

libs/hdf-converters/src/utils/fingerprinting.ts
```diff
@@ -33,6 +34,9 @@ export enum INPUT_TYPES {
 // Fields to look for inside of JSON structures to determine type before passing to hdf-converters
 const fileTypeFingerprints: Record<INPUT_TYPES, string[]> = {
   [INPUT_TYPES.ASFF]: ['Findings', 'AwsAccountId', 'ProductArn'],
+  [INPUT_TYPES.CHECKOV]: ["check_type", "results", "results.passed_checks",
+    "results.failed_checks", "results.skipped_checks", "results.parsing_errors",
+     "summary", "url"],
```
Amndeep7, line 39: i would probably get rid of 'url'

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+            impact: {path: 'severity', transformer: impactMapping},
+            code: {
```
Amndeep7: the code attribute here is to either show the actual test code (really only relevant with inspec) or be a holder for any data that couldn't be reasonably mapped elsewhere but is relevant to that particular result / shouldn't just be dumped in passthrough

i would put any attributes that aren't explicitly mapped from checkovcheck into here.

the codeblock should instead be put into the code_desc attribute

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+                      return ExecJSON.ControlResultStatus.Failed;
+                    } else {
+                      return ExecJSON.ControlResultStatus.Skipped;
+                    }
+                  }
+                },
+                code_desc: {
+                  transformer: (check: Record<string, unknown>): string =>
+                    formatCodeDesc(check)
+                },
```
Amndeep7: should contain all the metadata relevant to the target for that particular check/test, ex. the codeblock, repo_file_path, entity_tags - whatever is actually useful for people to know.  the remainder can be dumped into the code attribute

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+                message: {
+                  transformer: (check: Record<string, unknown>): string => {
+                    const result = _.get(check, 'check_result.result');
+                    const name = _.get(check, 'check_name');
+                    return `${result}: ${name}`;
+                  }
+                },
```
Amndeep7: should contain all the data relevant to what actually passed/failed like in the usual case it would explain like 'noexec should be enabled on the /home dir' or something like that.  however, it does not appear that there's any of this type of information available so i would just get rid of this attribute entirely since it's optional and the contents we're putting inside are duplicated (the control status shows the result string of pass/fail/skip; the check_name is the title)

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -0,0 +1,272 @@
...
+                start_time: {
+                  transformer: (): string => new Date().toISOString()
```
Amndeep7: leave this as the empty string since checkov does not seem to provide this data

Amndeep7 (changes requested):

Amndeep7:
@clem-field solid showing with this mapper.  mostly just nitpicks on how i'd like things to be improved, not too many fundamental changes.  don't forget to pull the branch before pushing again since i merged in main to get the workflows to run

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: corrected on local branch. Will review before push

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: matched pattern. Was missing and samples generated using the same pattern.

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: corrected on local branch, ready for review

libs/hdf-converters/index.ts
clem-field: ack

libs/hdf-converters/sample_jsons/checkov-mapper/sample-checkov-hdf.json
clem-field: corrected

clem-field: corrected in commit, pushed up for review.

libs/hdf-converters/src/mappings/CheckovToCciAndNistMappingData.ts
clem-field: original was hand, updated to use a sanitized file. Included the raw data option and file structure per instructions.

libs/hdf-converters/src/mappings/CheckovCciMappingData.ts
clem-field: CCI and NIST updated, should see clean mapping now.

libs/hdf-converters/src/checkov-mapper.ts
clem-field: I believe this to be fixed. I did have to set none to 0 and not 0.0

libs/hdf-converters/src/checkov-mapper.ts
clem-field: should be good now

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: might need to chat about this one. check updated throughput and let me know what you think.

libs/hdf-converters/src/checkov-mapper.ts
clem-field: good update, code represented well in the visualization now too.
image

libs/hdf-converters/src/checkov-mapper.ts
clem-field: Lets chat about what I'm missing here

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -1,255 +1,82 @@
+import fs from 'node:fs';
```
Amndeep7: prefer not using the node prefix since this library gets used in the browser context so it's just easier to never have it enabled.  i am aware sonarqube will complain about it, feel free to ignore that

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -1,255 +1,82 @@
...
+import {ExecJSON} from 'inspecjs';
 import {describe, expect, it} from 'vitest';
 import {CheckovMapper} from '../../../src/checkov-mapper';
-import {ExecJSON} from 'inspecjs';
-
-describe('CheckovMapper', () => {
-  const sampleCheckovJson = JSON.stringify({
```
Amndeep7: you should make this sample go into the sample_input_report and make tests from it as well since it has attributes that are different from the others, ex. here we have some non-null severities

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -1,255 +1,82 @@
...
-      (c: ExecJSON.Control) => c.id === 'CKV_AWS_136'
-    );
-    expect(mediumControl!.impact).to.equal(0.5); // MEDIUM
-
-    // null severity defaults to 0.5
-    const nullSeverityControl = controls.find(
-      (c: ExecJSON.Control) => c.id === 'CKV_AWS_41'
-    );
-    expect(nullSeverityControl!.impact).to.equal(0.5);
+import {omitHDFTimes, omitVersions} from '../../utils';
```
Amndeep7: if you made the start_time parameter be the empty string then you should no longer need to use the omitHDFTimes wrapper function

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -1,28 +1,93 @@
...
+  check_type: string;
+  results: {
+    passed_checks: CheckovCheck[];
+    failed_checks: CheckovCheck[];
+    skipped_checks: CheckovCheck[];
+    parsing_errors: unknown[];
+  };
+  summary: CheckovSummary;
+  url?: string;
+  [property: string]: unknown;
```
Amndeep7: is there a likelihood that the top level report will have any other fields?  probably don't need this if there won't be.

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -1,28 +1,93 @@
...
+};
+
+type CheckovReport = {
+  check_type: string;
+  results: {
+    passed_checks: CheckovCheck[];
+    failed_checks: CheckovCheck[];
+    skipped_checks: CheckovCheck[];
+    parsing_errors: unknown[];
```
Amndeep7: just out of curiosity can we force something such that it will have a parsing error and we can see what that structure will look like?  this is not a blocker since i think we will likely be shoving this all into passthrough wholesale

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -1,28 +1,93 @@
...
+
+type CheckovSummary = {
+  passed?: number;
+  failed?: number;
+  skipped?: number;
+  parsing_errors?: number;
+  resource_count?: number;
+  checkov_version?: string;
+  [property: string]: unknown;
+};
```
Amndeep7, line 60: is there any chance in variability in this data structure?  in all the samples provided, it seems like only that precise set of attributes were there every time.

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -1,28 +1,93 @@
...
+  file_path?: string;
+  file_line_range?: number[];
+  resource?: string;
+  resource_address?: string;
+  severity?: string | null;
+  guideline?: string;
+  bc_check_id?: string;
+  code_block?: Array<[number, string]>;
+  [property: string]: unknown;
+};
```
Amndeep7, line 51: some of these types seem wrong since you currently have it be unknown | type whereas we see in checkov_sample that many of the attributes are put in there and set to null.  consequently things like resource_address should be `resource_address?: string | null;` - https://github.com/clem-field/heimdall2/blob/feature/SKW-0000_checkov_aws/libs/hdf-converters/sample_jsons/checkov-mapper/sample_input_report/checkov_sample.json#L41

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -1,255 +1,82 @@
...
+    const mapper = new CheckovMapper(
+      fs.readFileSync(
+        'sample_jsons/checkov-mapper/sample_input_report/checkov_sample.json',
+        {encoding: 'utf-8'}
+      )
+    );
+
+    // Uncomment to regenerate baseline: // NOSONAR
+    // fs.writeFileSync(
+    //   'sample_jsons/checkov-mapper/checkov_sample-hdf.json',
```
Amndeep7: does not appear like checkov_sample was committed and the regression test for it is missing.  like i said last time, you can have the test like you're doing below, but you should also have a blanket regression test too like you've done for checkov_json

libs/hdf-converters/src/checkov-mapper.ts
```diff
@@ -1,28 +1,93 @@
...
+// Checkov uses 'none' (not 'info'). Scaled to [0,1].
+// =========================================================================
+
 const IMPACT_MAPPING: Map<string, number> = new Map([
   ['critical', 0.9],
   ['high', 0.7],
   ['medium', 0.5],
   ['low', 0.3],
-  ['info', 0.1]
+  ['none', 0]
```
Amndeep7: we should match their numbers as well so ex. low was 3.9

libs/hdf-converters/src/checkov-mapper.ts
Amndeep7: i think there was some confusion.  the codeblock should be provided in the code_desc attribute.  the code attribute can be considered a dumping ground for all of the otherwise unmapped attributes in a given check, ex. entity_tags.

if you want to keep the code style formatting for the code_block, you can wrap it with html (i.e. the <pre> tag) - see https://github.com/clem-field/heimdall2/blob/22cb035df0ffc17bcae391c941fee9903ae976a4/libs/hdf-converters/src/sonarqube-mapper.ts#L1244

Amndeep7 (dismissed):

libs/hdf-converters/src/mappings/CheckovToCciAndNistMappingData.ts
clem-field: Started with seeding AI opus 4.6, mapped CIS Benchmark checks to the known checks in Checkov. Mapped from CIS v8 rule id to Checkov. CIS to NIST, NIST to DISA CCI. Peer reviewed by two personnel after for completeness.

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field  still need to fix - look for default at scale not hardcoded to numeric and validate from checkov

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: see file comments

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - revisit and use html wrapper like example

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - make sure all fields are identified and brought into the HDF side. Anything not used property: string - unkown = code tab

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done now

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done 4/4/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - validate correct fields in correct tabs

clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field validate all DRY

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/src/utils/fingerprinting.ts
clem-field: url is part of all files

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field  - validate correct information on details, test, code per discussion

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - make sure that code_desc contains finding (pass/fail information) add the "fix" to message

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - - make sure that code_desc contains finding (pass/fail information) add the "fix" to message

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field remove [property: string]: unknown; from line 57

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - see if we can force a parsing issue

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - review all sample files and see if it ever breaks the pattern for type

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @clem-field - use sonars and to the 100ths

clem-field: done 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: Done 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done.

libs/hdf-converters/src/checkov-mapper.ts
clem-field: Done 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done - to the extent possible without paid subscriptions to API or Prisma.

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done - 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done - 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: done 4/5/26

libs/hdf-converters/src/checkov-mapper.ts
clem-field: should be done with the updated fields/mapping and additional schema work.

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @Amndeep7  - There are optional flags that could bring in other fields at top level with subscriptions. Would prefer to leave this for now until further user testing. Thoughts?

libs/hdf-converters/src/checkov-mapper.ts
clem-field: @Amndeep7  - Done - introduced skipped and parsing error checks with local code and checkov scanning. Also, documenting the optional flags for the saf cli side for how to enrich the checkov scans. Not 100% coverage but a lot more verbose and we are able to handle the enrichment over time.

libs/hdf-converters/src/checkov-mapper.ts
clem-field: Done - resolved with additional field analysis and field mapping. Far less `?` and guessing. Used about 100 files spanning AWS, Azure services from local code for updated field types and mapping.

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
clem-field: reviewed ~100 files from scanning local code across AWS, Azure services. Updated the types and where they map to. much better analysis but needs further User testing on other services and flags. I do know additional flags will bring in different information now too.

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: sample report added and passing testing.

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
clem-field: done

libs/hdf-converters/src/checkov-mapper.ts
Amndeep7: if it's possible to bring in additional fields, then yes let's leave this as is but make sure to put any of those other fields into the passthrough.  easiest way would be to lodash omit any of the fields we know about and then push all the rest of them into passthrough.

libs/hdf-converters/test/mappers/forward/checkov_mapper.spec.ts
```diff
@@ -0,0 +1,203 @@
+import fs from 'fs'; // NOSONAR - matches MITRE test patterns (node:fs not used in other tests)
```
Amndeep7: We don't need the nosonar comments, I'll take care of them directly in SQ.  Please remove.

Amndeep7 (approved):

Amndeep7:
sq didn't accept that either but i've given up and just done the manual override

PR #8200, opened 2026-06-03
feat(frontend): support custom S3 endpoints in S3 upload menu

Ninjapenguin16:
# feat(frontend): support custom S3 endpoints in S3 upload menu

## Summary
Lets the S3 upload tab target **any S3-compatible server**
(MinIO, Ceph, on-prem object stores, etc.) instead of only AWS, while keeping
AWS as the default when no endpoint is specified.

## Motivation
The S3 reader previously hard-wired the AWS endpoints for both STS and S3 and
always performed an STS `GetSessionToken` exchange. Users running S3-compatible
storage couldn't load results from those buckets, and many such servers don't
implement the AWS STS API at all.

## Changes
- **New "S3 Endpoint" field** in the credentials step. Blank = AWS (unchanged
  behavior); a custom URL points the S3 client at that server using path-style
  addressing (required by most non-AWS implementations).
- **New "Use credentials directly (skip STS)" option.** When enabled, the
  supplied long-lived access/secret keys are used as-is and no STS session is
  requested - for servers that don't support STS. MFA login is disabled in this
  mode since it has no meaning without STS.
- Endpoint is plumbed through the STS client (endpoint only), the S3 client
  (endpoint + path-style), and persisted on the resulting `Auth` so file
  fetches reuse it.
- Endpoint and skip-STS settings are cached in `localStorage` alongside the
  existing access/secret/region fields.
- `AuthCreds.sessionToken` is now optional to support the no-STS path.

## Backwards compatibility
Fully backwards compatible by default, with the endpoint left blank and skip-STS unchecked.

## Testing
- `yarn frontend test:ci` - all existing tests pass (no regressions).
- `yarn frontend build` - builds cleanly standalone (no backend deps), no type
  errors (fork-ts-checker).
- Manual: verified the unchanged AWS flow, and a custom endpoint with skip-STS
  against an S3-compatible server (bucket browse + file load).

## Notes for reviewers
- No backend changes, this is entirely Heimdall Lite (frontend).


Amndeep7:
Thank you for the contribution @Ninjapenguin16!  It's gonna take a few days to review so we'll get back to you if we have any comments.

Ninjapenguin16:
Bump

Ninjapenguin16:
@Amndeep7 Is the PR still under review?

Amndeep7:
Hi @Ninjapenguin16 , yes it is, but other tasks are currently taking priority for us so it might take a bit longer to get to this PR.

If you have a pressing need for it being merged in order to support one of our government sponsors, please reach out to amann@mitre.org so that we can appropriately prioritize this effort.

Amndeep7 (approved):

Amndeep7:
Thank you for your contribution @Ninjapenguin16!

PR #8201, opened 2026-06-04
Feat/saml auth

seanlongcc:


apps/frontend/src/assets/okta.svg
Amndeep7, line 1: Add to the wiki page or maybe the backend readme or something where we are sourcing the logos from.  That website mentioned that they would appreciate getting attribution so wouldn't hurt mentioning it.

apps/frontend/src/assets/saml.svg
Amndeep7, line 1: should probably dig a little bit more to find out where the logo was originally created just to confirm it is actually the SAML format logo as opposed to anything else

apps/backend/src/config/config.service.ts
Amndeep7: saml is not oauth so remove it from the list

Amndeep7, line 68: actually maybe the thing that makes the most sense is to adjust this dto to reflect auth strategies more broadly

```
Dto({
  ...all the other stuff,
  authStrategies: {
    localLoginEnabled: this.isLocalLoginAllowed(),
    ldap: this.get('LDAP_ENABLED')..., // maybe we should change the name to be consistent with the 'Enabled' suffix
    ...(make an object from the supportedOauth array where the keys are the array items and the values are if it is enabled or not),
    samlEnabled: this.isSamlAllowed() // implement this func
  },
})
```

so that it looks like this when processed:

```
Dto({
  ...all the other stuff,
  authStrategies: {
    localLoginEnabled: true,
    ldapEnabled: false,
    githubEnabled: true,
    gitlabEnabled: false,
    ...
    samlEnabled: true
  },
})
```

apps/backend/src/config/config.service.ts
```diff
@@ -37,7 +37,10 @@ export class ConfigService {
   enabledOauthStrategies() {
```
Amndeep7: instead of returning an array of the valid functions, make it return a map where the key is the oath strategy name and the value is whether or not it's enabled

Amndeep7: you're going to have to look for all the places that oauthstrategies was used and make sure that it can handle the new data format

apps/frontend/src/assets/saml.svg
Amndeep7: based off of discussion it seems like this is an unofficial logo so we probably shouldn't use it.  instead we should come up with some generic logo or not provide a logo at all.

apps/frontend/src/components/global/login/LocalLogin.vue
```diff
@@ -144,90 +144,102 @@
...
                 max-width="32"
                 max-height="32"
-                :src="require('@/assets/okta_mark.png')"
+                :src="require('@/assets/okta.svg')"
               />
               <div class="pl-2">Login with Okta</div>
             </v-btn>
+            <v-btn
+              v-show="authStrategySupported('saml')"
+              id="oauth-saml"
```
Amndeep7: saml is not oauth so shouldn't be called that

apps/frontend/src/components/global/login/LocalLogin.vue
```diff
@@ -144,90 +144,102 @@
...
+              class="mt-5 flex-fill"
+              plain
+              @click="oauthLogin('saml')"
```
Amndeep7: saml is not oauth so likely that means that this function should be renamed to better reflect that it is for auth methods that are single-click on the user's side within heimdall (whereupon they do need to fill out their credentials on the other side)

dev/saml/realm/heimdall-dev-realm.json
Amndeep7, line 1: delete this dev directory before merge

apps/backend/src/authn/saml.strategy.ts
Amndeep7, line 1: make sure to follow along with all the other locations that LDAP does stuff, ex:

- the apps/backend/src/users/dto/create-user.dto.ts
- example envvars file
- the helm chart would need to be updated
- the environment variables wiki page
- tests (https://github.com/ory/mocksaml)

https://github.com/search?q=repo%3Amitre%2Fheimdall2+ldap&type=code&p=1

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
+import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { PassportStrategy } from '@nestjs/passport';
+import type { Profile } from '@node-saml/passport-saml';
+import { Strategy } from '@node-saml/passport-saml';
+import winston from 'winston';
+import { ConfigService } from '../config/config.service';
+import { User } from '../users/user.model';
+import { AuthnService } from './authn.service';
+
+type SAMLProfile = Profile & {
```
Amndeep7: let's try to use the upstream object as much as possible before overwriting it with our custom type.  for example, they provide a mail and email (as backup) attribute that we should probably leverage ourselves.  the type says that those attributes are optional so we should validate that they exist and if they don't then we should throw an error during the user login process that states that we are not getting the right claims

https://github.com/node-saml/node-saml/blob/master/src/types.ts#L242-L243

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+  email: string;
+  firstName: string;
+  lastName: string;
```
Amndeep7: need to do research on if that type properly provides the name information that we are wanting to use

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+function samlBooleanConfigValue(value: string | undefined): boolean | undefined {
```
Amndeep7: if you want to retain this style of having a utility function properly handle checking the status of a envvar, then i would put it in a separate PR cause we should update all of the boolean envvars to use the utlity function

instead you can probably match the style that we're doing elsewhere where you just do something like configService.get('envvar').toLowercase() === 'true' or 'false'

apps/backend/src/authn/saml.strategy.ts
Amndeep7, line 1: nonblocking: consider making a helper function that returns all of the SAML envvars that we need to do stuff with SAML, ex. gets the envvar (which we will need to create) that names the attribute that we will use to do 'given_name' / first_name claim equivalent

it should also handle default values for stuff that was not provided by the user

ex. user provides SAML_IDENTIFIER_FORMAT but not SAML_IDP_ISSUER.  is there a default value that we can provide there?  otherwise this function should throw an error that explains what happened.

we will then call the function within the constructor in order to make it fill out the information and throw the error at the time of construction if all of our needs are not satisfied.


```
const getSAMLInformation = (configService: ConfigService): Record<string, string> => {
  return {
    SAML_IDENTIFIER_FORMAT: configService.get('SAML_IDENTIFIER_FORMAT') || defaultvalue,
    SAML_IDP_ISSUER: configService.get('SAML_IDP_ISSUER') || throw Error('value not provided'),
    SAML_FIRST_NAME_CLAIM: configService.get('SAML_FIRST_NAME_CLAIM') || 'attributes.firstName'
    ...
  };
};
```

Amndeep7: if you wanted to be fancy instead of returning a record<string, string> you could define a type that looks like {SAML_IDENTIFIER_FORMAT: string, blah blah}.

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  if (normalizedValue !== 'true' && normalizedValue !== 'false') {
+    throw new TypeError('SAML boolean config values must be "true" or "false"');
+  }
+
+  return normalizedValue === 'true';
+}
+
+function samlClaimAttribute(profile: SAMLProfile, attributeName: string): string | undefined {
+  const attributeValue = Reflect.get(profile, attributeName);
```
Amndeep7: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect

reading up for @Amndeep7

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+  return typeof attributeValue === 'string' ? attributeValue : undefined;
+}
+
+@Injectable()
+export class SAMLStrategy extends PassportStrategy(Strategy as any, 'saml') {
```
Amndeep7, line 52: try to remove 'as any' cause sometimes the strategy 'shape'/type matches the new passportstrategy type.  other times, especially older / less maintained strategies need to use the 'as any' escape mechanism.

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+      ),
+    ),
+    transports: [new winston.transports.Console()],
+  });
+
+  constructor(
+    private readonly authnService: AuthnService,
+    private readonly configService: ConfigService,
+  ) {
+    super({
```
Amndeep7, line 78: add support for all the other config options for SAML

especially make sure that any of the certs / proxies / etc options are defined

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+  async validate(profile: SAMLProfile): Promise<User> {
+    this.logger.debug('in saml strategy file');
+    this.logger.debug(JSON.stringify(profile, null, 2));
+    const emailAttribute = this.configService.get('SAML_EMAIL_ATTRIBUTE') || 'email';
+    const firstNameAttribute = this.configService.get('SAML_FIRST_NAME_ATTRIBUTE') || 'firstName';
+    const lastNameAttribute = this.configService.get('SAML_LAST_NAME_ATTRIBUTE') || 'lastName';
+    const email = samlClaimAttribute(profile, emailAttribute);
+    const firstName = samlClaimAttribute(profile, firstNameAttribute);
+    const lastName = samlClaimAttribute(profile, lastNameAttribute);
+    if (email && email.length > 0) {
```
Amndeep7: we also need to validate that first name and last name were provided properly, not just email

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  async validate(profile: SAMLProfile): Promise<User> {
+    this.logger.debug('in saml strategy file');
+    this.logger.debug(JSON.stringify(profile, null, 2));
+    const emailAttribute = this.configService.get('SAML_EMAIL_ATTRIBUTE') || 'email';
+    const firstNameAttribute = this.configService.get('SAML_FIRST_NAME_ATTRIBUTE') || 'firstName';
+    const lastNameAttribute = this.configService.get('SAML_LAST_NAME_ATTRIBUTE') || 'lastName';
+    const email = samlClaimAttribute(profile, emailAttribute);
+    const firstName = samlClaimAttribute(profile, firstNameAttribute);
+    const lastName = samlClaimAttribute(profile, lastNameAttribute);
```
Amndeep7: if any of these are undefined we should throw an error at this point stating that the user did not configure the envvars properly

Amndeep7 (changes requested):

apps/backend/src/authn/saml.strategy.ts
Amndeep7: @seanlongcc @owilliams23 yeah let's not use this, seems like it's intended usecase is for something else as per that MDN article: "The major use case of Reflect is to provide default forwarding behavior in Proxy handler traps."

apps/backend/src/authn/saml.strategy.ts
seanlongcc: removed, the upstream object provides what we need already

apps/backend/src/authn/saml.strategy.ts
seanlongcc: removed, turned into a helper file

apps/backend/src/authn/authn.module.ts
```diff
@@ -45,6 +46,7 @@ async function buildHttpsProxyAgent(proxyUrl: string): Promise<Agent> {
     GoogleStrategy,
     LDAPStrategy,
     ApiKeyService,
+    SAMLStrategy,
```
Amndeep7, line 49: nitpick: move this up so all the strategies are in alphabetical order

libs/common/interfaces/config/auth-strategy.interface.ts
```diff
@@ -0,0 +1,28 @@
...
+export const AUTH_STRATEGIES = [
+  AUTH_STRATEGY.LOCAL,
+  AUTH_STRATEGY.LDAP,
+  AUTH_STRATEGY.GITHUB,
+  AUTH_STRATEGY.GITLAB,
+  AUTH_STRATEGY.GOOGLE,
+  AUTH_STRATEGY.OKTA,
+  AUTH_STRATEGY.OIDC,
+  AUTH_STRATEGY.SAML,
+] as const;
```
Amndeep7: should be able to just do object.values(auth_strat) here instead of typing it all out by hand

libs/common/interfaces/index.js
Amndeep7, line 1: this file seems redundant between auth-strategy.interface.ts and index.ts.

test/support/pages/LoginPage.ts
```diff
@@ -11,6 +11,19 @@ export default class LoginPage {
     cy.get('[data-cy=ldapLoginButton]').click();
   }
 
+  loginSaml(username: string): void {
+    const origin = String(
+      Cypress.env('MOCKSAML_ORIGIN') || 'http://localhost:4000',
+    );
```
Amndeep7:
```suggestion
    const origin = 'http://localhost:4000';
```

think it's probably fine to hardcode it instead of using an undocumented envvar.

apps/frontend/src/assets/saml.svg
Amndeep7: where was the new logo sourced from?  might be worthwhile to instead swap it out with something that looks like this: https://worldvectorlogo.com/logo/saml-1

apps/frontend/src/components/global/login/LocalLogin.vue
```diff
@@ -139,95 +139,113 @@
...
+              class="mt-5 flex-fill"
+              plain
+              @click="startExternalLogin('saml')"
+            >
+              <v-img
+                max-width="32"
+                max-height="32"
+                :src="require('@/assets/saml.svg')"
+              />
+              <div class="pl-2">Login with SAML</div>
```
Amndeep7: instead of a hard coded 'login with saml' we should follow the example of OIDC_NAME and allow the user to specify what they want.  similar to oidc and unlike the other stuff like okta or google, saml is a standard that a bunch of identity providers support so we should make the admins identify what they want as the name

apps/frontend/src/views/Login.vue
```diff
@@ -103,16 +103,15 @@ export default class Login extends Vue {
 
   get anyAuthProvidersAvailable() {
     return (
-      ServerModule.localLoginEnabled || ServerModule.enabledOAuth.length !== 0
+      ServerModule.enabledAuthStrategies.includes('local') ||
+      ServerModule.enabledAuthStrategies.some(
+        (strategy) => strategy !== 'local' && strategy !== 'ldap'
+      )
```
Amndeep7: could be simplified to just check if the enabledauthstrats list is nonempty since we have now combined all auth formats into one list

apps/frontend/src/views/Login.vue
```diff
@@ -103,16 +103,15 @@ export default class Login extends Vue {
...
     );
   }
 
   get ldapenabled() {
```
Amndeep7: please fix capitalization for this function

test/integration/login.cy.ts
```diff
@@ -53,6 +53,14 @@ context('Login', () => {
       // Make sure all the fields exist
       userModalVerifier.verifyFieldsExist();
     });
+    it('authenticates a saml user', () => {
```
Amndeep7, line 57: both oidc and saml should have a login verifier test to make sure that the button for them is actually there

https://github.com/mitre/heimdall2/blob/master/test/support/verifiers/LoginPageVerifier.ts

test/support/pages/LoginPage.ts
```diff
@@ -11,6 +11,19 @@ export default class LoginPage {
     cy.get('[data-cy=ldapLoginButton]').click();
   }
 
+  loginSaml(username: string): void {
+    const origin = String(
+      Cypress.env('MOCKSAML_ORIGIN') || 'http://localhost:4000',
```
Amndeep7: speaking of which in a separate pr we should probably handle updating the cypress test readme to explain how to run the cypress tests + spin up the mocks such as this saml one

ex. explain that the idp cert needs to be in pem format or as a base64 encoded string (https://github.com/node-saml/node-saml?tab=readme-ov-file#configuration-option-idpcert) and that you have to use the same cert in the .env file as for the docker container

apps/backend/src/main.ts
```diff
@@ -69,7 +69,9 @@ async function bootstrap() {
   // Sessions was previously set to only be used for oauth callbacks
   // but now is used for Tenable authentication as well.
   if (
-    configService.enabledOauthStrategies().length ||
+    configService
+      .enabledAuthStrategies()
+      .some((strategy) => strategy !== 'local') ||
```
Amndeep7: i feel like there should be more thinking of what the original purpose was here and esp why sessions were restricted to just oauth and then later tenable.  why are ldap and local login not getting sessions?  was the choice made then correct?  should it be changed?  which path should saml follow?

in any case, this does change does not match the prior behavior as it will let non-oauth auth methods to go through

apps/backend/src/config/config.service.ts
```diff
@@ -1,6 +1,16 @@
...
-import {StartupSettingsDto} from './dto/startup-settings.dto';
+import { StartupSettingsDto } from './dto/startup-settings.dto';
+
+const OAUTH_AUTH_STRATEGIES = [
+  AUTH_STRATEGY.GITHUB,
+  AUTH_STRATEGY.GITLAB,
+  AUTH_STRATEGY.GOOGLE,
+  AUTH_STRATEGY.OKTA,
+  AUTH_STRATEGY.OIDC,
+] as const;
```
Amndeep7: this should be defined with the other auth strat types / enums

apps/backend/src/authn/resolve_claim.ts
Amndeep7, line 1: why was this extracted out into its own file?  should probably be left as a function within saml.strategy.ts though you don't need to make it an instance function of the strategy class

apps/backend/src/authn/resolve_claim.ts
```diff
@@ -0,0 +1,21 @@
...
+import _ from 'lodash';
+
+export function getRequiredClaim(
+  claims: Record<string, unknown>,
+  defaultClaimName: string,
+  configuredClaimName?: string,
+): string {
+  const resolvedClaimName = configuredClaimName || defaultClaimName;
+  const claimValue = _.get(claims, [resolvedClaimName])
+    ?? _.get(claims, resolvedClaimName);
```
Amndeep7: why do we need the array version of this get?  isn't that for defining a path into the data object, here 'claims'?  shouldn't the claims object be flat?

apps/backend/src/authn/resolve_claim.ts
```diff
@@ -0,0 +1,21 @@
...
+  const firstClaimValue = Array.isArray(claimValue)
+    ? claimValue.find(value => typeof value === 'string' && value.length > 0)
+    : claimValue;
```
Amndeep7: what is the point of identifying the first non-empty string within an array?  what stuff is being returned in these claims that we should be ignoring?

furthermore, people have first names that consist of more than one word (ex. Mary Anne) so i feel like logically this might be unsound

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  constructor(
+    private readonly authnService: AuthnService,
+    private readonly configService: ConfigService,
+  ) {
+    const samlAudience = configService.get('SAML_AUDIENCE');
+    super({
+      acceptedClockSkewMs: Number(configService.get('SAML_ACCEPTED_CLOCK_SKEW_MS') ?? 0),
+      additionalAuthorizeParams: configService.get('SAML_ADDITIONAL_AUTHORIZE_PARAMS') ? JSON.parse(configService.get('SAML_ADDITIONAL_AUTHORIZE_PARAMS')!) : undefined,
```
Amndeep7: should probably use .has instead of .get for the predicate part of the ternary

apps/backend/src/authn/saml.strategy.ts
```diff
@@ -0,0 +1,134 @@
...
+    return this.authnService.validateOrCreateUser(
+      getRequiredClaim(
+        profile,
+        'email',
+        this.configService.get('SAML_EMAIL_ATTRIBUTE'),
+      ),
+      getRequiredClaim(
+        profile,
+        'firstName',
+        this.configService.get('SAML_FIRST_NAME_ATTRIBUTE'),
```
Amndeep7: in order to match (i believe) elsewhere in the strategies areas, we should probably call this SAML_GIVEN_NAME_ATTRIBUTE and family name for the other var

Amndeep7 (changes requested):

libs/common/interfaces/index.js
seanlongcc: ⚠️ AI assisted explanation ⚠️ 
`index.js` was added during the authentication refactor that consolidated Local, LDAP, OAuth, OIDC, and SAML into shared auth-strategy definitions. Before that refactor, the common interfaces package was mostly used for TypeScript types, which disappear during compilation. The refactor introduced runtime values such as `AUTH_STRATEGY` and `AUTH_STRATEGIES`, used by `ConfigService` and `@IsIn(AUTH_STRATEGIES)`.
The compiled backend loads those values with `require('@heimdall/common/interfaces')`. Node resolves that directory through `index.js`; it cannot execute `index.ts` directly. Therefore, `index.js` provides the runtime CommonJS exports while `index.ts` provides TypeScript exports.
The duplicated definitions have been moved into `auth-strategy.json`, so both adapters now use one source of truth.
Removing `index.js` properly would require a separate `libs/common` packaging refactor: compiling the TypeScript entry point, adding package exports, changing build order, updating Docker/CI, and verifying every frontend/backend consumer of `@heimdall/common`.

apps/frontend/src/views/Login.vue
seanlongcc: nonempty check breaks LDAP only setups because LDAP is in the combined list but still uses a separate login tab, so it would show an empty login tab.

tonydao2 (approved):

PR #8414, opened 2026-07-15
Bug fix, entity expansion limit

NickShumar:
Expanding entity processing limit for fast-xml-parser


libs/hdf-converters/src/base-converter.ts
ejaronne, line 1: The expansion count needs to be greatly increased. Lots of Nessus text content is quoted and formatted, using for example:
```
&quot;  ->  "   one expansion
&apos;  ->  '   one expansion
&amp;   ->  &   one expansion
&lt;    ->  <   one expansion
&gt;    ->  >   one expansion
```
I was easily able to find over a million &quot; in a single-target nessus scan. Nessus files can also be very large against an entire subnet of multiple hosts.

Can the maxTotalExpansions be uncapped (unlimited)

libs/hdf-converters/src/base-converter.ts
```diff
@@ -69,6 +69,13 @@ export function parseXml(
     parseAttributeValue: false,
     parseTagValue: false,
     removeNSPrefix: true,
+    processEntities: {
+      enabled: true,
+      maxEntitySize: 10000,
+      maxTotalExpansions: 10000000,
```
ejaronne: Up the default to 1,000,000,000

libs/hdf-converters/src/base-converter.ts
Amndeep7: > Can the maxTotalExpansions be uncapped (unlimited)

not as far as i'm aware

Amndeep7 (changes requested):
can you please pare back to just the required changes (i.e. probably just the base converter and the nessus mapper file) 

the set of changes should probably just be the base converter exposing those parameters and then the nessus mapper overriding the expansion limit to our larger limit so that folks can usually easily handle it by hand.  however the additional change to the nessus mapper is a new parameter for the constructor that can further modify the expansion limit.  this parameter will be used by the saf cli.

libs/hdf-converters/src/nessus-mapper.ts
```diff
@@ -213,8 +215,17 @@ export class NessusResults {
   data: Record<string, unknown>;
   customMapping?: MappedTransform<ExecJSON.Execution, ILookupPath>;
   withRaw: boolean;
-  constructor(nessusXml: string, withRaw = false) {
-    this.data = parseXml(nessusXml);
+  constructor(
+    nessusXml: string,
+    withRaw = false,
+    maxTotalExpansions = DEFAULT_NESSUS_MAX_TOTAL_EXPANSIONS,
```
Amndeep7: this should be a parameter of type X2jOptions

libs/hdf-converters/src/nessus-mapper.ts
```diff
@@ -213,8 +215,17 @@ export class NessusResults {
...
+  ) {
+    this.data = parseXml(nessusXml, {
```
Amndeep7, line 229: we should then actually extract out the max total expansions parameter here, before using it as you are below

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #8459, opened 2026-07-23
fix(hdf-converters): expand collapsed AWS Config NIST sub-parts (#8458)

### LINKED ISSUE #8458: [BUG] AWS Config mapper collapses NIST control sub-parts into one tag — all but the first become unreachable
**Describe the bug**

The AWS Config NIST mapping stores multiple control sub-parts collapsed into a
single tag string — e.g. `AU-2(a)(d)`, `SI-4(a)(b)(c)`, `IA-2(1)(11)`,
`IA-5(1)(a)(d)(e)` — rather than as separate tags (`AU-2.a` + `AU-2.d`, etc.).

Because `parse_nist` flattens the parenthesized remainder into one specifier
chain and control matching is a strict prefix comparison, **only the first
sub-part in each collapsed tag is reachable**. Every subsequent sub-part is
silently dropped from filtering and 800-53 rollups — the control still *looks*
mapped, so there is no error and nothing obviously missing until you filter by a
sub-part and get no results.

**Where it comes from**

`libs/hdf-converters/src/mappings/AwsConfigMappingData.ts` splits `NIST-ID` on
`|` only, so collapsed sub-parts survive intact as single tags:

```ts
// root-account-mfa-enabled
'NIST-ID': 'AC-2(j)|IA-2(1)(11)'

// api-gw-execution-logging-enabled / elb-logging-enabled
'NIST-ID': 'AU-2(a)(d)|AU-3|AU-12(a)(c)'

// securityhub-enabled
'NIST-ID': 'AC-2(1)|AC-2(4)|AC-2(12)(a)|AC-2(g)|AC-17(1)|AU-6(1)(3)|CA-7(a)(b)|SA-10|SI-4(2)|SI-4(4)|SI-4(5)|SI-4(16)|SI-4(a)(b)(c)'
```

`AwsConfigMapping.searchNIST()` passes these through unchanged, and
`libs/inspecjs/src/nist.ts` (`parse_nist`, ~L202-217) matches the control then
splits the remainder on `SPEC_SPLITTER` (`/[\s\(\)\.]+/`) into a flat chain:

| raw tag | resulting `subSpecifiers` |
|---|---|
| `IA-2(1)(11)` | `[IA, 2, 1, 11]` |
| `AU-2(a)(d)` | `[AU, 2, a, d]` |
| `SI-4(a)(b)(c)` | `[SI, 4, a, b, c]` |

`NistControl.compare_lineage` (`nist.ts` ~L74-89) then returns `-1` on the first
differing element, so matching is strict-prefix.

**To Reproduce**

1. Convert an AWS Config result set containing the `root-account-mfa-enabled`
   rule (source identifier `ROOT_ACCOUNT_MFA_ENABLED`) to HDF. It carries
   `IA-2(1)(11)`.
2. Load the resulting HDF in Heimdall.
3. Filter / roll up by **`IA-2(1)`** — the control matches
   (`[IA,2,1]` is a prefix of `[IA,2,1,11]`).
4. Filter / roll up by **`IA-2(11)`** — the control does **not** match, because
   `[IA,2,11]` differs from `[IA,2,1,11]` at index 2.

The same holds for `api-gw-execution-logging-enabled` (`AU-2(a)(d)`): it is
reachable under `AU-2.a` but not `AU-2.d`; and for `securityhub-enabled`
(`SI-4(a)(b)(c)`): reachable under `SI-4.a`, invisible under `SI-4.b` / `SI-4.c`.

**Expected behavior**

Each control sub-part should be its own NIST tag, so that a rule mapped to
`AU-2(a)(d)` is discoverable under both `AU-2.a` and `AU-2.d`, and
`IA-2(1)(11)` yields the two sibling enhancements `IA-2(1)` and `IA-2(11)`.

**Additional context**

Three distinct shapes are collapsed in the mapping data, and they may not all
want the same treatment:

1. **Statement parts** — `AU-2(a)(d)`, `SI-4(a)(b)(c)`, `CA-7(a)(b)`,
   `AC-5(c)`, `CP-9(b)`, `SA-3(a)`. Should expand to sibling parts.
2. **Enhancements** — `IA-2(1)(11)`, `AU-6(1)(3)`. These are *sibling
   enhancements*, but the flattened chain renders 11 as a child of 1, which
   misrepresents the 800-53 hierarchy.
3. **Mixed** — `IA-5(1)(a)(d)(e)`, `AC-2(12)(a)`: an enhancement plus statement
   parts. Note `IA-5(1)(a)(d)(e)` flattens to six specifiers
   (`[IA,5,1,a,d,e]`), which exceeds the `max_specifiers: 5` default in
   `DEFAULT_CANONIZATION_CONFIG`.

**Open question for maintainers:** what is the canonical target form? Expanding
`AU-2(a)(d)` into `AU-2.a` + `AU-2.d` matches the period notation used elsewhere,
but `AU-2 (a)` + `AU-2 (d)` is also consistent with the canonization config's
`add_parens` / `add_periods` options. Happy to submit a PR against the mapping
data once the preferred form is confirmed — the bulk of the fix looks like it is
expanding the `NIST-ID` entries (and/or splitting them in
`AwsConfigMapping`'s constructor) rather than changing the parser.

**Impact**

This is a compliance-coverage problem rather than a crash: controls appear
mapped, so nothing surfaces as broken, but they do not appear under the sub-part
a reviewer would filter or roll up by. For teams using Heimdall as the system of
record for 800-53 coverage, that means silently under-reported control coverage.

**Related**

- #4715 — "Contextualization process should canonize NIST tags" is adjacent
  (Heimdall-side canonization, prompted by the XCCDF mapper); this issue is the
  AWS Config mapping data specifically.
- #2647 — earlier `aws_config` NIST tag issue (`unmapped`), closed; different
  problem in the same mapping.
- #2529 — coercing NIST tags to a list.

clem-field:
## Summary

Fixes #8458. The AWS Config → HDF NIST mapping stored multiple 800-53 sub-parts as single **collapsed** tags (e.g. `IA-2(1)(11)`, `AU-2(a)(d)`, `IA-5(1)(a)(d)(e)`). Because `parse_nist` flattens the parenthesized remainder into one specifier chain and `NistControl.compare_lineage` is a **strict prefix** match, only the *first* sub-part in each collapsed tag was reachable — every sibling was silently dropped from filtering and 800-53 rollups. Controls still *looked* mapped, so nothing surfaced as broken; they just never appeared under the sub-part a reviewer would filter by.

This PR expands the collapsed tokens into individual tags directly in `AwsConfigMappingData.ts` (the source of truth), leaving the parser untouched.

## What changed

`libs/hdf-converters/src/mappings/AwsConfigMappingData.ts` — 8 collapsed sibling tokens expanded across all occurrences (per-entry de-duplicated):

| collapsed | → expanded |
|---|---|
| `AU-2(a)(d)`, `AU-12(a)(c)`, `CA-7(a)(b)`, `SI-4(a)(b)(c)` | statement-part siblings |
| `AU-6(1)(3)`, `IA-2(1)(11)`, `IA-2(1)(2)(11)` | sibling enhancements |
| `IA-5(1)(a)(d)(e)` | `IA-5(1)(a)` / `IA-5(1)(d)` / `IA-5(1)(e)` — keeps the enhancement `(1)` prefix |

Two tokens are **deliberately left intact** — `AC-2(12)(a)` and `CM-8(3)(a)` are legitimate *enhancement + its own statement part* single controls, already reachable via prefix match; splitting them would produce incorrect base-level mappings (`AC-2(a)` ≠ `AC-2(12)(a)`).

## Behavior

Reproduction from the issue (`ROOT_ACCOUNT_MFA_ENABLED`, `IA-2(1)(11)`):

- **Before:** reachable under `IA-2(1)`, **invisible** under `IA-2(11)` (`[IA,2,11]` differs from flattened `[IA,2,1,11]` at index 2).
- **After:** reachable under **both** `IA-2(1)` and `IA-2(11)`.

## Tests

Adds `libs/hdf-converters/test/mappings/aws_config_mapping.spec.ts` (there was no prior coverage for `AwsConfigMapping`). It asserts the expansion, the kept-nested controls, and — via inspecjs `parse_nist`/`contains` — that previously-unreachable sub-parts now match under a filter.

## Scope / notes

- No parser or `AwsConfigMapping` logic change; only the mapping data and a new test.
- **Not** a match to `mitre/hdf-libs`: its Rev-4 data is identical to ours and still collapsed. hdf-libs only produces uncollapsed output on its **Rev-5** path, whose rows are a separately-authored 800-53 Rev4→Rev5 re-map (different controls), not a mechanical expansion. This fix is the correct treatment for the Rev-4 data Heimdall ships today.
- **Known limitation, tracked separately:** the mapping carries 106 rules, all Rev 4, while AWS currently publishes ~822 managed Config rules. A Rev-5 re-baseline / coverage expansion is a larger follow-up (issue to be filed).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


clem-field:
@wdower and @ejaronne  - can this get release and saf cli updated?

aaronlippold (approved):

PR #8468, opened 2026-07-29
feat(hdf-converters): select the NIST 800-53 revision for AWS Config mappings

clem-field:
## Summary

`AwsConfigMappingData` now carries a Rev 4 **and** a Rev 5 row for all 394 AWS Config managed rules, but `AwsConfigMapping` could only ever return one of them. It built a single flat lookup keyed by rule name and source identifier, so the two rows for a rule overwrote each other and only whichever came last in the table was reachable — half the table was dead data, and the row order was silently load-bearing.

This keys the lookups by revision, so both revisions are usable.

## Changes

- `AwsConfigMapping` takes the revision it resolves against: `new AwsConfigMapping(5)`. It defaults to Rev 4, so every existing caller keeps the tags it has today.
- `searchNIST(identifiers, rev?)` takes an optional per-lookup override, for callers that want one revision without building a second instance.
- The public `awsConfigRuleNameMappings` / `awsConfigRuleSourceIdentifierMappings` fields still expose the selected revision's table, so external readers are unaffected.
- `AwsConfigMapper` gains an optional fourth constructor argument `nistRev`, also defaulting to Rev 4, so an `aws_config` conversion can opt into Rev 5.
- The data file's header comment is updated: row order is now presentational rather than load-bearing.

## Why Rev 4 stays the default

Changing the default would change the NIST tags on every existing `aws_config` and Security Hub conversion. That is a data-visible change and belongs in its own decision, not smuggled in with the plumbing. This PR only makes Rev 5 *reachable*.

## Tests

`test/mappings/aws_config_mapping_revisions.spec.ts` covers:

- the default instance is Rev 4 and returns exactly what it returned before;
- the two revisions return their own controls for the same rule — Rev 5 folded `IA-2(11)` into `IA-2(2)`/`IA-2(6)`, and Rev 4 statement letters like `AC-2(j)` have no Rev 5 counterpart, so the sets must not be conflated;
- the per-lookup override;
- every rule in the table has both revisions;
- **every rule resolves under both revisions** — the direct regression test for the collision above.

The existing `aws_config_mapping.spec.ts` (#8458) is untouched and still passes.

## Notes

- Stacked on #8459, which adds the complete Rev 4 + Rev 5 table. Merge that first; this diff is the mapping-layer change on top.
- The ASFF Security Hub path (`case-security-hub.ts`) constructs its own `AwsConfigMapping()` internally and stays on Rev 4. Threading a revision through the ASFF options is a larger API change and is left as a follow-up.


libs/hdf-converters/src/mappings/AwsConfigMappingData.ts
wdower, line 1: @clem-field -- The actual mapping data and its provenance is the only thing holding up merge at this point; just wanted to point out we want to do some review on your process here before we fold it into Heimdall proper.

clem-field: @wdower - the file was predominantly existing and then used the following tiers:

Tier 1 — AWS Config "Operational Best Practices for NIST 800-53" (235 rows)
- Rev 4: https://docs.aws.amazon.com/config/latest/developerguide/operational-best-practices-for-nist-800-53_rev_4.html
- Rev 5: https://docs.aws.amazon.com/config/latest/developerguide/operational-best-practices-for-nist-800-53_rev_5.html

Tier 2 — Security Hub NIST 800-53 r5 standard, per-service control pages (Rev 5 only, 173 rows)
- Index: https://docs.aws.amazon.com/securityhub/latest/userguide/toc-contents.json
- 84 pages matching *-controls.html under https://docs.aws.amazon.com/securityhub/latest/userguide/ (count as of 2026-08-26; the index drives it)

Tier 3 — derived-theme heuristic (87 rows, Rev 5 only). No AWS mapping exists for these rules; the tier assigns them the NIST core AWS gave to ≥75% of same-theme rules in Tiers 1–2. Never invents a control.
- Method: https://github.com/mitre/hdf-libs/blob/main/hdf-mappings/scripts/generate-awsconfig-mappings.mjs (hdf-libs#175, merged 2026-07-26)
- Rule names: https://docs.aws.amazon.com/config/latest/developerguide/toc-contents.json ("List of Managed Rules")
- Controls: inherited from the Tier 1/2 rows — no external source

Cross-fill (294 rows). Existence check only, not a NIST-published mapping: keep the control if the target revision defines it, else widen to the base control, else drop.
- https://raw.githubusercontent.com/usnistgov/oscal-content/main/nist.gov/SP800-53/rev4/json/NIST_SP-800-53_rev4_catalog.json
- https://raw.githubusercontent.com/usnistgov/oscal-content/main/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_catalog.json
- Unpinned (main).

aaronlippold (approved):

PR #8469, opened 2026-07-30
FIPS 140-3 validated password hashing — ADR, packaging, and build pipeline

aaronlippold:
## Summary

Design and groundwork for replacing `bcryptjs` with PBKDF2 executed inside a FIPS 140-3 **validated module**, so Heimdall can be deployed on STIG-hardened and FedRAMP-scoped systems.

**This PR contains no application code changes.** It is the ADR, the packaging move, and the build/distribution pipeline that the implementation depends on. Implementation is tracked as epic `heimdall2-e25`.

> **Draft — nothing here is build-verified.** The RPM has never been built from its new home and the CI workflow has never run. See *Known gaps* below.

## Why

`bcryptjs` is pure JavaScript. It never calls `node:crypto` or OpenSSL, so on a FIPS-enabled host it runs **outside the validated module, undetected and unblocked**. The OS cannot see it.

The exposure is narrower than it first appears, and the ADR is careful about this:

- **V-222542** (CAT I) requires salted iterated hashing and prohibits MD5. It never mentions FIPS validation — **bcrypt already satisfies it.**
- **V-222571** and **V-222572** (both CAT II) are what we actually fail.

So: two CAT II findings, not a CAT I failure. Worth fixing; not worth overstating to an assessor.

### 1. ADR-006 — `docs/adr-006-fips-validated-password-hashing.md`

The substance of this PR. Read it first; everything else follows from it.

Every normative claim is marked **[V]** verified against a primary source or **[U]** unverified. That discipline exists for a reason documented in the ADR header: an earlier review pass produced confident, well-formatted citations — STIG check text, CMVP guidance, FedRAMP rule IDs — that had **never been read**. The reviewer retracted them, and on retraction found its own central argument ran backwards.

Corrections that survived independent verification:

- **`--force-fips` is wrong on RHEL.** Red Hat's own Node prints *"Using options related to FIPS is not recommended, configure FIPS in openssl instead."* Verified by running the UBI9 image. This deleted three cards' worth of design.
- The IG block quote is **2.4.C**, not 2.4.A — and reading it properly, it *supports* the design rather than blocking it.
- **V-222571 quoted in full.** The earlier draft quoted 172 of 1,238 characters and called it verbatim, omitting a risk-acceptance path and an NSS clause that makes `PASSWORD_HASH_ALGORITHM=sha256` itself a finding on NSS.
- **IA-7 dropped** — it governs authenticating *to* a cryptographic module, not password verification. **V-16793 dropped** — retired in 2016. **Zeroization dropped** — IG 9.6.A explicitly exempts hashed passwords.
- Both cited CMVP certificates were wrong. #4985 is the OpenSSL Project's module, not Red Hat's; #4754 is libgcrypt and Historical. Correct is **#4857**.
- Measured performance **inverts** the earlier risk rating: `bcryptjs` cost 14 is **1120 ms**, so this is a 7.7× improvement, not a regression. The real cost — a libuv threadpool ceiling — was undocumented.

### 2. Repository hygiene

- `libs/hdf-converters/LICENSE.md` was a **fragment** — no copyright line, no case number, no Apache-2.0 grant, and it referenced "the following Notice" while containing none.
- **Two published npm packages shipped a `license` field with no license text**: `@mitre/heimdall-lite` and `inspecjs`.
- All 8 workspaces now declare a license; all 3 published packages ship it. Private workspaces deliberately do **not** get npm-discovery metadata — Babel, Jest, and Vue all keep theirs minimal.
- `SECURITY.md` and `CONTRIBUTING.md` added (OpenSSF Scorecard Security-Policy gap), written for this stack rather than templated.

### 3. RPM packaging → `packaging/rpm/`

Moved in from `mitre/saf-packaging`. The spec's `%files` tracks build output and `Version:` tracks the app version, so the two must change together.

The import left **four references** pointing at saf-packaging's sibling layout; none resolved here. Repaired:

- `fetch-source.sh` **not ported** — the source is two directories up now. `git archive` of the tag matching `VERSION`, which also keeps airgapped builds working without a mirror.
- `heimdall-cli` is a separate repo now, cloned at a pinned **`HEIMDALL_CLI_REF`** — the contract carrier recording exactly which CLI a given RPM shipped.
- Version resolution moved off `rpmspec` (RHEL-only) to POSIX `sed`, with **drift detection** between the `VERSION` file and the spec. That drift is not hypothetical: the copy on `feat/rpm-build` sat at 2.12.6 while the shipped one tracked 2.13.1, and nothing caught it.
- Fixed a real bug: `CLI_COMMIT` was `git rev-parse HEAD` in the *packaging* repo, so `heimdall-cli --version` reported a heimdall2 commit as the CLI commit.

### 4. CI — `.github/workflows/build-rpm.yml`

el8/el9 × x86_64/aarch64 on **native** runners (no QEMU), install smoke test in a clean container, release artifacts with build provenance. Structure adapted from `redis/memtier_benchmark`.

heimdall2 has never built an RPM in CI, and `v2.13.1` ships zero downloadable assets.

## Related

- **https://github.com/mitre/heimdall-cli** — extracted to its own repo. `go.mod` already declared `module github.com/mitre/heimdall-cli`, so `go install` was broken while it lived inside saf-packaging.
- **COPR `mitresaf/saf`** (ID 249476) — build farm, `epel-8/9` × `x86_64/aarch64`.

## Known gaps

| Gap | Why it matters |
|---|---|
| **Nothing is build-verified** | The RPM has never built from its new home; the workflow has never run. This gates deleting saf-packaging's copies. |
| **`saf.repo` / `heimdall-server.repo` are wrong** | They disagree about the GPG key and reference a namespace that never existed. Blocked until COPR's first build generates a key. |
| **`heimdall-cli` writes bcrypt** | Ships in lockstep — a FIPS-gated server would refuse credentials from the break-glass tool. |
| **§15 has open `[U]` items** | Whether a container on a FIPS host activates the provider without its own `fipsmodule.cnf` needs a real FIPS host. |

## Decisions that aren't mine

- **GPG signing.** COPR signs with its own per-project key. Whether MITRE also wants detached signatures is a security-team call.
- **Package manager.** Yarn 1.22.22 is end-of-life. Tracked separately as `heimdall2-30c.4` — Berry vs pnpm vs npm workspaces, not a foregone conclusion.

## Review notes

Start with the ADR. The commits are grouped: ADR (4), repo hygiene (6), RPM packaging (3), CI (1), plus a `NOTICE.md` add-and-revert whose revert message documents that the notice was consolidated into `LICENSE.md` — worth keeping in history.

Authored by: Aaron Lippold<lippold@gmail.com>


wdower:
**Review: ADR-006 accuracy check — approving, with corrections to fold in**

I did an independent verification pass on the ADR's `[V]` claims — against this repo at `2e1649c9e`, the cited external repos, the ASD STIG (V6R4), the CMVP certificate records, and the NIST/FedRAMP source documents. The core argument chain holds up: the STIG-scoped (not FIPS-140-3-scoped) framing of the bcrypt problem, the IG 2.4.C reasoning for PBKDF2 hash-and-compare, the full V-222571 check text, the RHEL host-inherits-FIPS model, and every in-repo call-site/schema/concurrency claim I checked are accurate as written. The §6 JavaScript traps all reproduce, and `bcryptjs.compare()` on a PHC-format hash does resolve `false` without throwing. I'm approving on that basis. The following should be corrected before this is used as an assessor-facing reference:

**1. Wrong module version string, and the SSP guidance built on it (§15 / Certificates) — the one that matters.** The ADR states the UBI9 provider self-identifies as `3.0.7-cda111b5812c30d4` and that this "is the module name and version an SSP must cite." Cert #4857's validated version is **`3.0.7-395c1a240fbfffd8`** (stated on both the certificate page and the security policy cover); the `cda111b…` string appears in no CMVP record. This isn't a pinning mistake we can fix in the Dockerfile — it's structural: Red Hat validated one specific openssl build and has shipped security errata since, so any current UBI image carries a newer (non-validated) build of the same module, essentially always. The remedy is documentation, not pinning: the SSP should cite cert #4857 and its validated version, disclose the deployed build as a Red Hat maintenance build of that module, self-affirm the operational environment per CMVP Management Manual §7.9 (Level 1), and lean on the FedRAMP Cryptographic Module policy — the same document the ADR already cites for FRR8 — which directs CSPs to prioritize security patching over staying on the frozen validated binary. Pinning the literally-validated openssl build is the strictly-worse alternative (forgoes every CVE fix since validation) and would contradict the ADR's own security-over-compliance tiebreaker. §15 should be rewritten accordingly, and the whole posture verified on the planned FIPS-host trip alongside the existing `[U]` items — in particular whether the containerized provider activates at all without its own `fipsmodule.cnf`, which is upstream of any version-citation question.

**2. bcryptjs does call `node:crypto` (Context).** bcryptjs v3 imports `crypto` and uses `crypto.randomBytes()` for salt generation. The claim should be narrowed to "its *hash computation* is pure JavaScript" — the conclusion (hash generation outside the validated module, V-222571 fires) is unchanged, but the sentence fails the document's own precision standard.

**3. heimdall-cli does not read the `PASSWORD_*` env vars (§9).** The Go CLI hardcodes its password rules; only the retired Python predecessor (in saf-packaging history) read `PASSWORD_MIN_LENGTH`/`PASSWORD_REQUIRE_CLASSES`/`PASSWORD_MAX_CONSECUTIVE`. The §9 "contract" is aspirational — teaching the CLI those vars should join the §14 blocking cross-repo list. (Also noted: the CLI's `make man` target points at a `cmd/gen-manpages` dir that doesn't exist in that repo.)

**4. Keycloak row is mis-attributed (Context table).** The "will not be able to login / ask users to reset" quotes are real but describe **argon2** (Keycloak 25+ default), not bcrypt. The refuse-and-reset pattern the ADR leans on is genuine; the algorithm is wrong.

**5. Mattermost row is stale.** As of Mattermost v11 it defaults to PBKDF2-HMAC-SHA256 @ 600k with a `requirefips` build tag. The "docs inaccurate / ungated bcrypt" contradiction no longer holds — current Mattermost is now a *supporting* precedent for this design rather than a cautionary one.

**Minor:** OWASP's current PBKDF2-SHA512 figure is 220,000 (not 210,000), and the sheet now explicitly recommends "600,000 or more" for FIPS-140 contexts — strengthens the parameter choice. ASD V6R4's revision date is 2025-09-09 (not 2025-10-01); V-222542 currently lists only CCI-004062. Cert #4746 went Historical on 2026-07-30, consistent with the ADR's warning. rpm#1877 documents rpm 4.17/openSUSE, not RHEL 9/rpm 4.16 (behavior claim fine, attribution loose). The COPR project also has s390x chroots and uses the Pulp backend, where retention is "5 most recent successful builds," not the 14-day rule — GitHub-Releases-as-archive still follows. The 140sp4857 policy requires "a *portion* of the salt" ≥128 bits from the DRBG — slightly weaker than the ADR's phrasing; the design clears it either way.

**Structural, non-blocking:** the recurring "the prior draft claimed X" backreferences are confusing for anyone who never read the prior draft — the corrections history should move to an appendix so the document states its current position directly. And the RPM/COPR/distribution material (§14, parts of §12) is separable from the FIPS hashing argument; nothing in the STIG calculus depends on packaging, and splitting it would make both halves easier to review. Finally, "carded separately" for the JWT/V-222570 finding should be verified as real — the referenced epic isn't resolvable from this repo, and this ADR is currently the only written record of that finding.

PR #8510, opened 2026-08-10
Adjust Tenable integration by requiring auth and validating host URLs

DMedina6:
- Add JwtAuthGuard to TenableController (login + proxy endpoints)

- Modify TENABLE_HOST_URL config option (newline-separated allowlist)

- Reject login requests whose host_url origin isn't in the allowlist

- Restricts host_url to http/https, trims whitespace before validation, rejects (instead of silently stripping) any extra path/query/fragment

- Delete cached tenable credentials on logout and when a different user tries to log into tenable

### DIFF
--- a/apps/backend/.env-example
+++ b/apps/backend/.env-example
@@ -41,6 +41,7 @@ NGINX_HOST=<Templated out as the 'server_name' for the NGINX configuration (no d
 ## External interfaces 
 SPLUNK_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Splunk host (no default, must be set if connecting to Splunk)>
 TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable)>
+TENABLE_HOST_URLS=<Comma-separated allowlist of full Tenable.SC host URLs permitted for login (e.g. https://a.example.com,https://b.example.com). Client-supplied host_url values not in this list are rejected>
 
 # Authentication
 
--- a/apps/backend/.env-example
+++ b/apps/backend/.env-example
@@ -40,8 +40,7 @@ NGINX_HOST=<Templated out as the 'server_name' for the NGINX configuration (no d
 
 ## External interfaces 
 SPLUNK_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Splunk host (no default, must be set if connecting to Splunk)>
-TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable)>
-TENABLE_HOST_URLS=<Comma-separated allowlist of full Tenable.SC host URLs permitted for login (e.g. https://a.example.com,https://b.example.com). Client-supplied host_url values not in this list are rejected>
+TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable). For multiple allowed hosts, separate with newlines (not commas, since commas are valid URI characters), e.g. TENABLE_HOST_URL='https://a.example.com\nhttps://b.example.com'. Client-supplied host_url values not matching an entry in this list are rejected>
 
 # Authentication
 
--- a/apps/backend/.env-example
+++ b/apps/backend/.env-example
@@ -40,7 +40,12 @@ NGINX_HOST=<Templated out as the 'server_name' for the NGINX configuration (no d
 
 ## External interfaces 
 SPLUNK_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Splunk host (no default, must be set if connecting to Splunk)>
-TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable). For multiple allowed hosts, separate with newlines (not commas, since commas are valid URI characters), e.g. TENABLE_HOST_URL='https://a.example.com\nhttps://b.example.com'. Client-supplied host_url values not matching an entry in this list are rejected>
+TENABLE_HOST_URL=<Full URL for the Tenable.SC host, including port (defaults to :443 if the client doesn't specify one). For multiple allowed hosts, wrap the value in quotes and separate with newlines, e.g.
+```
+TENABLE_HOST_URL='https://a.example.com
+https://b.example.com:8443'
+```
+Client-supplied host_url values not matching an entry in this list are rejected>
 
 # Authentication
 
--- a/apps/backend/config/app_config.ts
+++ b/apps/backend/config/app_config.ts
@@ -52,9 +52,9 @@ export default class AppConfig {
     }
   }
 
-  // Newline-separated allowlist of Tenable.SC hosts permitted for the login/proxy endpoints.
-  // Newlines are used instead of commas since commas are valid, unencoded URI characters.
-  // A single host works the same as before since there's just one entry to split.
+  // Newline-separated allowlist of Tenable.SC hosts for the login/proxy endpoints
+  // (commas are valid, unencoded URI characters, so they can't be the delimiter).
+  // Entries must include a protocol; malformed entries throw at startup.
   getTenableHostUrl(): string[] {
     const tenable_host_url = this.get('TENABLE_HOST_URL');
     if (tenable_host_url === undefined) {
@@ -63,7 +63,23 @@ export default class AppConfig {
     return tenable_host_url
       .split(/\r?\n/)
       .map((url) => url.trim())
-      .filter((url) => url.length > 0);
+      .filter((url) => url.length > 0)
+      .map((url) => {
+        let parsed: URL;
+        try {
+          parsed = new URL(url);
+        } catch {
+          throw new Error(
+            `Invalid TENABLE_HOST_URL entry "${url}": must be a complete URL including protocol (e.g. https://example.com)`
+          );
+        }
+        if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+          throw new Error(
+            `Invalid TENABLE_HOST_URL entry "${url}": protocol must be http or https`
+          );
+        }
+        return parsed.origin;
+      });
   }
 
   getDatabaseName(): string {
--- a/apps/backend/config/app_config.ts
+++ b/apps/backend/config/app_config.ts
@@ -1,5 +1,6 @@
 import * as dotenv from 'dotenv';
 import * as fs from 'fs';
+import {parseHostUrl} from '../src/utils/url_validation';
 
 export default class AppConfig {
   private envConfig: {[key: string]: string | undefined};
@@ -64,28 +65,10 @@ export default class AppConfig {
       .map((url) => url.trim())
       .filter((url) => url.length > 0)
       .map((url) => {
-        let parsed: URL;
-        try {
-          parsed = new URL(url);
-        } catch {
+        const parsed = parseHostUrl(url);
+        if (!parsed) {
           throw new Error(
-            `Invalid TENABLE_HOST_URL entry "${url}": must be a complete URL including protocol (e.g. https://example.com)`
-          );
-        }
-        if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
-          throw new Error(
-            `Invalid TENABLE_HOST_URL entry "${url}": protocol must be http or https`
-          );
-        }
-        // Only protocol + hostname + port are allowed; ports are permitted since
-        // Tenable.SC may run on a non-default port.
-        const hasExtra =
-          (parsed.pathname !== '' && parsed.pathname !== '/') ||
-          parsed.search !== '' ||
-          parsed.hash !== '';
-        if (hasExtra) {
-          throw new Error(
-            `Invalid TENABLE_HOST_URL entry "${url}": must contain only a protocol, hostname, and optional port (no path, query, or fragment)`
+            `Invalid TENABLE_HOST_URL entry "${url}": must be a complete http(s) URL containing only protocol, hostname, and optional port`
           );
         }
         return parsed.origin;
--- a/apps/backend/config/app_config.ts
+++ b/apps/backend/config/app_config.ts
@@ -61,6 +61,19 @@ export default class AppConfig {
     }
   }
 
+  // Comma-separated allowlist of Tenable.SC hosts permitted for the login/proxy endpoints.
+  getTenableHostUrls(): string[] {
+    const tenable_host_urls = this.get('TENABLE_HOST_URLS');
+    if (tenable_host_urls !== undefined) {
+      return tenable_host_urls
+        .split(',')
+        .map((url) => url.trim())
+        .filter((url) => url.length > 0);
+    } else {
+      return [];
+    }
+  }
+
   getDatabaseName(): string {
     const databaseName = this.get('DATABASE_NAME');
     const nodeEnvironment = this.get('NODE_ENV');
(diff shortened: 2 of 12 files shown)


apps/backend/.env-example
```diff
@@ -41,6 +41,7 @@ NGINX_HOST=<Templated out as the 'server_name' for the NGINX configuration (no d
 ## External interfaces 
 SPLUNK_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Splunk host (no default, must be set if connecting to Splunk)>
 TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable)>
+TENABLE_HOST_URLS=<Comma-separated allowlist of full Tenable.SC host URLs permitted for login (e.g. https://a.example.com,https://b.example.com). Client-supplied host_url values not in this list are rejected>
```
Amndeep7: commas are potentially allowed characters in uri so i think we should not use it as a divider.  instead i think we can use newlines, ex.

envvar='https://blah
http://127.0.0.1
https://other.place'

Amndeep7: instead of having multiple variables, i think we should just override the tenable_host_url variable so handle both the single and multiple case.  this'll probably require us changing the type of that from string to string[] and corresponding changes in a few other places.

apps/backend/config/app_config.ts
```diff
@@ -61,6 +61,23 @@ export default class AppConfig {
     }
   }
 
+  // Comma-separated allowlist of Tenable.SC hosts permitted for the login/proxy endpoints.
```
Amndeep7: like i said elsewhere, don't do comma separated, i think this should be newline separated.  also we should maybe expand the validation that we do so that it's not just confirming it's not the empty string but also maybe use that validator library we have somewhere or other to confirm that it is in fact a url with the protocol and everything

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,66 @@ const TENABLE_CSP_NOT_SET =
...
+  ) {}
+
+  // Returns the matching allowlisted origin for host_url, or null if none match
+  // or if host_url carries a path/query/fragment beyond a bare origin.
+  private resolveAllowedHostUrl(host_url: string): string | null {
+    const allowlist = this.configService.getTenableHostUrls();
+    if (allowlist.length === 0) {
+      return null;
+    }
+    // Default to https when no scheme is provided, so "example.com" is treated
```
Amndeep7: i think it is reasonable to expect the protocol to be provided by the user, if it is not, then it is invalid for our use.  this goes back to what i said elsewhere about doing validation at the config service level where it'll throw out an error about this.

to be clear, what i mean by this is that we should not normalize it in this manner at all aside from doing the trimming, which in and of itself should also probably be happening at the configservice level anyways

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,66 @@ const TENABLE_CSP_NOT_SET =
...
+    try {
+      // Throws for unparseable input, caught below and treated as not allowed.
+      const parsed = new URL(normalizedInput);
+
+      // Reject any scheme other than http/https (e.g. file:, javascript:, ftp:).
+      if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+        return null;
+      }
+
+      // Reject anything beyond a bare origin rather than silently stripping it,
```
Amndeep7: again, feels like this validation should happen in the config area

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,66 @@ const TENABLE_CSP_NOT_SET =
...
+        (parsed.pathname !== '' && parsed.pathname !== '/') ||
+        parsed.search !== '' ||
+        parsed.hash !== '';
+      if (hasExtra) {
+        return null;
+      }
+
+      // Compare origins (scheme + host + port) rather than raw strings so that
+      // equivalent URLs (default ports, trailing slash, casing) still match.
+      const match = allowlist.find((allowed) => {
```
Amndeep7: fix the sonarqube finding here re: some over find

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -52,17 +113,29 @@ export class TenableController {
       throw new HttpException('Missing credentials', HttpStatus.BAD_REQUEST);
     }
 
+    const allowedHostUrl = this.resolveAllowedHostUrl(host_url);
+    if (!allowedHostUrl) {
+      throw new HttpException(
```
Amndeep7, line 90: question: does this show up properly in the snackbar as an error for the user?  esp since sometimes the error like flashes for a second and then swaps to a different notification

Amndeep7 (changes requested):

problem-context.md
Amndeep7, line 1: delete this

apps/backend/.env-example
```diff
@@ -40,7 +40,7 @@ NGINX_HOST=<Templated out as the 'server_name' for the NGINX configuration (no d
 
 ## External interfaces 
 SPLUNK_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Splunk host (no default, must be set if connecting to Splunk)>
-TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable)>
+TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable). For multiple allowed hosts, separate with newlines (not commas, since commas are valid URI characters), e.g. TENABLE_HOST_URL='https://a.example.com\nhttps://b.example.com'. Client-supplied host_url values not matching an entry in this list are rejected>
```
Amndeep7: don't need the line about commas.  if you use the quotes you can just put an actual newline in there instead of the escaped character so that it's easier to read esp for non-technical folks.  maybe something like the below.

```suggestion
TENABLE_HOST_URL=<The full Uniform Resource Locator (URL) without the port for the Tenable.SC host (no default, must be set if connecting to Tenable). For multiple allowed hosts, wrap the value in quotes and separate with newlines,  e.g. `TENABLE_HOST_URL='https://a.example.com
https://b.example.com'.` Client-supplied host_url values not matching an entry in this list are rejected.>
```

apps/backend/.env-example
Amndeep7, line 1: we'll also need to update the envvar wiki page

Amndeep7, line 1: we'll also probably need to update the heimdall helm chart

apps/backend/config/app_config.ts
```diff
@@ -52,13 +52,34 @@ export default class AppConfig {
     }
   }
 
-  getTenableHostUrl(): string {
+  // Newline-separated allowlist of Tenable.SC hosts for the login/proxy endpoints
+  // (commas are valid, unencoded URI characters, so they can't be the delimiter).
```
Amndeep7: get rid of the line re: commas

apps/backend/config/app_config.ts
```diff
@@ -52,13 +52,34 @@ export default class AppConfig {
...
     const tenable_host_url = this.get('TENABLE_HOST_URL');
-    if (tenable_host_url !== undefined) {
-      return tenable_host_url;
-    } else {
-      return '';
+    if (tenable_host_url === undefined) {
+      return [];
     }
+    return tenable_host_url
+      .split(/\r?\n/)
```
Amndeep7: you should be able to split on just '\n' and let the trim handle the '\r'

apps/backend/config/app_config.ts
```diff
@@ -52,13 +52,34 @@ export default class AppConfig {
...
+          throw new Error(
+            `Invalid TENABLE_HOST_URL entry "${url}": must be a complete URL including protocol (e.g. https://example.com)`
+          );
+        }
+        if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+          throw new Error(
+            `Invalid TENABLE_HOST_URL entry "${url}": protocol must be http or https`
+          );
+        }
+        return parsed.origin;
```
Amndeep7, line 74: since this is generating strings that are going in the cors thing we should probably make sure that the provided urls only consist of the protocol and the hostname.  no ports, no query strings, etc.

apps/backend/src/tenable/tenable.controller.ts
DMedina6: Yep 
image

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,43 @@ const TENABLE_CSP_NOT_SET =
...
+      }
+
+      // Reject extra path/query/fragment
+      const hasExtra =
+        (parsed.pathname !== '' && parsed.pathname !== '/') ||
+        parsed.search !== '' ||
+        parsed.hash !== '';
+      if (hasExtra) {
+        return null;
+      }
```
Amndeep7: similar (but i think slightly different) to above we should make sure that this only allows for protocol, hostname, and port, but no other parameters.  for ex. this will allow username/password to go through.

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,43 @@ const TENABLE_CSP_NOT_SET =
...
+  // Resolves host_url to its allowlisted origin, or null if not allowed
+  private resolveAllowedHostUrl(host_url: string): string | null {
+    const allowlist = this.configService.getTenableHostUrl();
+    if (allowlist.length === 0) {
+      return null;
+    }
+    try {
+      const parsed = new URL(host_url); // requires a protocol; throws otherwise
+
+      if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
```
Amndeep7: add a comment to explain why we reject not http/https

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,43 @@ const TENABLE_CSP_NOT_SET =
...
+      // Reject extra path/query/fragment
+      const hasExtra =
+        (parsed.pathname !== '' && parsed.pathname !== '/') ||
+        parsed.search !== '' ||
+        parsed.hash !== '';
+      if (hasExtra) {
+        return null;
+      }
+
+      const match = allowlist.includes(parsed.origin);
```
Amndeep7: add a comment along the lines of restricting to the allowlist helps prevent ssrf

Amndeep7: nominally this will fail if they are using a nonstandard port for the protocol which is allowed, i.e. we add 443 as a port by default for https but they could use basically whatever else and it would be valid but would crash here cause "For URLs using the ftp:, http:, https:, ws:, and wss: schemes, the [protocol](https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol) followed by //, followed by the [host](https://developer.mozilla.org/en-US/docs/Web/API/URL/host). **Same as host, the [port](https://developer.mozilla.org/en-US/docs/Web/API/URL/port) is only included if it's not the default for the protocol.**" - https://developer.mozilla.org/en-US/docs/Web/API/URL/origin

i'm pretty sure the envvar says to not include a port on it

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -30,8 +33,43 @@ const TENABLE_CSP_NOT_SET =
...
+
+      // Only the parsed origin (never the raw input) is used downstream
```
Amndeep7: i feel like this comment is misleading as the 'parsed origin' is directly derived from the raw input

apps/frontend/src/utilities/tenable_util.ts
```diff
@@ -140,7 +140,7 @@ export class TenableUtil {
     const DEFAULT_REJECT_MSG =
       `${error.name}: ${(error.response?.data?.message ?? error.message)}, 
        ${error.response?.data?.code ?? error.code}`;
-    const TENABLE_HOST_URL = ServerModule.tenableHostUrl;
+    const TENABLE_HOST_URLS = ServerModule.tenableHostUrl;
```
Amndeep7: can probably leave the variable as is unless you think it's better to make this like the one place where we refer to it in plural now

Amndeep7 (changes requested):

apps/backend/.env-example
Amndeep7, line 1: we'll also probably need to update the tenable integration part of the integrations wiki page

Amndeep7 (changes requested):

apps/backend/src/tenable/tenable.controller.ts
```diff
@@ -52,114 +90,132 @@ export class TenableController {
...
       const result = await axios.get(fullUrl, {
         headers: {
           'x-apikey': `accesskey=${accesskey}; secretkey=${secretkey}`
         }
       });
 
-      // Assign the Tenable credentials to the session
-      req.session.tenable = {host_url, accesskey, secretkey};
+      // Store the normalized, allowlisted origin rather than the raw client value.
+      req.session.tenable = {host_url: allowedHostUrl, accesskey, secretkey};
```
Amndeep7: AI review comment that seems like it could be feasible so please review it:

> JwtAuthGuard proves only that the caller has a Heimdall account; these credentials remain associated solely with the browser’s Express session. The Tenable “sign out” UI only clears client state, and /users/logout revokes the JWT without clearing session.tenable. On a shared/reused browser session, a subsequently authenticated user can call the proxy with the prior user’s Tenable keys. Store the authenticated Heimdall user ID with the credentials, verify it on every proxy request, and clear the credentials on both Tenable and Heimdall logout. Please add an A-login → logout → B-login → proxy-is-denied regression test.

You don't need to write an explicit test unless you want to, but please do validate by hand that logging in with a second account does not allow one to reuse the access/secret key.

Amndeep7 (changes requested):

apps/backend/src/tenable/tenable.controller.ts
DMedina6: Ports / non-default ports can now be specified for both the environment variable / app-config, and the frontend / client side input so it's consistent and flexible

apps/backend/src/tenable/tenable.controller.ts
DMedina6: I thought about just removing the comment, but decided instead to just directly use the entry from the env file allow list

apps/backend/src/tenable/tenable.controller.ts
Amndeep7: huh i thought that the csp directives didn't allow ports but i guess they do: 

https://github.com/mitre/heimdall2/blob/master/apps/backend/src/main.ts#L57

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/connect-src#source-expression-list

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#host-source

"The scheme, port number, and path are optional."

so guess this change is fine

apps/backend/config/app_config.ts
```diff
@@ -52,13 +52,44 @@ export default class AppConfig {
...
+          );
+        }
+        if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+          throw new Error(
+            `Invalid TENABLE_HOST_URL entry "${url}": protocol must be http or https`
+          );
+        }
+        // Only protocol + hostname + port are allowed; ports are permitted since
+        // Tenable.SC may run on a non-default port.
+        const hasExtra =
```
Amndeep7: both of the hasExtra processing areas need to exclude everything aside from hostname + protocol + optionally port

considering that this processing code is the same, maybe we can extract out to some common utility function and use it in both places.

Amndeep7 (changes requested):

Amndeep7 (approved):

PR #8573, opened 2026-08-23
feat(hdf-converters): add KICS JSON to HDF mapper

### LINKED ISSUE #8572: Add a KICS mapper (native JSON, not SARIF)
## Summary

There is no KICS converter in hdf-libs, heimdall2 or the SAF CLI today. KICS
(Checkmarx, Apache-2.0) is a widely used IaC scanner covering Terraform,
CloudFormation, ARM, Kubernetes, Ansible, Docker, CI/CD definitions and more —
1,811 queries in the current image.

KICS emits SARIF, so `sarif2hdf` is a partial path. It is a lossy one, and the
native JSON carries materially more.

## Measured fidelity gap

Scanned a real Terraform tree with `checkmarx/kics:latest`, emitting JSON and
SARIF from the same run: **849 findings across 72 queries**.

What SARIF preserves — and this is better than most scanners manage — is the CWE
taxonomy: **72 of 72 rules** carry a proper `relationships[].target` against a
`CWE` toolComponent. Rule name, description, `helpUri` and `riskScore` also
survive.

What SARIF drops:

| Native field | Fate in SARIF | Why it matters |
| --- | --- | --- |
| `expected_value` | dropped on all 849 findings | the message keeps only `actual_value`, so you get what it *is* and never what it *should be* |
| `issue_type` | dropped | `MissingAttribute` 767 / `IncorrectValue` 68 / `RedundantAttribute` 14 — add-a-block versus fix-a-value |
| `similarity_id` | dropped | KICS's own stable finding fingerprint, the natural cross-run dedup key |
| `resource_name`, `search_value`, `description_id` | dropped | |
| severity | coarsened 5 → 4 | `CRITICAL` and `HIGH` both collapse to `error` |

`resource_type` survives only as prose inside the message string, not as a
structured field, so it cannot be filtered on.

## Why this is a cheap converter to build

Unlike most scanners, KICS ships its own catalog as data: 1,811
`metadata.json` files under `/app/bin/assets/queries/`, each with `id`,
`queryName`, `severity`, `category`, `descriptionText`, `descriptionUrl`,
`platform`, `cloudProvider`, `cwe` and `riskScore`, with the rule's `query.rego`
beside it. Nothing has to be reverse-engineered or synthesized to produce
profile-grade output.

Every query carries a `cwe`, so `tags.nist` resolves through the existing CWE
mapping with no new mapping table.

## Proposed mapping

| HDF field | Source |
| --- | --- |
| control / requirement id | `query_id` (stable UUID) |
| title | `query_name` |
| desc | `description` |
| impact | `severity`, five levels — see note below |
| `tags.nist` / `tags.cci` | derived from `cwe` |
| `tags` | `platform`, `cloud_provider`, `category`, `risk_score`, `issue_type`, `resource_type` |
| refs | `query_url` |
| `code_desc` | `file_name`, `line`, `resource_type`, `search_key` |
| `results[].message` | `expected_value` vs `actual_value` — the remediation pair |
| passthrough | scan counters, `files_scanned`, `queries_total`, `kics_version` |

Granularity: one control per query, with each occurrence a result. KICS already
groups its output that way — `queries[].files[]` — so it maps directly.

## Gotchas found while measuring

- `oldSeverity` appears alongside `severity` in the catalog; KICS is migrating
  its scale. A converter must read `severity` and ignore `oldSeverity`, or
  findings silently downgrade.
- Five native severities (`CRITICAL`/`HIGH`/`MEDIUM`/`LOW`/`INFO`) need five
  impact values. Mapping `INFO` to `0.0` would report Not Applicable and drop
  those findings from the compliance score.
- The same query id appears across platforms with different `platform` and
  `cloud_provider`; resolution should not assume `query_id` alone is unique per
  scan context.

## Fixtures

Reproducible from a small synthetic Terraform tree — no proprietary input
required. Happy to contribute the converter along with fixtures covering
findings across severities, a scan reporting parse failures, and the
zero-findings case.

## Scope in this repo

`libs/hdf-converters/src/kics-mapper.ts` built on `BaseConverter`, exported from
`index.ts`, `INPUT_TYPES.KICS` plus fingerprint entry in
`src/utils/fingerprinting.ts`, a case in `apps/frontend/src/store/report_intake.ts`,
sample data under `sample_jsons/kics_mapper/`, a spec, and a line in the
supported-formats README list.

Detection note: KICS JSON is distinctive — `queries`, `severity_counters`,
`scan_id`, `kics_version` — so the fingerprint can key on names no other
converter uses, rather than the generic `results`/`version` pair.

Related: mitre/hdf-libs#238 tracks the same converter for the v3 line.

clem-field:
Closes #8572.

Maps the native `kics scan --report-formats json` format, plus its fingerprint, frontend intake wiring, sample data and spec.

## Why not sarif2hdf

KICS emits SARIF, and unusually its SARIF is *well formed* — in a real scan, 72 of 72 rules carried a proper CWE taxonomy relationship. That is better than most scanners manage.

What SARIF drops is everything about remediation and identity:

| Native field | Fate in SARIF | Consequence |
| --- | --- | --- |
| `expected_value` | dropped on **100%** of findings | the message keeps only `actual_value`, so a control can state what the configuration *is* but never what it *should be* |
| `issue_type` | dropped | MissingAttribute 767 / IncorrectValue 68 / RedundantAttribute 14 in the scan measured — add-a-block versus fix-a-value |
| `similarity_id` | dropped | KICS's own stable finding fingerprint |
| `resource_name`, `search_value`, `description_id` | dropped | |
| severity | 5 levels to 4 | CRITICAL and HIGH both become `error` |

## Design notes

**One control per query**, each occurrence a result. KICS already groups its output as `queries[].files[]`, so it maps directly.

**No severity maps to impact 0.** `impact 0.0` reports Not Applicable in HDF and would drop the finding from the compliance score rather than rating it low; INFO and TRACE map to 0.1.

**Fingerprint** keys on `kics_version` and `severity_counters` alongside `queries`, verified against all 134 sample files in `sample_jsons` with no collision in either direction.

## NIST mapping provenance — worth a look

Controls carry a `nistMapping` tag reading `cwe-derived` or `static-fallback`, so a resolved mapping stays distinguishable from a static default.

The motivation is measurable. KICS carries a CWE on **all 1,811** of its shipped queries across 102 distinct CWEs, but only **30 of those resolve** against `CweNistMappingData` — **52% of queries by volume**. The largest misses are the most-used checks: CWE-778 (211 queries), CWE-665 (113), CWE-710 (111), CWE-668 (78).

Without the marker, roughly half of a KICS run carries `SA-11, RA-5` with nothing to indicate it is a fallback rather than a mapping. The source CWE is retained in tags even when it does not resolve, so the gap is visible rather than silent.

Thirteen mappers in this repo resolve `tags.nist` through the same table, so this is likely a broader condition than KICS. Raising it here rather than expanding scope; happy to drop the tag if maintainers would rather not set the precedent in a converter PR.

The equivalent converter for the v3 line is mitre/hdf-libs#240, and mitre/hdf-libs#239 tracks the mapping work this anticipates.

## Fixtures

Generated from a small synthetic Terraform tree, so they are reproducible and carry no proprietary input: findings across four severities, and a genuine zero-findings scan.

## Testing

13 tests covering rule grouping, the severity scale, the provenance marker, retention of unresolved CWEs, the remediation pair, and the zero-findings case.

`vitest run` in `libs/hdf-converters`: 4 failures, all reproducing identically on an unmodified `master` checkout (3 sonarqube tests that reach a live API, 1 splunk reverse mapper).

### DIFF
--- a/apps/frontend/src/store/report_intake.ts
+++ b/apps/frontend/src/store/report_intake.ts
@@ -22,6 +22,7 @@ import {
   INPUT_TYPES,
   IonChannelMapper,
   JfrogXrayMapper,
+  KicsMapper,
   MsftSecureScoreResults,
   NessusResults,
   NetsparkerResults,
@@ -230,6 +231,8 @@ export class InspecIntake extends VuexModule {
     switch (typeGuess) {
       case INPUT_TYPES.JFROG:
         return new JfrogXrayMapper(convertOptions.data).toHdf();
+      case INPUT_TYPES.KICS:
+        return new KicsMapper(convertOptions.data).toHdf();
       case INPUT_TYPES.MSFT_SEC_SCORE:
         return new MsftSecureScoreResults(convertOptions.data).toHdf();
       case INPUT_TYPES.ASFF:
--- a/libs/hdf-converters/README.md
+++ b/libs/hdf-converters/README.md
@@ -21,22 +21,23 @@ OHDF Converters supplies several methods to convert various types of security to
 13. [**gosec-mapper**] - gosec results JSON file
 14. [**ionchannel-mapper**] - SBOM data from Ion Channel
 15. [**jfrog-xray-mapper**] - JFrog Xray results JSON file
-16. [**msft-secure-mapper**] - Microsoft Secure Score results file
-17. [**nessus-mapper**] - Nessus XML results file
-18. [**netsparker-mapper**] - Netsparker XML results file
-19. [**neuvector-mapper**] - NeuVector JSON results file
-20. [**nikto-mapper**] - Nikto results JSON file
-21. [**prisma-mapper**] - Prisma Cloud Scan Report CSV file
-22. [**sarif-mapper**] - SARIF JSON file
-23. [**scoutsuite-mapper**] - ScoutSuite results from a Javascript object
-24. [**snyk-mapper**] - Snyk results JSON file
-25. [**sonarqube-mapper**] - SonarQube vulnerabilities for the specified project name and optional branch or pull/merge request ID name from an API
-26. [**splunk-mapper**] - Splunk instance
-27. [**trufflehog-mapper**] - Trufflehog results json file
-28. [**twistlock-mapper**] - Twistlock CLI output file
-29. [**veracode-mapper**] - Veracode Scan Results XML file
-30. [**xccdf-results-mapper**] - SCAP client XCCDF-Results XML report
-31. [**zap-mapper**] - OWASP ZAP results JSON
+16. [**kics-mapper**] - KICS infrastructure-as-code scan results JSON file
+17. [**msft-secure-mapper**] - Microsoft Secure Score results file
+18. [**nessus-mapper**] - Nessus XML results file
+19. [**netsparker-mapper**] - Netsparker XML results file
+20. [**neuvector-mapper**] - NeuVector JSON results file
+21. [**nikto-mapper**] - Nikto results JSON file
+22. [**prisma-mapper**] - Prisma Cloud Scan Report CSV file
+23. [**sarif-mapper**] - SARIF JSON file
+24. [**scoutsuite-mapper**] - ScoutSuite results from a Javascript object
+25. [**snyk-mapper**] - Snyk results JSON file
+26. [**sonarqube-mapper**] - SonarQube vulnerabilities for the specified project name and optional branch or pull/merge request ID name from an API
+27. [**splunk-mapper**] - Splunk instance
+28. [**trufflehog-mapper**] - Trufflehog results json file
+29. [**twistlock-mapper**] - Twistlock CLI output file
+30. [**veracode-mapper**] - Veracode Scan Results XML file
+31. [**xccdf-results-mapper**] - SCAP client XCCDF-Results XML report
+32. [**zap-mapper**] - OWASP ZAP results JSON
 
 ### NOTICE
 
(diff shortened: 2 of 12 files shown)


clem-field:
### Added: a scan coverage record

Following review feedback that a KICS profile reads as 100% failed.

KICS reports violations only. Its output has no record of the queries that ran
without finding anything — `queries[]` holds only those that fired — so no
passing requirement can be derived from it, and a converted profile is
failures-only by construction. On a real Terraform scan that means **72 of 2,034
executed queries** appear in the document, and the compliance ratio renders as
100% failed when 1,962 checks ran clean.

The converter now emits a `kics-scan-coverage` requirement carrying the
denominator:

```
queriesExecuted:     2034
queriesWithFindings:   72
filesScanned:         294
filesParsed:          292
```

with a description stating plainly that KICS does not enumerate passing queries
and the ratio should not be read as a pass rate. Impact 0 reports Not
Applicable, matching `buildNoFindingsRequirement`, so the record cannot skew the
ratio it exists to explain.

This is deliberately a *record*, not synthesised passes. KICS tells us how many
queries ran but not which ones, so naming them would be invention. Reconstructing
the passing set is possible in principle — the shipped catalog enumerates all
1,811 queries, so a converter could diff findings against the queries applicable
to the platforms scanned — but that needs the catalog alongside the converter and
is the same machinery as Not-Applicable seeding. Out of scope here; worth its own
issue if maintainers want it.

Issue #8610, opened 2026-09-01
`impactMapping()` returns 0 for unmapped severities, indistinguishable from a legitimate impact of 0.0


wdower:
`libs/hdf-converters/src/base-converter.ts`:

```ts
export function impactMapping(
  mapping: Map<string, number>
): (severity: unknown) => number {
  return (severity: unknown): number => {
    if (typeof severity === 'string' || typeof severity === 'number') {
      return mapping.get(severity.toString().toLowerCase()) || 0;
    } else {
      return 0;
    }
  };
}
```

A severity value missing from a mapper's `IMPACT_MAPPING` silently produces impact 0 — the same value as a deliberate "informational / not applicable" mapping. Every mapper that uses `impactMapping()` inherits this. Impact 0 has load-bearing semantics downstream (`hdf2ckl` maps impact 0 to `Not_Applicable`; Heimdall buckets it as "No impact"), so a mapping miss doesn't just lose fidelity — it actively reclassifies findings.

Concrete instance with measured production impact: #8611 (`twistlock2hdf`). This issue is for the general behavior.

wdower:
Decision (implemented): `impactMapping()` now takes an optional default value (`0.5`) and warns once per distinct unmapped severity — a miss is loud and can no longer masquerade as a legitimate mapped `0.0`, which stays silent (the lookup now distinguishes "mapped to 0" from "not in the map" instead of using `|| 0`).

Per-mapper severity additions (e.g. the Twistlock values from #8611) are handled on their own issues.

Issue #8611, opened 2026-09-01
`twistlock2hdf`: unmapped severities silently become impact 0, and hdf2ckl then marks those findings Not_Applicable


wdower:
Twistlock scan output can contain severities that are missing from `IMPACT_MAPPING` in `libs/hdf-converters/src/twistlock-mapper.ts`:

```ts
const IMPACT_MAPPING: Map<string, number> = new Map([
  ['critical', 0.9],
  ['important', 0.9],
  ['high', 0.7],
  ['medium', 0.5],
  ['moderate', 0.5],
  ['low', 0.3]
]);
```

Real Twistlock reports also emit at least `unassigned`, `unimportant`, and `negligible`. None are in the map, so `impactMapping()` in `base-converter.ts` falls through to its `|| 0` default and the finding is written into OHDF with impact 0.0 — with no warning, and indistinguishable from a genuine 0.0.

**Reproduce:** convert any Twistlock report containing a finding with `"severity": "unassigned"`, then run the result through `hdf2ckl`. The finding's stanza has `STATUS: Not_Applicable`.

**Suggested fix:** add the missing severities to `IMPACT_MAPPING` (`unassigned` and `unimportant` presumably map somewhere in the 0.3–0.5 band; `negligible` near the floor — needs a decision). The silent-miss behavior of `impactMapping()` itself is a separate, converter-wide issue: see companion issue #8610.

The reporters have offered matched sample reports from their corpus as regression fixtures.

wdower:
We're going to go with: `unassigned` / `not yet assigned` / `untriaged` → 0.5, `unimportant` → 0.1, `negligible` → 0.1, based on Prisma's docs on how it aligns upstream vendor severity levels into Twistlock. We will update the mapper to properly catch those severities. 

I also want to add a fallthrough guard so that even if an unmapped severity shows up in source data, it gets set as 0.5 impact instead of 0.0 (on the logic that 0.5 is the average and therefore the least opinionated impact level). Lots of converters already map unknown statuses to 0.5. See #8610.

PR #8617, opened 2026-09-03
fix(hdf-converters): warn on unmapped severities instead of silently mapping to impact 0

wdower:
impactMapping() now distinguishes a mapping miss from a legitimate mapped 0.0: misses warn once per distinct value and default to a conservative 0.5 (caller-overridable) instead of returning 0, which hdf2ckl exported as Not_Applicable — silently removing open findings from checklists. Mirrors hdf-libs' severityToImpactWithAliases convention.

Snyk expected-output fixtures updated: its map lacks 'critical', so those findings move from silent 0 to warned 0.5 (proper 0.9 mapping tracked separately).

Root-cause fix for #8610.


libs/hdf-converters/src/base-converter.ts
```diff
@@ -96,14 +96,26 @@ export function parseCsv(csv: string): unknown[] {
 }
 
 export function impactMapping(
-  mapping: Map<string, number>
+  mapping: Map<string, number>,
+  defaultValue = 0.5
 ): (severity: unknown) => number {
+  const warned = new Set<string>();
   return (severity: unknown): number => {
     if (typeof severity === 'string' || typeof severity === 'number') {
```
Amndeep7, line 104: i would replace this to better use lodash's more comprehensive checks

libs/hdf-converters/src/base-converter.ts
```diff
@@ -96,14 +96,26 @@ export function parseCsv(csv: string): unknown[] {
 }
 
 export function impactMapping(
```
Amndeep7, line 98: not all mappers use the implementation of impactmapping within the base converter.  maybe they should?

furthermore, not all mappers use an implementation of impactmapping at all.  they should be researched to find out what they're doing instead.

Amndeep7 (changes requested):

libs/hdf-converters/src/base-converter.ts
wdower: Tried to answer the question of "which converters actually use this part of base converter."

Audited all 30 converters that set `impact`: 28 extend `BaseConverter`, but only 18 call `impactMapping()`, so this PR's fix doesn't reach the other 12.

Eight of those have nothing for it to reach — no severity vocabulary at all: constants (`nikto`, `trufflehog`, `ionchannel`), numeric scores (`neuvector` `score_v3/10`, `conveyor`, `msft-secure-score`), a compliance boolean (`aws-config`), and pass-through of impact that's already in HDF form (`case-previously-hdf`).

Four do map a severity vocabulary outside the helper:

- **`asff-mapper`** — `IMPACT_MAPPING.get(impact) || 0`, an inline copy of exactly the bug this PR fixes. Still vulnerable.
- **`prisma-mapper`** — `SEVERITY_LOOKUP[severity]` with no fallback at all, so an unmapped severity yields `impact: undefined`, which is worse than 0. Same scanner family as Twistlock, so the vendor severities from #8611 (`unimportant`, `negligible`, `untriaged`, `unassigned`) hit it today.
- **`cyclonedx-sbom-mapper`** — `IMPACT_MAPPING.get(...) ?? 0.5`, commented "defaulting to medium/0.5". Already does what this PR proposes.
- **`ckl-mapper/checklist-mapper`** — throws an Error. Loudest of the four; not vulnerable.

The ones that don't have a fallthrough (and can therefore be bit by this bug) are `asff` and `prisma`, but we might just want to make all of them that do severity mapping use the helper and work consistently. I think we can do that without disturbing the rest of them that have a genuinely different relationship with severity.

Amndeep7: > The ones that don't have a fallthrough (and can therefore be bit by this bug) are asff and prisma, but we might just want to make all of them that do severity mapping use the helper and work consistently. I think we can do that without disturbing the rest of them that have a genuinely different relationship with severity.

i'm ok with this

wdower: Implemented.

Also added a guard to lowercase the severities passed to `impactMapping`, since some of the converters use UPPERCASED and some use lowercased severities and we should just nip that in the bud right here.

PR #8618, opened 2026-09-03
fix(hdf-converters): map Twistlock's distro-vendor pass-through severities

wdower:
For OS-vendor-maintained packages Twistlock/Prisma Cloud reports the vendor's own severity string instead of NVD's, so distro vocabulary reaches the scan report. Closest doc available describing the possible severity levels one can see in the source format is the [vendor table in Prisma's CVSS-scoring doc](https://docs.prismacloud.io/admin-guide/vulnerability-management/cvss-scoring).

So for severity level mappings into OHDF, we're going to go with:
- unassigned / not yet assigned / untriaged -> 0.5 (on the theory that we usually bucket "unknown" severity levels as "average" to split the difference)
- unimportant / negligible -> 0.1 (the lowest ratings that aren't 0.0, which we usually take to mean "N/A")

Adds a (synthetic) fixture carrying all five severities plus a mapped control case, with targeted impact assertions. I am following up with the folks who brought #8611 to our attention to see if we can get a copy of the Twistlock scan that triggered the original error so that we can confirm it works with that one.

With #8610's warn-and-default fix this closes the Twistlock half of the severity data-loss report in #8611.

### DIFF
--- a/libs/hdf-converters/sample_jsons/twistlock_mapper/sample_input_report/twistlock-twistcli-vendor-severities-sample.json
+++ b/libs/hdf-converters/sample_jsons/twistlock_mapper/sample_input_report/twistlock-twistcli-vendor-severities-sample.json
@@ -1,0 +1,211 @@
+{
+  "results": [
+    {
+      "id": "sha256:b1c0237ebd29860066656372da10d8d7da33b34715986f74b3d5a7e4ba060d1b",
+      "name": "registry.io/test:1a7a431a105aa04632f5f5fbe8f753bd245add0a",
+      "distro": "Red Hat Enterprise Linux release 8.6 (Ootpa)",
+      "distroRelease": "RHEL8",
+      "digest": "sha256:c0274cb1e0c6e92d2ccc1e23bd19b7dddedbaa2da26861c225d4ad6cec5047f4",
+      "collections": [
+        "All",
+        "TEST-COLLECTION"
+      ],
+      "packages": [
+        {
+          "type": "os",
+          "name": "expat",
+          "version": "2.2.5-8.el8",
+          "licenses": [
+            "MIT"
+          ]
+        },
+        {
+          "type": "os",
+          "name": "nss-util",
+          "version": "3.67.0-7.el8_5",
+          "licenses": [
+            "MPLv2.0"
+          ]
+        },
+        {
+          "type": "os",
+          "name": "dotnet-host",
+          "version": "6.0.5-1.el8_6",
+          "licenses": [
+            "MIT and ASL 2.0 and BSD and LGPLv2+ and CC-BY and CC0 and MS-PL and EPL-1.0 and GPL+ and GPLv2 and ISC and OFL and zlib"
+          ]
+        }
+      ],
+      "applications": [
+        {
+          "name": "asp.net-core",
+          "version": "5.0.17",
+          "path": "/usr/lib64/dotnet/dotnet"
+        },
+        {
+          "name": ".net-core",
+          "version": "5.0.17",
+          "path": "/usr/lib64/dotnet/dotnet"
+        }
+      ],
+      "complianceDistribution": {
+        "critical": 0,
+        "high": 0,
+        "medium": 0,
+        "low": 0,
+        "total": 0
+      },
+      "complianceScanPassed": true,
+      "vulnerabilities": [
+        {
+          "id": "CVE-2021-43529",
+          "status": "affected",
+          "cvss": 9.8,
+          "description": "DOCUMENTATION: A remote code execution flaw was found in the way NSS verifies certificates. This flaw allows an attacker posing as an SSL/TLS server to trigger this issue in a client application compiled with NSS when it tries to initiate an SSL/TLS connection.  Similarly, a server application compiled with NSS, which processes client certificates, can receive a malicious certificate via a client, triggering the flaw. The highest threat to this vulnerability is confidentiality, integrity, as well as system availability.              STATEMENT: The issue is not limited to TLS. Any applications that use NSS certificate verification are vulnerable; S/MIME is impacted as well.  Similarly, a server application compiled with NSS, which processes client certificates, can receive a malicious certificate via a client.  Firefox is not vulnerable to this flaw as it uses the mozilla::pkix for certificate verification. Thunderbird is affected when parsing email with the S/MIME signature.  Thunderbird on Red Hat Enterprise Linux 8.4 and later does not need to be updated since it uses the system NSS library, but earlier Red Hat Enterprise Linux 8 extended life streams will need to update Thunderbird as well as NSS.             MITIGATION: Red Hat has investigated whether a possible mitigation exists for this issue, and has not been able to identify a practical example. Please update the affec",
+          "severity": "critical",
+          "packageName": "nss-util",
+          "packageVersion": "3.67.0-7.el8_5",
+          "link": "https://access.redhat.com/security/cve/CVE-2021-43529",
+          "riskFactors": [
+            "Remote execution",
+            "Attack complexity: low",
+            "Attack vector: network",
+            "Critical severity",
+            "Recent vulnerability"
+          ],
+          "impactedVersions": [
+            "*"
+          ],
+          "publishedDate": "2021-12-01T00:00:00Z",
+          "discoveredDate": "2022-05-18T12:24:22Z",
+          "layerTime": "2022-05-16T23:12:25Z"
+        },
+        {
+          "id": "CVE-2022-1650",
+          "status": "under investigation",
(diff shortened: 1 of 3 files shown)


libs/hdf-converters/test/mappers/forward/twistlock_mapper.spec.ts
```diff
@@ -1,8 +1,37 @@
 import fs from 'fs';
-import {describe, expect, it} from 'vitest';
+import {afterEach, describe, expect, it, vi} from 'vitest';
 import {TwistlockResults} from '../../../src/twistlock-mapper';
 import {omitVersions} from '../../utils';
 
+// Sequential: these tests spy on the shared console (vitest.config.ts runs
+// tests concurrently by default), see issue #8611 for the severity values.
```
Amndeep7, line 7: delete the comment

libs/hdf-converters/test/mappers/forward/twistlock_mapper.spec.ts
```diff
@@ -1,8 +1,37 @@
...
+describe.sequential('twistlock_mapper_severity_mapping', () => {
```
Amndeep7, line 8: rework this to be aligned with the standard style of test we've been doing w/r to generating a good hdf output that you then can use in future tests to compare against literally - see other tests.

if you'd like, you can keep the extra bit about mocking the console but i dunno if that's worth it.

libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -12,13 +12,30 @@ import {
...
+// report alongside Prisma's own tiers (mitre/heimdall2#8611; see the vendor
+// table in Prisma's CVSS-scoring doc). The no-rating values — 'unassigned',
+// 'not yet assigned', and Ubuntu's 'untriaged' — take the conservative
+// medium, matching dependency-track's 'unassigned' and grype's 'unknown'.
+// 'negligible' (Ubuntu: "technically a security problem" but theoretical/no
+// real damage) and 'unimportant' (Debian: the problem does not affect the
+// shipped binary package — Prisma ranks it even below negligible) are the
+// minimal ratings: 0.1 rather than 0.0 because the 0.0 tier is reserved for
+// informational non-findings across this codebase, and impact 0 exports as
+// Not_Applicable in checklists, hiding a finding the scanner reported.
```
Amndeep7, line 26: get rid of these comments and replace them with the link to the prisma docs.  i'd put that link in the internet archive so that we can retrieve it in the future if necessary.

libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -12,13 +12,30 @@ import {
...
   ['medium', 0.5],
   ['moderate', 0.5],
-  ['low', 0.3]
+  ['unassigned', 0.5],
+  ['not yet assigned', 0.5],
+  ['untriaged', 0.5],
+  ['low', 0.3],
+  ['unimportant', 0.1],
+  ['negligible', 0.1]
 ]);
```
Amndeep7, line 39: my understanding is that 'end-of-life' could also show up as an option but it does some processing to convert that into a score elsewhere?  might need to do some more research there

libs/hdf-converters/src/twistlock-mapper.ts
```diff
@@ -12,13 +12,30 @@ import {
...
   ['high', 0.7],
   ['medium', 0.5],
   ['moderate', 0.5],
-  ['low', 0.3]
+  ['unassigned', 0.5],
+  ['not yet assigned', 0.5],
+  ['untriaged', 0.5],
+  ['low', 0.3],
+  ['unimportant', 0.1],
+  ['negligible', 0.1]
```
Amndeep7, line 38: we should look at an actual example of this to determine if it's a 0.1 which is still marked as low in heimdall but the impact value is less that the actual low or if should mark it as 0, i.e. NA

libs/hdf-converters/sample_jsons/twistlock_mapper/sample_input_report/twistlock-twistcli-vendor-severities-sample.json
```diff
@@ -0,0 +1,211 @@
+{
```
Amndeep7, line 1: how did you generate this test file?  was it based off of a sample file coming from the repo that you adjusted, one from the customer that you anonymized, or madeup wholesale by the ai?

if the third option, please recreate it as one of the other two options with the manual adjustments necessary to exercise the various severities

Amndeep7 (changes requested):

libs/hdf-converters/sample_jsons/twistlock_mapper/sample_input_report/twistlock-twistcli-vendor-severities-sample.json
wdower: This file started out as the twistcli fixture we already had in the repo, but with a bunch of the extraneous packages removed (original fixture is 4k lines long) and with a few of the severities hand-edited to be the ones we're trying to get tests for, since those severities aren't in the current fixture.

So it's based off of a file that was already in here, but to be fair I did adjust statuses.

I checked, but I don't think we have a way to validate against Prisma's schema on the open internet. Possible that there's some error introduced by doing this. I want to source a real output file and try to convert it; if it works fine, then I say we keep this little handcrafted artisanal fixture for testing purposes.

libs/hdf-converters/src/twistlock-mapper.ts
wdower: https://github.com/hlxsites/prisma-cloud-docs/blob/main/docs/en/compute-edition/32/admin-guide/vulnerability-management/cvss-scoring.adoc

So they normalize EOL to low/med/high/crit, which we handle.

libs/hdf-converters/test/mappers/forward/twistlock_mapper.spec.ts
wdower: I think the console thing is worth it (guess I'm biased though. . .)

libs/hdf-converters/src/twistlock-mapper.ts
wdower: I don't have an actual Prisma file around as discussed, but I'm not sure it would tell us much here. Prisma, as noted, is just passing through what upstream vendors use for these severities, so the definition comes from Ubuntu: https://ubuntu.com/security/cves/about

Ubuntu calls `negligible`: "Technically a security problem, but only theoretical in nature, requires a very special situation, has almost no install base, or does no real damage. These typically will not receive security updates unless there is an easy fix and some other issue causes an update."

It's the lowest category of CVE Ubuntu tracks but it is still a CVE. My argument is that makes it non-zero but very low. Also note that Ubuntu has a "low" category and that "negligible" is explicitly lower than that.

We are not terribly consistent in Heimdall overall on how we assign severities, so I might be in conflict with some other converters, but no matter what we pick we will be. Suggest we don't get too hung up on individual severity calls in the same way that we don't with NIST mappings per-converter.

libs/hdf-converters/src/twistlock-mapper.ts
Amndeep7: yeah but i read "Note: 'end-of-life' will be set in the vulnerability status" as it will be in the file like this but in the ui it'll turn into the other thing which is why i'm concerned about having to map it.  if you think i'm reading it wrong, then yeah guess it's alright.

libs/hdf-converters/src/twistlock-mapper.ts
Amndeep7: i feel like that's a bit apples to oranges as the mappings discussion was about granularity whereas this is about if a user actually has to do a remediation of some kind (the fail associated with .1) vs the informational of a n/a.  

but yeah if they report it as a cve then yeah let's just give it a low and move on.  if they feel it was important enough to be a cve then it's important enough to count as a fail.  the users will just have to poam it out as a vendor is not gonna supply a fix in those scenarios until ubuntu actually does come out with one as described.

libs/hdf-converters/sample_jsons/twistlock_mapper/sample_input_report/twistlock-twistcli-vendor-severities-sample.json
Amndeep7: agreed

libs/hdf-converters/src/twistlock-mapper.ts
wdower: My read is as long as "end-of-life" status has already been mapped to some other status in the twistlock file on disk, then by the time we see it, EOL won't be in the source data. So we don't have to map it, it already was.

PR #8619, opened 2026-09-03
Hadolint mapper

Amndeep7:
- [x] Fixed a bug in the Nikto mapper
- [x] Cleaned up hdf converters a little bit
- [x] Created Hadolint mapper
  - [x] Has an option to extend the metadata by reaching out to the github wiki pages
  - [x] Has a Eugene-generated NIST mapping for the current set of rules


libs/hdf-converters/src/hadolint-mapper.ts
```diff
@@ -0,0 +1,192 @@
...
+  const pageWithoutTitle = lines.slice(1).join('\n');
+  return pageWithoutTitle ? markdownIt.render(pageWithoutTitle) : '';
+}
+
+async function fetchRuleDescription(code: string): Promise<string> {
+  try {
+    const {data} = await axios.get<string>(wikiUrl(code));
+    return extractRuleDescription(data);
+  } catch {
+    return '';
```
wdower, line 74: Should we warn here in the same way we do with the NIST mapping function if the rule description fetch doesn't find anything? Does not finding the rule in the wiki that imply that the rule is missing or malformed?

libs/hdf-converters/test/mappers/forward/hadolint_mapper.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+describe('hadolint_shellcheck_mapper', () => {
+  it('Successfully converts Hadolint ShellCheck data', async () => {
+    const mapper = new HadolintMapper(
+      fs.readFileSync(
+        'sample_jsons/hadolint/sample_input_report/heimdall_dockerfile.hadolint_shellcheck.json',
+        {encoding: 'utf-8'}
+      )
+    );
+
+    // fs.writeFileSync(
```
wdower, line 69: cleanup, I think

libs/hdf-converters/src/mappings/HadolintNistMapping.ts
```diff
@@ -0,0 +1,22 @@
...
+    );
+  }
+
+  controlForRule(rule: string): string | undefined {
+    return this.data.find((item) => item.rule === rule)?.control;
+  }
+
+  nistTag(rule: string): string[] {
+    const control = this.controlForRule(rule);
+    return control === undefined ? DEFAULT_TESTING_NIST_TAGS : [control];
```
wdower, line 20: global.ts already has DEFAULT_STATIC_CODE_ANALYSIS_NIST_TAGS. Do we consider something analyzing Dockerfiles to be SAST? Should we just reuse the existing global var, or did you intentionally add this one to be the really general catch-all tag?

Amndeep7: Eugene said that these don't have a security component per se which is why they don't have the RA-5, just the SA-11 by default.  Consequently I thought it would be worth to make a new var esp since the junit stuff coming down the line

libs/hdf-converters/test/mappers/forward/hadolint_mapper.spec.ts
Amndeep7: we intentionally leave those in on most of the mapper tests to make it easy to regenerate the sample files.

libs/hdf-converters/src/hadolint-mapper.ts
Amndeep7: https://github.com/hadolint/hadolint#rules

This says that it's an incomplete list of rules.  I don't know if this implies that the wiki will also not be complete.

I think we warn on the nist mapping not being there since it's something we can fix.  The rule description not being there is on the hadolint and shellcheck projects.  Not sure if we need to mention it.

wdower (approved):

PR #8694, opened 2026-09-21
Fix - Update Hadolint Shellcheck HDF Sample JSON

henrymontesanoiii:
Refreshed hadolint-shellcheck after ShellCheck wiki update to address unit test failing on updated HTML content of ShellCheck wiki text


libs/hdf-converters/test/mappers/forward/hadolint_mapper.spec.ts
```diff
@@ -3,6 +3,23 @@ import {describe, expect, it} from 'vitest';
 import {HadolintMapper} from '../../../src/hadolint-mapper';
 import {HadolintNistMapping} from '../../../src/mappings/HadolintNistMapping';
 import {omitVersions} from '../../utils';
+import { ExecJSON } from 'inspecjs';
+
+function omitRuleDescriptions(hdf: ExecJSON.Execution): ExecJSON.Execution {
```
Amndeep7: Put this in the utils directory instead of inline here.  Also rename to 'Descs' instead of descriptions since (confusingly) there is a descriptions attribute.  Might want to reconsider implementation slightly after looking at the other functions as an example.

libs/hdf-converters/test/mappers/forward/hadolint_mapper.spec.ts
```diff
@@ -3,6 +3,23 @@ import {describe, expect, it} from 'vitest';
...
+  return {
+    ...hdf,
+    profiles: hdf.profiles.map((profile) => ({
+      ...profile,
+      controls: profile.controls.map(({desc: _desc, ...control}) => control)
+    }))
+  } as ExecJSON.Execution;
+}
+
+function controlsHaveDescriptions(hdf: ExecJSON.Execution): boolean {
```
Amndeep7: Same thing re: renaming function.  I think this should probably also be in the utils module.  Use lodash for checks like if something is s string since their implementation is more rigorous for that specific case.

libs/hdf-converters/test/mappers/forward/hadolint_mapper.spec.ts
```diff
@@ -127,15 +144,15 @@ describe('hadolint_mapper_rule_descriptions', () => {
...
-          fs.readFileSync(
-            'sample_jsons/hadolint/hadolint-shellcheck-hdf-with-rule-descriptions.json',
-            {encoding: 'utf-8'}
-          )
-        )
-      )
+    const actual = await mapper.toHdf();
+    const expected = JSON.parse(fs.readFileSync(
+      'sample_jsons/hadolint/hadolint-shellcheck-hdf-with-rule-descriptions.json',
+      'utf-8'
```
Amndeep7: Do you not need the wrapper object?

Amndeep7 (changes requested):

